PR #8, opened 2023-03-11
initial p2p code

Boog900:


common/src/hardforks.rs
```diff
@@ -0,0 +1,54 @@
+use chrono::NaiveDateTime;
+
+use crate::network::Network;
+
+
+pub struct HardForks {
+    network: Network
+}
+
+// this function blindly unwraps only call when you know the timestamp is good
```
SyntheticBird45, line 5: Rust convention is to format satefy comment :
`// SAFETY: Reason`
maybe just change this comment as 
```
// This function use unwrap
// SAFETY : Its caller responsability to ensure the timestamp is valid.
```

SyntheticBird45:
I can't really understand if the network logic is correctly done. But my overall impression is that this is well organized but with a lack of documentation (that might be fixed in the future). Anyway you can merge whenever you want

PR #19, opened 2023-04-27
add logos

mmyjona:
Cuprate is dark blue superconducting material. so proposed idea is a mix of rust claw and magnetic field.

via bounty post [link](https://www.reddit.com/r/Monero/comments/12zrowe/cuprate_04_xmr_bounty_searching_for_a_logo/)


SyntheticBird45:
Hi, that's a great logo. After discussion we're very curious to see a version where the *C* of Cuprate is the magnetic lines, the claw would be inside with a little more blue than rust (darker this time tho). We would greatly appreciate 👍🏻 *(also you must have seen it, but we're Cuprate, not Caprate)*

mmyjona:
> Hi, that's a great logo. After discussion we're very curious to see a version where the _C_ of Cuprate is the magnetic lines, the claw would be inside with a little more blue (darker this time tho) than rust . We would greatly appreciate 👍🏻 _(also you must have seen it, but we're Cuprate, not Caprate)_

Check version 2. The new idea is dark blue magnet and orange crab claw as letter c. Magnetic lines are just too messy for minimalist style.

SyntheticBird45:
While I like the idea. I personally think the C should be taller or wider. And with boog we think it really miss something behind. It's too minimalistic

Boog900:
Thank you for working on these logos, sadly we have decided to go with another person but we are grateful for the time you put in to make these.

PR #32, opened 2023-10-26
Add `typos` + `cargo doc` CI

hinto-janai:
- Adds https://github.com/crate-ci/typos to CI
- Adds `cargo doc` checks to CI
- Adds `typos.toml` for false-positives and excluded files
- Fixes existing typos
- Changes `cargo fmt` to only run on 1 OS


hinto-janai:
@Boog900 ready.

.github/workflows/ci.yml
```diff
@@ -84,19 +96,23 @@ jobs:
...
-      run: cargo clippy --workspace --all-targets --all-features -- -D warnings
+      run: cargo clippy --workspace --all-features -- -D warnings
 
     - name: Test
-      run: cargo test --all-features --workspace --all-targets
+      run: cargo test --all-features --workspace
 
     # TODO: upload binaries with `actions/upload-artifact@v3`
     - name: Build
-      run: cargo build --all-features --all-targets --workspace
```
Boog900, line 102: `--all-targets` does not mean all target archs it means tests, benches, libs etc so we should keep it for `clippy` and `build`. For `test` it seems specifying  `--all-targets` acctually reduces the amount of our tests ran: it skips docs tests so we should keep `--all-targets` removed for that

.github/workflows/ci.yml
```diff
@@ -13,16 +13,18 @@ env:
   RUST_BACKTRACE: "full"
   # Increase thread stack size to 8 megabytes.
   RUST_MIN_STACK: 8000000
+  # Fail on documentation warnings.
+  RUSTDOCFLAGS: '-D warnings'
```
Boog900, line 17: can this be passed in like we do with clippy or can you not do that with `doc`?

Boog900 (changes requested):

.github/workflows/ci.yml
hinto-janai: You cannot - we could do `RUSTDOCFLAGS='-D warnings' cargo doc` but that causes a force-rebuild, so it's set for the whole CI environment.

hinto-janai:
`handshake_cuprate_to_monerod` failed https://github.com/Cuprate/cuprate/actions/runs/7909525054/job/21590698059?pr=32#step:11:490.

.github/workflows/ci.yml
Boog900: ah alright yeah this is better then

Boog900 (approved):
It passed after a re-run hopefully that doesn't happen too often.

PR #35, opened 2023-10-30
Database

hinto-janai:
## Part 1
This is the 1st in a chain of PRs that will re-implement `database/`.

Each PR will focus on a specific scope for review purposes.

This one is for:
- general design decisions
- broad file structure
- data type definitions
- small stuff like lints, naming, etc

## Goals
The main goal for this re-implementation is allow for a clean separation between the specific operations other parts of Cuprate would like to perform (`get_block()`, `add_block()`) and the underlying database (`LMDB`, `sanakirja`, etc) such that the callers do not know - and do not care about the actual underlying database.

A benefit this will allow is for Cuprate to "hot-swap" databases, ideally simply by building with a different `--feature`:
```bash
# Build Cuprate using heed.
cargo build --release --feature heed
```
```bash
# Build Cuprate using sanakirja.
cargo build --release --feature sanakirja
```

After testing to see which database has the best characteristics for the types of operations Cuprate performs (considering latency, throughput, parallelism, etc), a "one true" database will be selected as the default, although the underlying harness to quickly swap to another one or even add more in the future will always be there.

This database "layer" must also:
- Be interfaced as a `tower::Service`
- Not block the async runtime

## Divergence from `monerod`
Cuprate's database will diverge from [`monerod`](https://github.com/monero-project/monero)'s in the following ways:

1. Schema will most likely be different
2. Compatibility with `.bitmonero` as-is is low priority 

## Database to implement
The database implementations in mind (for now) include:
- LMDB (via [heed](https://github.com/meilisearch/heed))
- [sanakirja](https://docs.rs/sanakirja)

Both of these will be forked and maintained by Cuprate as there are `monerod`-specific things that must be added (e.g output lookup using [sub-keys](https://github.com/monero-project/monero/blob/ac02af92867590ca80b2779a7bbeafa99ff94dcb/src/blockchain_db/lmdb/db_lmdb.cpp#L3447C1-L3449C80)).

## Layers
This system is logically abstracted into 5 layers internally.

Starting from the lowest layer:

| Layer | Name | Description |
|-------|------|-------------|
| 1     | Database | This is the actual database, or a thin Rust shim on-top of the database that calls database operations directly, e.g `get()`, `commit()`, `delete()`, etc
| 2     | Trait | This is the `trait` that abstracts over _all_ databases and allows keeping the function signatures and behavior the same but allows for swapping out databases; each database will have this implemented located in `src/backend/`, with each database (`LMDB`, `sled`, etc) having its own file defining the mappings. This `trait` is meant to cover all features across databases, and will have provided methods that may not necessarily be the most efficient - if a database can implement a method in a better way, it is re-implemented and will shadow the provided version
| 3 | Abstract Database | This is a concrete object and handle to _some_ database that implements the generic `trait`
| 4 | Thread | This is the dedicated _writer_ thread that is the logical owner of the abstract database. It acts as a sort of "write queue" that maps channels messages from other Cuprate crates into database _write_ calls into eventual responses to the callers via a channel
| 5 | Services | This is the public API that other Cuprate crates will interface with; the abstract database will have `tower::Service<ReadRequest>` implemented where `ReadRequest` is an `enum` with all the specific high-level read-only requests other Cuprate crates need e.g. `get_block()` and `tower::Service<WriteRequest>` where it is the same, but with writes e.g. `add_block()` - the callers will be responsible for execution if this is a read (or perhaps a request is sent to a "reader" thread-pool?), and write operations will be "queued" by sending them to "the thread" which eventually returns the result of the function

<div align="center">
    
</div>

## The `tower::Service`
The `tower::Service` will act as the public API for the database that other parts of Cuprate will call. The service will contain an async channel that requests can be sent down.

They only care about making a database request and receiving a result - they have no view into the internal implementation.

A request will look like:
```rust
struct Request {
  request: DatabaseRequest,
  response_chan: tokio::sync::oneshot::Sender<Result<DatabaseResponse, Err>>
}
```

## Parallelism
The databases planned to be implemented all support MVCC, so Cuprate itself doesn't necessarily need to bake-in any synchronization.

Readers scale per core (in ideal situations) with MVCC-style databases, and as such, Cuprate will spawn, control, and maintain a pool of threads (outside of `rayon`, most likely manually spawned) that will handle multiple read and write requests.

These database threads will operate in a `work-stealing, lockless-queue` fashion, where readers/writers outside of `database/` send some requests, and the database threads:
1. Pop the request out
2. Do the work
3. Send result to the requester

## Messages
If we were to go with a split read/write service we would need 4 `enum`'s:
- A read request enum which is given to the read services
- A write request enum which is given to the write service
- A joint read/write enum which the services send to the channel
- A response enum which is returned down the oneshot

## Previous Ideas
These are previous design ideas for the database implementation that were ruled out.

```
I = Idea
R = Reason for being a bad idea
```

**I.** All outer Cuprate crates send/receive *all* database-related requests (via channels) to a single "database" thread that logically owns the database object
**R.** This effectively limits concurrent readers to 1, as all requests/responses would be funneled through a single thread

**I.** Just put the DB operations in the async runtime
**R.** By doing this we risk taking out an async thread from the pool if the DB blocks


SyntheticBird45:
> This effectively limits concurrent readers to 1, as all requests/responses would be funneled through a single thread

My original idea was that there was an external thread that would logically own the writer thread and multiple reader threads. This thread is just owning the two working-queue for read and write operations. It is effectively funneling requests through a single thread (which is performance loss) but it doesn't funnel responses since the workers are using the oneshot themselves. It it just equivalent to having actors outside of the async runtime in order to avoid blocking it.

> Just put the DB operations in the async runtime

This might be misleading as written as it is. Our main concerns were that tokio blocking threads were not adequate for the work since the database is quickly pulling in and out blocking tasks, leading to async halting.
I think we should reconsider this after some research, and help from others in rust community, it seems impossible that such things happen.

Tokio is indeed using a different threadpool for blocking tasks with a working-queue based mechanism. The number of threads grow and shrink based on demand. The time before an unused thread is closed can be configured. The maximum amount of threads used for blocking tasks can be configured. The interval between polling of work can also be configured. Since database operations are quick we can reduce the time between polling in order to more quickly push new transactions and pull results. 
Smol-rs (another async runtime) blocking pool implementation is basically the same as Tokio one, is standalone and also make it impossible for async to halt on high-demand.

TL;DR: If async are halted by database operations, it is OS scheduling fault, not runtime one

SyntheticBird45:
Otherwise the database draft seems good to me, we should just make another draft on what high-level request would the database need. I loose days of my life sorting the monerod's database call and only keeping the essential ones, it maybe productive to start listing what we need.

Boog900:
> Our main concerns were that tokio blocking threads were not adequate for the work since the database is quickly pulling in and out blocking tasks, leading to async halting.

Using tokio blocking threads just moves the threads from the database to tokio, the separate threads still exist. Instead of having the DB be async and just immediately call `spawn_blocking` on a request we move the whole actor to it's own thread. I don't think `quickly pulling in and out blocking tasks, leading to async halting` was our concern, at least it wasn't mine. My concern was with putting the DB in the async runtime: the DB could block and I felt spawning separate threads for the actors were a better solution than using `spawn_blocking` on each request.

> we should just make another draft on what high-level request would the database need. 

Yep ill be uploading a schema and here is what the consensus code uses for DB requests/ responses so is a minimum requirement: https://github.com/Cuprate/cuprate/blob/ec4fe9b79d39fbb8bf00d4eb54f9599c7c06a317/consensus/src/lib.rs#L114C1-L132C1

SyntheticBird45:
> Using tokio blocking threads just moves the threads from the database to tokio, the separate threads still exist.
 
 Indeed
 > Instead of having the DB be async and just immediately call spawn_blocking on a request we move the whole actor to it's own thread.
 > My concern was with putting the DB in the async runtime: the DB could block and I felt spawning separate threads for the actors were a better solution than using spawn_blocking on each request.
 
 orry for the confusion. I extrapolated the scenario in which the node is halted by high-demand in blocking code, I thought this was our concern but you're right, we didn't discuss deeply into that.

Anyway, we've the guarantee that it shouldn't happen. The choice, as you said in your first sentence, is choosing to implement our own actors or use tokio ones. The performance overhead might not be significant between our own and tokio's impl (i'll try to craft some quick benchmarks to be sure).
 
 If we imagine that tokio and custom actors are similar in performance (working-queue, oneshot channel):
#### Pros for using tokio:
- No need to reinvent the wheel, so database dev can go faster on other tasks
- convenient joinhandle, the execution and return of the result is handled transparently by tokio
- dynamic thread supply, less thread if not needed (when node is synced), more thread if needed (node is actively syncing).
- Fault recovery, if a thread crash (I hope not), a new thread will quickly come and replace it.
- (opinion) we can suppose that Tokio devs have made fine-tuned optimization on the channel/synchronization between the end of the task and the future that return it.
#### Pros for using custom impl:
- Potentially simpler implementation than Tokio, so more easy to debug in case of breakage.
- Might add more features in the future.
- Spare ourselves the cloning of `Arc<dyn AbstractDatabase>` at every transactions, depends on whether actors/threads clone it at start or receive it. maybe @hinto-janai can highlight us on this.

Just to be clear, I'm not saying one is better than the other but we should carefully check the pros and cons of each to decide whether it something to invest time on or not. My opinion, at the moment, is that I don't see what benefits would developing our own actors would bring us over tokio ones.

Boog900:
The cost of making our own actors is not high IMO compared to other tasks.

By using `spawn_blocking` we also lose the option of being able to act on readiness, unless we mix in a semaphore or something as we aren't able to poll the capacity of the channel which is a big point of `tower`.

If I'm honest I don't think the performance difference is going to be great enough to warrant a big overhaul of the design, and would question if you would be able to make a bench that would accurately portray our situation.

> No need to reinvent the wheel, so database dev can go faster on other tasks

All we are doing is spwaing threads and using a channel, we won't be building a whole runtime.

> convenient joinhandle, the execution and return of the result is handled transparently by tokio

Inconvenient spawning of tasks though - needing to move everything and make sure it's `Send` and whatever.

> dynamic thread supply, less thread if not needed (when node is synced), more thread if needed (node is actively syncing).

We can do this based on the amount of requests in the channel. (later though, not now)

> Fault recovery, if a thread crash (I hope not), a new thread will quickly come and replace it.

If a thread crashes I want the whole of Cuprate to crash.

I really don't think this is worth it, Cuprates DB design has been going back and forth for a year now, and although I am ok with changes I don't feel this provides enough (any) benefit.

SyntheticBird45:
> By using spawn_blocking we also lose the option of being able to act on readiness, unless we mix in a semaphore or something as we aren't able to poll the capacity of the channel which is a big point of tower.

No need for semaphore, we could add an atomic count synced through executing blocking tasks. Tower::is_ready() just give Poll(Ready) if the atomic account is less than the maximum or whatever logic behind.

> All we are doing is spwaing threads and using a channel, we won't be building a whole runtime.
> Inconvenient spawning of tasks though - needing to move everything and make sure it's Send and whatever.

You're right on this one.

> We can do this based on the amount of requests in the channel. (later though, not now)

For sure, it'll require some request/s logic and some tuning tho

> If a thread crashes I want the whole of Cuprate to crash.

True

Boog900:
> No need for semaphore, we could add an atomic count synced through executing blocking tasks. Tower::is_ready() just give Poll(Ready) if the atomic account is less than the maximum or whatever logic behind.

We need to wake when we are ready if we return `Poll::Pending` so I don't think this will work.

hinto-janai:
> Spare ourselves the cloning of `Arc<dyn AbstractDatabase>` at every transactions

I think most threads who need to access the DB would already have an `Arc<AbstractDatabase>` somewhere, not clone for each transaction - although maybe some infrequent one-off jobs would do this.

hinto-janai:
This is ready. 

Overview:
1. The current `database/` -> `old_database/`
2. `database/README.md` docs added
3. `cargo docs` added
4. Most types & function signatures added (but not implemented)
5. `tower::Service` + thread-pool layed out (but not implemented)

I think reading `README.md` and:
```bash
cargo docs -p cuprate-database --open
```
will give a better overview of this PR than the diffs.

There's lots of missing documentation/functions/implementations, but the general structure is mostly done.

To be done:
- `heed` backend implementation
- `sanakirja` backend implementation
- Compatible (de)serialization methods (`heed` and `sanakirja` differ)
- Sub-key support (in both `heed` and `sanakirja`)
- Implementation of `trait Monero` (high-level database functions)
- Proper error enums
- Proper `Response/Request` enums
- The `tower::Service` + `read/write` thread-pool system
- Tests for the thread-pool system
- Tests for the database
- Probably many missing functions from the major traits (`Env`, `Database`, etc)

This could also wait until https://github.com/Cuprate/cuprate/pull/52 is merged to check if CI works.

database/src/monero.rs
```diff
@@ -0,0 +1,198 @@
...
+///
+/// # TODO
+/// TODO: What is a good name for this trait?
+///
+/// TODO: These functions should pretty much map 1-1 to the `Request` enum.
+///
+/// TODO: These are function names from `old_database/` for now.
+/// The actual underlying functions (e.g `get()`) aren't implemented.
+#[allow(missing_docs)]
+pub trait Monero: Env {
```
Boog900, line 47: I don't think this trait is needed, I think these methods should just be functions which take in a generic `RoTx`/`RwTx`. This would mean an underlying DB can't re-implement a method in a more optimized way for that back-end, but I don't think this is an issue as if our `Table` trait contain enough info (more on that in another comment) then the back-end can optimize for generic tables with certain characteristics.

database/src/service/free.rs
```diff
@@ -0,0 +1,137 @@
...
+//---------------------------------------------------------------------------------------------------- Init
+#[cold]
+#[inline(never)] // Only called once.
+/// Initialize the database, the thread-pool, and return a read/write handle to it.
+///
+/// This initializes a [`OnceLock`] containing the [`ConcreteEnv`],
+/// meaning there is only 1 database per program.
+///
+/// Calling this function again will return handles to the same database.
+pub fn init() -> &'static (DatabaseReadHandle, DatabaseWriteHandle) {
```
Boog900, line 34: IMO I think it would be better to have the function `init` just spawn the DB and give handles back to the caller who can then give handles to the DB to whatever parts of Cuprate need it, instead of allowing any part of Cuprate to call `db_read`/ `db_write` and get a handle

database/src/service/free.rs
```diff
@@ -0,0 +1,137 @@
...
+    DATABASE_HANDLES.get_or_init(|| {
+        // Initialize the database itself.
+        let db: ConcreteEnv = todo!();
+        // Leak it, the database lives forever.
+        //
+        // TODO: there's probably shutdown code we have to run.
+        // Leaking may not be viable, or atleast, we need to
+        // be able to run destructors.
+        let db: &'static ConcreteEnv = Box::leak(Box::new(db));
```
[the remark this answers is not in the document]
Boog900: this would also make shutdown easier.

database/src/service/read.rs
```diff
@@ -0,0 +1,178 @@
...
+    service::{request::ReadRequest, response::ReadResponse},
+    ConcreteEnv,
+};
+
+//---------------------------------------------------------------------------------------------------- Types
+/// The read response from the database reader thread.
+///
+/// This is an `Err` when the database itself errors
+/// (e.g, get() object doesn't exist).
+type Response = Result<ReadResponse, RuntimeError>;
```
Boog900, line 19: I think this is just meant to represent the successful state, so this should just be `ReadResponse` and then the Future should return Result<Response, Error>

database/src/service/read.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn main(mut self) {
+        loop {
+            // 1. Hang on request channel
+            // 2. Map request to some database function
+            // 3. Execute that function, get the result
+            // 4. Return the result via channel
+            let (request, response_send) = match self.receiver.recv() {
+                Ok((r, c)) => (r, c),
+                Err(e) => {
+                    // TODO: what to do with this channel error?
```
Boog900, line 135: you just return, an error means all handle are dropped so the DB should shutdown (If we change to and Arc)

database/src/service/read.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[cold]
+    #[inline(never)] // Only called once.
+    pub(super) fn init(db: &'static ConcreteEnv) -> DatabaseReadHandle {
+        // Initalize `Request/Response` channels.
+        let (sender, receiver) = crossbeam::channel::unbounded();
+
+        // TODO: slightly _less_ readers per thread may be more ideal.
+        // We could account for the writer count as well such that
+        // readers + writers == total_thread_count
+        let readers = cuprate_helper::thread::threads().get();
```
Boog900, line 104: We don't have to do this now - we can add it later but I would like there to be a config where you would be able to set the amount of threads and other things that can be changed about the DB.

database/src/service/request.rs
```diff
@@ -0,0 +1,57 @@
...
+//---------------------------------------------------------------------------------------------------- Import
+
+//---------------------------------------------------------------------------------------------------- Constants
+
+//---------------------------------------------------------------------------------------------------- Request
+#[derive(Debug)]
+/// Either a read or write request.
+///
+/// TODO: not sure if we actually need this.
+pub enum Request {
```
Boog900, line 14: I also don't think we need this if we have separate threads for read and write ops

database/src/service/response.rs
```diff
@@ -0,0 +1,53 @@
...
+//---------------------------------------------------------------------------------------------------- Import
+
+//---------------------------------------------------------------------------------------------------- Constants
+
+//---------------------------------------------------------------------------------------------------- Response
+#[derive(Debug)]
+/// Either a read or write response.
+///
+/// TODO: not sure if we actually need this.
+pub enum Response {
```
Boog900, line 14: We do need this, as it would be good to have the read and write services return the same thing.

I think this could be changed though as the only thing a write op should return is `Ok` so instead of having 3 enums just add `Ok` to the read response and make that the `Response` enum

database/src/service/write.rs
```diff
@@ -0,0 +1,210 @@
...
+/// The write response from the database writer thread.
+///
+/// This is an `Err` when the database itself errors
+type Response = Result<WriteResponse, RuntimeError>;
+
+/// The `Receiver` channel that receives the write response.
+type ResponseRecv = tokio::sync::oneshot::Receiver<Response>;
+
+/// The `Sender` channel for the response.
+type ResponseSend = tokio::sync::oneshot::Sender<Response>;
```
Boog900, line 24: same here as with the read service

database/src/service/write.rs
```diff
@@ -0,0 +1,210 @@
...
+//---------------------------------------------------------------------------------------------------- DatabaseWriteHandle
+/// Write handle to the database.
+///
+/// This is cheaply [`Clone`]able handle that
+/// allows `async`hronously writing to the database.
+///
+/// Calling [`tower::Service::call`] with a [`DatabaseWriteHandle`] & [`WriteRequest`]
+/// will return an `async`hronous channel that can be `.await`ed upon
+/// to receive the corresponding [`WriteResponse`].
+#[derive(Clone, Debug)]
```
Boog900, line 35: I would remove the `Clone` as only one place should write to the DB.

database/src/service/write.rs
```diff
@@ -0,0 +1,210 @@
...
+    ///
+    /// During this time, it would be wasteful if `Request2`'s was
+    /// just being waited on, so instead, `Writer2` handles that
+    /// work while `Writer1` is in-between tasks.
+    ///
+    /// This leeway is pretty tiny (doesn't take long to allocate a channel
+    /// and send a response) so there's only 2 Writers for now (needs testing).
+    ///
+    /// The database backends themselves will hang on write transactions if
+    /// there are other existing ones, so we ourselves don't need locks.
```
Boog900, line 118: Although this is a good idea, only one place will write to the DB and as the requestor *needs* to await the response (otherwise the operation could have failed) you should never have 2 write requests in the channel.

So I think there should only be 1 write thread.

database/src/table.rs
```diff
@@ -1,181 +1,49 @@
...
+//---------------------------------------------------------------------------------------------------- Import
+use crate::pod::Pod;
+
+//---------------------------------------------------------------------------------------------------- Table
+/// Database table metadata.
+///
+/// Purely compile time information for database tables.
+/// Not really an accurate name for `K/V` database but
+/// this represents the metadata of a `K/V` storing object.
+pub trait Table {
```
Boog900, line 12: This table should probably contain more information about the data it's storing so the backend can optimize for certain tables.

Currently I think the only thing that's missing that we would be able to optimize for is:
`const CONSTANT_SIZE: bool`
 
for if the table's values are all the same size.

database/src/table.rs
```diff
@@ -1,181 +1,49 @@
...
+
+    /// Name of the database table.
+    const NAME: &'static str;
+
+    cfg_if::cfg_if! {
+        if #[cfg(all(feature = "sanakirja", not(feature = "heed")))] {
+            // TODO: fix this sanakirja bound.
+
+            /// Concrete key type.
+            type Key: Pod + sanakirja::Storable;
```
Boog900, line 29: I think we should have a trait for DBKeys as well: 
```rust
trait DBKey {
    /// A bool for if this table requires multiple keys to reach a value.
    const DUP: bool;
    type Primary;
    /// This will only be used for DUP tables so can be set to a default type for single key tables
    type Secondary;

    fn primary(self) -> Self::Primary;

    fn primary_secondary(self) -> (Self::Primary, Self::Secondary);
}
```

So then the back-end would do something like:

```rust
fn get_val<T: Table>(key: T::Key) -> T::Value {
   if <T::Key as DBKey>::DUP {
      let (primary, secondary) = key.primary_secondary();
      // get value with both keys
   } else {
       // get val normally.
   }
}
```

Boog900 (commented):
(I didn't mean to approve before you replied, my bad)

hinto-janai:
@Boog900 macOS CI seems to be running <1.75 Rust, it's not allowing `-> impl Trait` https://github.com/Cuprate/cuprate/actions/runs/7876938190/job/21492022001?pr=35#step:8:112

Boog900:
@hinto-janai add `rustup update` as a step to CI?

database/src/table.rs
hinto-janai: Should this be on `trait Table`? It would then be associated with a table:
```rust
pub trait Table {
    const NAME: &'static str;
    const CONSTANT_SIZE: bool;
    const DUPLICATE: bool;

    type KeyPrimary: Pod;
    type KeySecondary: Pod;
    type Value: Pod;

    fn primary(self) -> Self::Primary;
    fn primary_secondary(self) -> (Self::Primary, Self::Secondary);
}
```
If it were a separate trait, it would be implemented on the key type itself so 2 tables with the same key would conflict, right?
```rust
trait DBKey { /* ... */ }
impl DBKey for u8 {
    /* ... */
    const DUP: bool = true;
}

// Table with dup keys.
impl Table for Table1 {
    /* ... */
    type Key = u8;
}

// Table without dup keys.
impl Table for Table2 {
    /* ... */
    type Key = u8; // <- same key, different table, DBKey impl conflicts
}
```

Boog900: It could be, if you think that would be better.

My idea was for a type like `u8` to always be `DUP: bool = false` and for there to be a single type `DupKey` that sets `DUP = true`. The `DupKey` would look something like:

```rust
struct DupKey<P, S> {
   primary: P,
   secondary: S 
}
```

and then if the table needed a dup key we would set the key to `DupKey<u8, u8>` for example. 

I am happy to do it either way.

hinto-janai: How would `fn primary(self) -> Self::Primary` work? `Table` is implemented on an empty struct so `self` is pretty much `()`.

Would this work?

```rust
struct DupKey<P, S> {
   primary: P,
   secondary: S 
}

trait Table {
    /* ... */
    fn primary_secondary(key: DupKey<Self::Primary, Self::Secondary>) -> (Self::Primary, Self::Secondary);
}

impl Table for /* table */ {
    type Primary = u8;
    type Secondary = u64;

    /* ... */

    fn primary_secondary(key: DupKey<u8, u64>) -> (Self::Primary, Self::Secondary) {
        (key.primary, key.secondary)
    }
}
```

hinto-janai: Although `DupKey` would have to be the `Primary` as well, so maybe:

```rust

impl Table for /* table */ {
    type Primary = DupKey<u8, u64>;
    type Secondary = u64;

    /* ... */

    fn primary_secondary(key: DupKey<u8, u64>) -> (Self::Primary, Self::Secondary) {
        (key, key.secondary)
    }
}
```
is `fn primary()` needed? We would have direct access to `T::Key` anyway in the `fn get_val<T: Table>(key: T::Key) -> T::Value` example

Boog900: You are slowly getting to my idea :) I think it would probably be easier to go with the separate trait right?

hinto-janai:
@Boog900 All changes added.

SyntheticBird45:
Hi,
I checked you code and I think it's great. As well as boog's review and your changes. I just added a small comment on cargo toml description but beside that I'm for merging

database/src/key.rs
```diff
@@ -0,0 +1,127 @@
...
+    (
+        $(
+            $t:ident // Key type.
+        ),* $(,)?
+    ) => {
+        $(
+            impl Key for $t {
+                const DUPLICATE: bool = false;
+                type Primary = $t;
+                type Secondary = $t;
```
Boog900, line 72: I think the should be the unit type `()` with a TODO to make it the never type `!` when that stabilizes

database/src/table.rs
```diff
@@ -1,181 +1,49 @@
...
+    const NAME: &'static str;
+
+    /// Whether the table's values are all the same size or not.
+    const CONSTANT_SIZE: bool;
+
+    // TODO: fix this sanakirja bound.
+    cfg_if::cfg_if! {
+        if #[cfg(all(feature = "sanakirja", not(feature = "heed")))] {
+            /// Primary key type.
+            type Key: Key + Pod + sanakirja::Storable;
```
Boog900, line 31: The `Pod` and `Storable` bound should be on the `Primary` and `Secondary` types of the `Key` trait

Boog900 (approved):
Thank you, I have left a couple more comments but these are small and can be handled in a follow up PR

database/src/table.rs
hinto-janai: This means `DupKey` needs to implement `Pod`. How should it be implemented?

database/src/key.rs
hinto-janai: https://github.com/Cuprate/cuprate/pull/60

database/src/table.rs
hinto-janai: https://github.com/Cuprate/cuprate/pull/61

vorot93:
What was the reasoning for the switch away from MDBX?

vorot93:
What was the reasoning behind the switch away from MDBX?

SyntheticBird45:
@vorot93 I gave you a first answer of Matrix.
Hopefully @Boog900 have a more accurate answer to your question.

vorot93:
@SyntheticBird45 all I got is Matrix 'undeciphered message' gibberish after not logging in for a while. Unironically, that platform is a poor choice for comms, compared to both easy ([Telegram](https://telegram.org)) and hip decentralized web3 ([SimpleX Chat](https://simplex.chat)) alternatives.

SyntheticBird45:
@vorot93 I'm sorry, I can't find how to verify respond to your verification request. Matrix UX is indeed poor. I would prefer that you forget my answer I tried to give you. @Boog900 should answer that himself.

vorot93:
By trying out as many databases as possible, you never really extract the benefits of using one and lose out on performance.

Moreover, blockchain node workloads are very specific (random read/sequential write-heavy) and almost all LMDB-class embedded storage engines are incomplete garbage (including all Rust ones).

Just go all in on MDBX and focus on business logic. You won't regret it.

Boog900:
Yeah maybe we lose the ability to micro-optimize a specific DB by building abstractions that work with many DBs but at the moment I think trying out multiple DBs and seeing what performs best for our workload is the right approach. 

We will eventually settle on a "main" DB, when that happens we can then start looking at optimizing more for that DB.

Maybe that will be MDBX, and if it is at least we will _know_ that MDBX is the best choice instead of always having the if.

PR #43, opened 2023-12-10
JSON-RPC 2.0

hinto-janai:
This is a straight copy+paste of https://github.com/hinto-janai/json-rpc.

It only contains types and (de)serialization for [`JSON-RPC 2.0`](https://www.jsonrpc.org) and doesn't contain any extra stuff like HTTP/Client/etc.

There are already impls out there ([Bitcoin](https://github.com/apoelstra/rust-jsonrpc), [Parity](https://github.com/paritytech/jsonrpsee)) but I think JSON-RPC 2.0 is small enough to maintain on our own.

Moving it in-tree for Cuprate instead of an `external/` submodule is probably desired since changes can be made directly.

## Todo
This impl currently has:
- 0-copy (de)serialization with `Cow<'_, T>`
- Bunch of helper functions and `const` values

It is missing:
- More testing
- Documentation
- [Batching](https://www.jsonrpc.org/specification#batch) support (although, I don't think Monero supports this)


hinto-janai:
Will re-open when needed to avoid merge conflicts.

PR #45, opened 2023-12-12
`helper/`

hinto-janai:
## What
`helper/` is the kitchen-sink crate for very generic, not necessarily Cuprate specific functions, types, etc.

If a 3rd party's crate/functions/types are small enough, it could be moved here to trim dependencies and allow easy modifications.

## Currently Implemented
`helper` is split into modules - each one filling in some niche. Each module is feature gated with `#[cfg()]`, pulls in dependencies only when needed, and is `#[no_std]` if possible. All features/modules are enabled by default.

A list of the modules and some examples of what they're used for.

| Module   | Example function/type |
|----------|-----------------------|
| `asynch` | `InstaFuture<T>` - future that is ready immediately
| `atomic` | `AtomicF(32\|64)` - atomic float backed by `AtomicU(32\|64)`
| `crypto` | Cryptographic code - `check_point()`
| `num`    | Generic number code - `median<T>(nums: &[T]) -> T` where `T` is any number
| `thread` | Thread code - `THREADS()` for system thread count
| `time`   | System time and general time utils - `current_time()` for current UNIX timestamp


kayabaNerve:
https://docs.rs/futures/latest/futures/future/fn.ready.html for "InstaFuture" (`futures` is even already a dependency of your helper crate).

The docstring on `check_point` is wrong. It checks the point is *canonically encoded*. "canonical" is ambiguous. Also, I'd hope for a much better function name as "check" alone is not immediately clear as to what's being checked.

I'm unsure why you'd want a static for a percentage of threads instead of a function which returns the percentage. The addition of multiple static values with semi-arbitrary hardcoded percentages takes a lot of space in comparison and increases memory usage. While the ~16 bytes isn't notably on most computers, you can start counting single-digit bytes on embedded systems.

The docstring on `current_time` is wrong. It panics if the system time is before the Unix epoch, not if it 'fails to get the system time'. The function name is also imprecise as it doesn't return the current time, a concept. It returns the current unix timestamp, a specific measurement of the current time (as the docstring comments).

Since `chrono` is a dependency for this module, I have no idea why you wouldn't use its functions for most of the functionality present in that file. `chrono::offset::Local::now().time()` trivializes accessing `hms`.

hinto-janai:
Everything added so far is lazily copy+pasted from other places (including from [Cuprate](https://github.com/Cuprate/cuprate/blob/92652b26a29fb8e43561b1dba986a0646bd8c429/consensus/src/helper.rs#L61)).

This is a draft mostly for discussion, nothing here is final.

Boog900:
> https://docs.rs/futures/latest/futures/future/fn.ready.html can replace "InstaFuture" (futures is even already a dependency of your helper crate).

I did not know this existed, thanks! It seems pretty much the exact same as `InstaFuture`.

> It checks the point is canonically encoded. "canonical" is ambiguous. Also, I'd hope for a much better function name as "check" alone is not immediately clear as to what's being checked.

Yep thats my bad, I would also say this function does not need to be in `helper/` only the consensus code needs this function so it can just be kept there.

kayabaNerve:
I didn't mean to call anyone out/be overtly harsh, sorry if so. I just saw a PR and had comments on it. Just trying to chime in and help as I can :)

hinto-janai:
Note to self -  This PR's `AtomicF(32|64)` can be replaced by either [`crossbeam::atomic::AtomicCell<f32/f64>`](https://docs.rs/crossbeam/latest/crossbeam/atomic/struct.AtomicCell.html) or [`atomic::Atomic<f32/f64>`](https://docs.rs/atomic/0.6.0/atomic/struct.Atomic.html) as these are lock-free internally (on platforms that support 32/64-bit atomic ops respectively).

Boog900:
I'm going to merge this as is with the expectation that a lot of this will be changed/ removed if not needed.

PR #50, opened 2024-02-04
cryptonight: silence `ftime` deprecated warning

hinto-janai:
Silences this warning when running `cargo doc/clippy/build`.

```
warning: cryptonight-cuprate@0.1.0: c/oaes_lib.c: In function ‘oaes_get_seed’:
warning: cryptonight-cuprate@0.1.0: c/oaes_lib.c:515:9: warning: ‘ftime’ is deprecated: Use gettimeofday or clock_gettime instead [-Wdeprecated-declarations]
warning: cryptonight-cuprate@0.1.0:   515 |         ftime (&timer);
warning: cryptonight-cuprate@0.1.0:       |         ^~~~~
warning: cryptonight-cuprate@0.1.0: In file included from c/oaes_lib.c:45:
warning: cryptonight-cuprate@0.1.0: /usr/include/sys/timeb.h:29:12: note: declared here
warning: cryptonight-cuprate@0.1.0:    29 | extern int ftime (struct timeb *__timebuf)
warning: cryptonight-cuprate@0.1.0:       |            ^~~~~
```

Either `gettimeofday` or `clock_gettime` is recommended but I think they have slightly different precision semantics.

https://man7.org/linux/man-pages/man3/ftime.3.html

PR #52, opened 2024-02-07
ci: include macos + windows

hinto-janai:
Installs `boost` on macOS + Windows and adds them back to CI.

To be fixed:
- `monerod` only spawned on Linux https://github.com/hinto-janai/cuprate/actions/runs/7808923461/job/21299883505
- Windows error from https://github.com/Cuprate/cuprate/pull/50 (https://github.com/hinto-janai/cuprate/actions/runs/7808733161/job/21299591221)


SyntheticBird45:
Pardon my indulgence, but why do we need `boost` ?

hinto-janai:
I'm not sure, it was added in https://github.com/Cuprate/cuprate/pull/33. It seems to only be used here: https://github.com/Cuprate/cuprate/blob/1f677562eb753d07e2d035f0690a22e096140aae/cryptonight/c/warnings.h#L14

hinto-janai:
I think this PR works, but it's stuck on `cryptonight/` using `#include <unistd.h>` in some places which doesn't exist on Windows.

`cryptonight/build.rs` needs whatever workarounds Monero's CMake files use.

Boog900:
On windows Monero uses MSYS2 as a build environment: https://github.com/monero-project/monero?tab=readme-ov-file#on-windows

I think that provides `unistd.h` right? 

I don't know how easy it will be to work around that.

SyntheticBird45:
Compiles fine with `msys2-runtime-devel` package installed

Boog900:
Good news cryptonight-cuprate builds, bad news RandomX does not.

Boog900:
I think this is going to require adding `windows-gnu` support to randomX or adding `windows-msvc` to cryptonight-cuprate.

SyntheticBird45:
> I think this is going to require adding windows-gnu support to randomX* or adding windows-msvc to cryptonight-cuprate.

Take whatever is less work. For such algorithms that doesn't require specific platform API, you can take whatever you want. I think gnu is the way to go since some people on stackoverflow said it generated better quality code. Now if it is more easy to edit cryptonight ig we can go to msvc

Boog900:
Ok building works but we can't reach monerod on windows....

Boog900:
Ok it all works now.

When running tests Cuprate will now look in the `target` folder for monerod if it's not there it will download it, then it will spawn it with the flags specified.

@hinto-janai is there anything else you want to add or can I merge

hinto-janai:
Thanks for this, it did not look fun...

Can you add documentation explaining `test-utils/`? A small 1 line README is ok, I can fill it in eventually.

Boog900:
@hinto-janai done :)

hinto-janai:
OK for merge.

PR #56, opened 2024-02-10
helper: use `crossbeam::atomic::AtomicCell` for atomic floats

hinto-janai:
Adds `crossbeam` to the workspace's dependencies, and replaces `helper/`'s atomic float implementation with `crossbeam::atomic::AtomicCell<float>`.

As noted in https://github.com/Cuprate/cuprate/pull/45#issuecomment-1898503471 this:
- Is lock-free
- Allows `const fn new()`

A static assertion is also added to make sure targets that don't support lock-free atomic `32/64-bit` operations fail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,6 +370,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1137cd7e7fc0fb5d3c5a8678be38ec56e819125d8d7907411fe24ccb943faca8"
+dependencies = [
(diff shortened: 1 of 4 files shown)


Boog900:
What is the plan for these atomic floats, do you have something in mind @hinto-janai?

hinto-janai:
Off the top of my head it could be used as the sync percentage in `cuprated` so readers/writers and maybe the RPC server don't need to take a lock.

I don't have anything particular in mind though.

SyntheticBird45:
the sync percentage could be expressed as an integer too. I don't really know if float cpu operations are supposed to be slower than integer ones but it's free to add

Boog900 (approved):

PR #61, opened 2024-02-13
database: move `Pod` bound from `Table` -> `Key`

hinto-janai:
## Part 3
https://github.com/Cuprate/cuprate/pull/60 <- Previous

Will fix https://github.com/Cuprate/cuprate/pull/35#discussion_r1488316707.


Boog900:
The impl Key for DupKey should be:

```rust
impl<P, S> Key for DupKey<P, S>
where
    P: Key + Copy,
    S: Key + Copy,
{
    const DUPLICATE: bool = true;

    type Primary = P;
    type Secondary = S;
    #[inline]
    fn primary(self) -> Self::Primary {
        self.primary
    }
    #[inline]
    fn primary_secondary(self) -> (Self::Primary, Self::Secondary) {
        (self.primary, self.secondary)
    }
}
```

Boog900:
Actually the P and S should have bounds of `Pod` not `Key`

hinto-janai:
@Boog900 ready.

Boog900 (approved):

PR #62, opened 2024-02-13
database: Errors

hinto-janai:
## Part 4
https://github.com/Cuprate/cuprate/pull/61 <- Previous

- Adds to `cuprate_database`'s error types and impls conversions from `heed` & `sanakirja`'s errors
- Fixes some (de)serialization trait bounds for the backends


hinto-janai:
Ready for review. I'm not sure how some of these errors should be handled.

database/src/error.rs
```diff
@@ -38,10 +42,21 @@ pub enum InitError {
...
     ShuttingDown,
 
-    // TODO: this could be removed once we have all errors figured out.
+    /// A fatal error occurred.
+    ///
+    /// This is for errors that _should_ be unreachable
+    /// but we'd still like to panic gracefully.
+    ///
+    /// # Invariant
+    /// If this error is received, all(?) of Cuprate should shutdown.
```
SyntheticBird45, line 51: I'm doubtful about this one. 
The policy should be to crash whenever possible as we don't know if this bug is a related to an exploitation of an unsafe code. 
However, this is the database, fwiw, lmdb has never suffered from any serious vuln in its existence. Being able to restart the database on the fly would be cool but we're supposing the database is *invulnerable*. @Boog900 do you have thoughts on this one ?

Boog900: This comment is under `InitError`, I think it's ok to have errors like this for initialization. Although we will probably just panic if init fails, it would be good to not have the DB panic.

I do agree with this for `RuntimeError`, any fatal `RuntimeError` should just be a panic.

database/src/error.rs
```diff
@@ -2,61 +2,148 @@
...
+    KeyNotFound,
+
+    /// The database environment has reached
+    /// maximum memory map size, it must be
+    /// increased.
+    //
+    // TODO: `sanakirja` automatically resizes, `heed` does not.
+    // I guess this should be `unreachable!()` for `sanakirja`?
+    #[error("not enough space in database environment memory map")]
+    MapFull,
```
Boog900, line 90: We do need to handle this for `heed` as well, in the DB. 

As LMDB according to the docs here: https://docs.rs/heed/0.20.0-alpha.9/heed/struct.Env.html#method.resize only allows resizing when there are no other txs, we need a way for a RWTX to pause reader and for it to check that all readers are paused before resizing.

We would probably need a static `RWLock`, all reader will acquire a read lock and hold the guard for the duration of the tx, to do this we would have to create a wrapper type for heeds tx which includes the guard. Then heed writters will only obtain the write lock when they need to resize the env. This would be a little bit of overhead but I can't think of a better way

database/src/error.rs
```diff
@@ -2,61 +2,148 @@
...
+/// 2. (De)serialization
+///
+/// as `cuprate_database` upholds the invariant that:
+///
+/// 1. All tables exist
+/// 2. (De)serialization never fails
+#[derive(thiserror::Error, Debug)]
 pub enum RuntimeError {
-    // TODO: replace string with actual error type.
+    /// The given key already existed in the database.
```
Boog900, line 75: There are some errors here that I don't think should be passed back out of the DB, they could still be defined here but they should just panic.

database/src/error.rs
Boog900: I did put a comment here but it appeared twice, so I deleted one and it deleted both :)

I don't remember exactly what said but this comment if under `InitError` and I think it's ok to return this type of error on initialization.

For `RuntimeError` any fatal error should be a panic

database/src/error.rs
hinto-janai: Will implement next PR. We need some type of signal for `shutdown` too, the writer thread will be busy flushing to disk & hanging on [`Env::prepare_for_closing`](https://docs.rs/heed/0.20.0-alpha.9/heed/struct.Env.html#method.prepare_for_closing) so it won't be able to respond with `RuntimeError::ShuttingDown`.

Instead of `RuntimeError::MapFull` existing, the writer thread should just automatically resize when needed, right? From the perspective of the writer thread:

1. Receive write request
2. Execute transaction
3. If transaction returns resize error, resize and try again until success
4. Respond

database/src/error.rs
hinto-janai: Which ones specifically? Should they be removed and panic in the `From` impl?

database/src/error.rs
Boog900: What code needs to be ran on shutdown? can't we just leave it for heed and sanakirja's drop impl?

> Instead of RuntimeError::MapFull existing, the writer thread should just automatically resize when needed, right? From the perspective of the writer thread:

Yeah, step 3 should probably be bounded so we don't have an infinite resize issue if something goes wrong in the DB but that is the idea

database/src/error.rs
Boog900: `MapFull`, `PageFull`, `TxMustAbort`, `ReadersFull`, `Corrupt`, `Fatal`, `Unknown`.

IMO the only errors that should be given out from the DB are errors on specific requests i.e. `KeyNotFound` repeating this request with a valid key should then return ok.

Whereas with an error like `Corrupt` the request might have been a valid one but the DB is in an invalid state, if we were to return this error it could cause the rest of Cuprate to enter an invalid state, unless they too were to check DB errors which would cause parts of the DB to leak into places I don't want it to.

How exactly these are handled I don't mind, it would probably be better to remove them and just panic in the from impl (unless these can be fixed like `MapFull`).

Is `InvalidVersion` an invalid Cuprate DB version? If so I think this should be removed from the `RuntimeError` as well and just be an `InitError`

database/src/error.rs
SyntheticBird45: > We need some type of signal for shutdown too, the writer thread will be busy flushing to disk & hanging on [Env::prepare_for_closing](https://docs.rs/heed/0.20.0-alpha.9/heed/struct.Env.html#method.prepare_for_closing) so it won't be able to respond with RuntimeError::ShuttingDown.

iirc, reth implementation use a simple `Arc<AtomicBool>` in `Relaxed` ordering. This is sufficient imo, all reader threads drop out their env reference on Shutdown signal, and the write thread will eventually be freed.

hinto-janai: > What code needs to be ran on shutdown?

Just signaling that we're shutting down and flushing to disk I think? Both [`heed`](https://docs.rs/heed/0.20.0-alpha.9/src/heed/env.rs.html#321) and [`sanakirja`](https://docs.rs/sanakirja/latest/src/sanakirja/environment/mod.rs.html#73) don't seem to flush in their drop impls.

> iirc, reth implementation use a simple Arc\<AtomicBool\> in Relaxed ordering

They use `SeqCst`. It needs to be at least `Acquire+Release` or else readers may continue even after the writer has signaled it is shutting down. This does mean readers will `load()` an atomic bool each read, probably negligible in the grand scheme though.

SyntheticBird45: > Just signaling that we're shutting down and flushing to disk I think?

Maybe have a behavior so that while shutdown the tower::Service don't return poll_ready. But even then the database is the only primary components that need graceful shutdown. The rest could shutdown abruptly without any consequences.

> This does mean readers will load() an atomic bool each read, probably negligible in the grand scheme though.

This would be the cheapest way to do it. Using a channel would internally use an AtomicPtr anyway.

Boog900: Shutdown should only happen when both DB handles are dropped so we could just add a `shutdown` method to `Env` and impl `Drop` for `Env` just calling that:

```rust
impl<T: Env> Drop for T {
   fn drop(&self) {
      if !self.shutdown() {
          println!("Failed to shutdown the DB cleanly");
      }
   }
}
```

Boog900: ^that assumes `shutdown` returns a bool, it could return a Result

Boog900: but txs, when they are committed are flushed to disk, right? so there should be no situation where the DB is dropped without a tx getting submitted if we shutdown when all handles are dropped.

So I don't think it's necessary us doing custom shutdown logic unless we were to allow a caller to shutdown the DB without all handles being dropped (which I would rather not have)

hinto-janai: > but txs, when they are committed are flushed to disk, right?

Yes I think by default [LMDB flushes to disk per transaction commit](http://www.lmdb.tech/doc/group__mdb.html#ga85e61f05aa68b520cc6c3b981dba5037) but it'll be the slowest option, doubly bad for HDD. It would be equivalent to [`monerod --db-sync-mode=safe:sync`](https://monero.stackexchange.com/questions/8575/more-detailed-explanations-for-the-db-sync-mode-parameters-i-e-safe-fast-fa).

If we wanted something like `monerod --db-sync-mode=fastest:async` then we would use `MDB_NOSYNC` and only `sync()` at the very end.

We can also support both and everything in-between like `monerod`.

hinto-janai: `sanakirja` seems to flush per transaction commit too https://docs.rs/sanakirja/latest/src/sanakirja/environment/muttxn.rs.html#330-333.

Boog900: ah ok then yeah, custom shutdown code it is

hinto-janai:
Ready for review.

database/src/key.rs
```diff
@@ -20,28 +20,14 @@ pub trait Key {
...
-            type Secondary: Pod;
-        }
-    }
+    /// The primary key type.
+    type Primary: Pod;
+
+    /// The secondary key type.
+    ///
+    /// Only needs to be different than [`Key::Primary`]
+    /// if [`Key::DUPLICATE`] is `true`.
```
Boog900, line 29: These docs are no longer correct `Secondary` is `Infallible` now

Boog900 (approved):

PR #63, opened 2024-02-14
ci: only `cargo fmt` on 1 OS

hinto-janai:
Turns out `cargo fmt` checks all code, regardless of OS or `#[cfg]` options.


hinto-janai:
Added to https://github.com/Cuprate/cuprate/pull/32.

PR #64, opened 2024-02-14
helper: add `constants` & `COMMIT`

hinto-janai:


Boog900:
I think this should probably be in the `cuprate` binary, I don't think this will need to be used elsewhere

hinto-janai:
The RPC server could use it - `monerod` returns a commit in `get_info` (if non-release build).

It would also be used in any future binaries.

Boog900:
ah ok, is this ready to merge?

Boog900 (approved):

PR #65, opened 2024-02-14
Cleanup & Document consensus

Boog900:
This PR has turned into a pretty big clean up of the consensus code, specifically `cuprate-consensus` the crate which holds the services that handle verification. `cuprate-consensus-rules` should be mostly unchanged.

This PR also fixes some issues with our consensus code, regarding cross-block batch verification, which is batch verifying txs together from multiple blocks. 

I have decided to remove cross-block batch verification for now, in the future we can look for better solutions like staged syncing.

Sadly by not cross-block batch verifying we will lose out some performance when syncing, I think this is ok for now as we will have a fast sync mode where verification will be skipped for known block hashes which would a lot quicker anyway.

I have also decided to remove our RPC scanner, as although it was extremely quick for what it was doing, syncing a Cuprate node should be even faster so I can't see a reason to keep it around. Without cross-block batch verifying the RPC scanner would be extremely slow as well.


Boog900:
@hinto-janai This is ready for review.

I haven't finished documenting all of consensus but the diff is getting pretty large + this PR fixes issues in our consensus code so I would rather merge this now and finish the rest in a future PR.

Also I need to merge this before I unify the 2 database request/ response enums as this changes some of the variants.

consensus/src/context/task.rs
```diff
@@ -0,0 +1,232 @@
...
+    ) -> Result<ContextTask, ExtendedConsensusError>
+    where
+        D: Database + Clone + Send + Sync + 'static,
+        D::Future: Send + 'static,
+    {
+        let ContextConfig {
+            difficulty_cfg,
+            weights_config,
+            hard_fork_cfg,
+        } = cfg;
```
hinto-janai, line 68: Probably don't do this but this code made me wonder if this was possible, and it is:
```rust
    pub async fn init_context<D>(
        ContextConfig {
            difficulty_cfg,
            weights_config,
            hard_fork_cfg,
        }: ContextConfig,
        mut database: D,
    ) -> Result<ContextTask, ExtendedConsensusError>
    where
        D: Database + Clone + Send + Sync + 'static,
        D::Future: Send + 'static,
    {
        // already destructured
```

consensus/src/context/tokens.rs
```diff
@@ -1,3 +1,10 @@
+//! Tokens
+//!
+//! This module contains tokens which keep track of the validity of certain data.
+//! Currently there is 2 token:
+//! - [`ValidityToken`]
+//!
```
hinto-janai, line 6: Seems like there's only 1?

consensus/src/context/tokens.rs
```diff
@@ -7,39 +14,20 @@ pub struct ValidityToken {
...
 
 impl ValidityToken {
+    /// Creates a new [`ValidityToken`]
     pub fn new() -> ValidityToken {
         ValidityToken {
             token: CancellationToken::new(),
         }
     }
 
+    /// Returns if the data is still valid.
```
hinto-janai, line 24:
```suggestion
    /// Returns `true` if the data is still valid.
```

consensus/src/lib.rs
```diff
@@ -76,73 +75,120 @@ where
...
     pub vote: HardFork,
 
+    /// The blocks timestamp.
     pub timestamp: u64,
+    /// The blocks cumulative difficulty.
     pub cumulative_difficulty: u128,
 
+    /// The blocks weight.
     pub block_weight: usize,
+    /// The blocks long term.
```
hinto-janai, line 119:
```suggestion
    /// The blocks long term weight.
```

consensus/src/lib.rs
```diff
@@ -76,73 +75,120 @@ where
...
-    let tx_svc = transactions::TxVerifierService::new(database.clone());
-    let block_svc = block::BlockVerifierService::new(ctx_svc, tx_svc.clone(), database);
+    let tx_svc = TxVerifierService::new(database.clone());
+    let block_svc = BlockVerifierService::new(ctx_svc, tx_svc.clone(), database);
     Ok((block_svc, tx_svc))
 }
 
+/// An internal trait used to represent a database so we don't have to write [`tower::Service`] bounds
+/// everywhere.
 pub trait Database:
```
hinto-janai, line 85: "internal" has the idea of being `pub(crate)` or similar, I think a different word would be better here. "Type alias trait" or something.

Could also explain the trait and what it represents.

consensus/src/lib.rs
```diff
@@ -76,73 +75,120 @@ where
...
     pub block_weight: usize,
+    /// The blocks long term.
     pub long_term_weight: usize,
 }
 
+/// A database request to the database [`tower::Service`]
 #[derive(Debug, Clone)]
 pub enum DatabaseRequest {
+    /// A block extended header request.
+    /// Must return: [`DatabaseResponse::BlockExtendedHeader`]
```
hinto-janai, line 127: There's already docs on `cuprate_types::service::{Response,Request}`, maybe copy those?

Btw, cargo docs will render this as a single line, there needs to be a space between lines.

Actually, what is the plan here? Why not use/edit the `Request/Response` from `types`?

consensus/src/lib.rs
```diff
@@ -76,73 +75,120 @@ where
...
 }
 
 #[derive(Debug)]
 pub enum DatabaseResponse {
+    /// A block extended header response.
     BlockExtendedHeader(ExtendedBlockHeader),
+    /// A block hash response.
     BlockHash([u8; 32]),
 
+    FilteredHashes(HashSet<[u8; 32]>),
```
hinto-janai, line 168:
```suggestion
    /// TODO
    FilteredHashes(HashSet<[u8; 32]>),
```

consensus/src/transactions.rs
```diff
@@ -1,94 +1,104 @@
...
+//!
+//! This module contains the [`TxVerifierService`] which handles consensus validation of transactions.
+//!
 use std::{
     collections::HashSet,
     future::Future,
     ops::Deref,
     pin::Pin,
-    sync::Arc,
+    sync::{Arc, Mutex as StdMutex},
```
hinto-janai, line 10: Why `Mutex` -> `StdMutex`?

consensus/src/transactions.rs
```diff
@@ -1,94 +1,104 @@
...
+    /// Both semantic validation and contextual validation are needed.
+    SemanticAndContextual,
+    /// Only contextual validation is needed.
+    Contextual,
+}
 
-    Ok(txs)
+/// Represents if a transaction has been fully validated and under what conditions
+/// the transaction is valid in the future.
+#[derive(Debug, Clone)]
```
hinto-janai, line 51: Could be `Ord` if `HardFork` did the same.
```suggestion
#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
```

consensus/src/transactions.rs
```diff
@@ -101,42 +111,62 @@ impl TransactionVerificationData {
...
-        txs: Vec<Arc<TransactionVerificationData>>,
+    /// Verifies a batch of prepared txs.
+    Prepped {
+        /// The transactions to verify.
+        // TODO: Can we use references to remove the outer `Arc`? probably wont play nicely with rayon_spawn_async though
+        txs: Arc<[Arc<TransactionVerificationData>]>,
+        /// The current chain height.
         current_chain_height: u64,
+        /// The top hash.
+        top_hash: [u8; 32],
```
hinto-janai, line 137: Hash of a block? or tx?

consensus/src/transactions/contextual_data.rs
```diff
@@ -10,166 +10,36 @@
...
 ///
 /// This function batch gets all the ring members for the inputted transactions and fills in data about
 /// them.
-pub async fn batch_fill_ring_member_info<'a, D: Database + Clone + Send + Sync + 'static>(
-    txs_verification_data: impl Iterator<Item = &Arc<TransactionVerificationData>> + Clone,
-    hf: &HardFork,
-    re_org_token: ReOrgToken,
+pub async fn batch_get_ring_member_info<'a, 'b, D: Database>(
+    txs_verification_data: impl Iterator<Item = &'a Arc<TransactionVerificationData>> + Clone,
+    hf: &'b HardFork,
```
hinto-janai, line 40: Why does `hf` need a lifetime? `HardFork` is `Copy` (and just a `u8`)

consensus/src/transactions/contextual_data.rs
```diff
@@ -197,38 +67,88 @@ pub async fn batch_fill_ring_member_info<'a, D: Database + Clone + Send + Sync +
...
+/// Refreshes the transactions [`TxRingMembersInfo`], if needed.
+#[instrument(level = "debug", skip_all)]
+pub async fn batch_get_decoy_info<'a, 'b, D: Database + Clone + Send + 'static>(
+    txs_verification_data: &'a [Arc<TransactionVerificationData>],
+    hf: &'b HardFork,
+    mut database: D,
+) -> Result<
+    impl Iterator<Item = Result<DecoyInfo, ConsensusError>> + Captures<(&'a (), &'b ())>,
+    ExtendedConsensusError,
+> {
```
hinto-janai, line 105: Same here, could get rid of the `Captures` hack?

consensus/src/transactions/contextual_data.rs
```diff
@@ -197,38 +67,88 @@ pub async fn batch_fill_ring_member_info<'a, D: Database + Clone + Send + Sync +
...
+                None
+            };
+
+            TxRingMembersInfo::new(ring_members_for_tx, decoy_info, tx_v_data.version)
+                .map_err(ConsensusError::Transaction)
+        })
+        .collect::<Result<_, _>>()?)
 }
+
+/// Refreshes the transactions [`TxRingMembersInfo`], if needed.
```
hinto-janai, line 96:
```suggestion
/// Refreshes the transactions [`TxRingMembersInfo`], if needed.
///
/// # Panics
/// This functions panics if `hf == HardFork::V1` as decoy info
/// should not be needed for V1.
```

SyntheticBird45:
Not in the PR changes but I think big fn like `verify_prepped_transactions`, `transactions_needing_verification`,  in `src/transactions.rs` would benefit from comments in the middle to keep track of each step

consensus/src/context/difficulty.rs
```diff
@@ -166,30 +169,24 @@ impl DifficultyCache {
...
 
         for (new_timestamp, hf) in blocks {
-            self.timestamps.push_back(new_timestamp);
-            self.cumulative_difficulties
-                .push_back(self.cumulative_difficulty() + *difficulties.last().unwrap());
-            if u64::try_from(self.timestamps.len()).unwrap() > self.config.total_block_count() {
+            timestamps.push_back(new_timestamp);
+
+            let last_cum_diff = cumulative_difficulties.back().copied().unwrap_or(1);
+            cumulative_difficulties.push_back(last_cum_diff + *difficulties.last().unwrap());
```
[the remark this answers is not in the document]
Boog900: I think these are pretty much the same right?

consensus/src/context/task.rs
Boog900: ah interesting!

consensus/src/context/tokens.rs
Boog900: changed, there was 2

consensus/src/transactions.rs
Boog900: I'm not going to make this Ord as I don't think it really makes sense

consensus/src/transactions.rs
Boog900: to make it clear it's not an async mutex

consensus/src/lib.rs
Boog900: In another PR these will be replaced with the enums in `cuprate_types` I didn't want to do it here as the DB would need updating

consensus/src/lib.rs
Boog900: Will be documented in a future PR

consensus/src/lib.rs
Boog900: wIll update in the PR where the request.response enums are unified.

Boog900:
> Not in the PR changes but I think big fn like verify_prepped_transactions, transactions_needing_verification, in src/transactions.rs would benefit from comments in the middle to keep track of each step 

Will do in a future PR

consensus/src/context/rx_vms.rs
```diff
@@ -1,3 +1,8 @@
+//! RandomX VM Cache
+//!
+//! This module keeps track of the RandomX VM to calculate the next blocks PoW, if the block needs a randomX VM and potentially
```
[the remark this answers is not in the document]
Boog900: The cache will always have the VM to calculate the next blocks POW _and_ potentially other VMs around that height so I think is and is correct

consensus/src/context/difficulty.rs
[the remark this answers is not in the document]
hinto-janai: Was thinking it would be slower since `.copied()` copies the full `Option<u128>` (size_of = 32 instead of 16) although it seems like the extra ref indirection may be worse:
```rust
use std::hint::black_box as bb;
use std::{collections::VecDeque, time::Instant};

fn main() {
    loop {
        let mut cumulative_difficulties = VecDeque::from([0]);
        let difficulties = vec![0];

        let now = Instant::now();
        for _ in 0..1_000_000_000 {
            // let _ = bb(old(&mut cumulative_difficulties, &difficulties));
            let _ = bb(new(&mut cumulative_difficulties, &difficulties));
        }
        println!("{}", now.elapsed().as_secs_f32());
    }
}

fn old(cumulative_difficulties: &mut VecDeque<u128>, difficulties: &Vec<u128>) {
    let last_cum_diff = bb(cumulative_difficulties.back().copied().unwrap_or(1));
    cumulative_difficulties.push_back(last_cum_diff + *difficulties.last().unwrap());
}

fn new(cumulative_difficulties: &mut VecDeque<u128>, difficulties: &Vec<u128>) {
    let last_cum_diff = bb(cumulative_difficulties.back().unwrap_or(&1));
    cumulative_difficulties.push_back(last_cum_diff + difficulties.last().unwrap());
}
```

hinto-janai (approved):

PR #66, opened 2024-02-16
imporove Monerod spawn stability 

Boog900:
This PR removes the ability for tests to share monerod's, and makes it so if a test fails the monerod spawned for that test will output its logs.

### DIFF
--- a/test-utils/src/monerod.rs
+++ b/test-utils/src/monerod.rs
@@ -6,19 +6,18 @@
 use std::{
     ffi::OsStr,
     io::Read,
-    net::{IpAddr, Ipv4Addr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener},
     process::{Child, Command, Stdio},
     str::from_utf8,
     thread::panicking,
     time::Duration,
 };
 
-use rand::Rng;
 use tokio::{task::yield_now, time::timeout};
 
 mod download;
 
-const LOCAL_HOST: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+const LOCALHOST: IpAddr = IpAddr::V4(Ipv4Addr::LOCALHOST);
 const MONEROD_VERSION: &str = "v0.18.3.1";
 const MONEROD_STARTUP_TEXT: &str =
     "The daemon will start synchronizing with the network. This may take a long time to complete.";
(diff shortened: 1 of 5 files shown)


Boog900:
example of an output from a failed test: https://github.com/Cuprate/cuprate/actions/runs/7934176577/job/21664554644#step:11:420

Boog900:
@hinto-janai @SyntheticBird45 this is ready

test-utils/src/monerod.rs
```diff
@@ -4,169 +4,139 @@
...
-        tokio::task::spawn(manager.run(rx));
-    }
+pub async fn monerod<T: AsRef<OsStr>>(flags: impl IntoIterator<Item = T>) -> SpawnedMoneroD {
+    let path_to_monerod = download::check_download_monerod().await.unwrap();
+    let mut rng = rand::thread_rng();
+
+    // Use random ports and *hope* we don't get a collision (TODO: just keep an atomic counter and increment?)
+    let rpc_port: u16 = rng.gen_range(1500..u16::MAX);
+    let p2p_port: u16 = rng.gen_range(1500..u16::MAX);
+    let zmq_port: u16 = rng.gen_range(1500..u16::MAX);
```
hinto-janai, line 38:
```rust
fn get_available_port(already_taken: &[u16]) -> u16 {
    loop {
        // Using `0` makes the OS return a random available port.
        let port = std::net::TcpListener::bind("127.0.0.1:0")
            .unwrap()
            .local_addr()
            .unwrap()
            .port();

        if !already_taken.contains(&port) {
            return port;
        }
    }
}

let rpc_port = get_available_port(&[]);
let p2p_port = get_available_port(&[rpc_port]);
let zmq_port = get_available_port(&[rpc_port, p2p_port]);
```

test-utils/src/monerod.rs
```diff
@@ -4,169 +4,139 @@
...
 use rand::Rng;
-use tokio::{
-    process::{Child, Command},
-    sync::{mpsc, oneshot},
-};
+use tokio::{task::yield_now, time::timeout};
 
 mod download;
 
 const LOCAL_HOST: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
```
hinto-janai, line 21: `std::net::Ipv4Addr::LOCALHOST`?

test-utils/src/monerod.rs
hinto-janai: Ah, but there are multiple `monerod`'s... `already_taken` can be `Arc<Mutex<Vec<u16>>` and updated for all I guess.

Boog900: The OS will see that the ports are taken unless they are started at the same time, I think the chance of this happening is probably low enough to just go with your first idea

hinto-janai (approved):

PR #77, opened 2024-02-25
database: docs & small fixes for `Config`

hinto-janai:
## Part 6
https://github.com/Cuprate/cuprate/pull/68 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/79

Documents and fixes/adds some things to `cuprate_database::config`.

### DIFF
--- a/database/src/config.rs
+++ b/database/src/config.rs
@@ -308,9 +308,18 @@ pub enum ReaderThreads {
     /// | `ReaderThreads::Percent(2.0)`      | 32 (saturating)
     /// | `ReaderThreads::Percent(f32::NAN)` | 32 (non-normal default)
     ///
-    /// # `0`
+    /// # `0.0`
     /// `ReaderThreads::Percent(0.0)` represents "use maximum value",
     /// as such, it is equal to [`ReaderThreads::OnePerThread`].
+    ///
+    /// # Not quite `0.0`
+    /// If the thread count multiplied by the percentage ends up being
+    /// non-zero, but not 1 thread, the minimum value 1 will be returned.
+    ///
+    /// ```rust
+    /// # use cuprate_database::config::*;
+    /// assert_eq!(ReaderThreads::Percent(0.000000001).as_threads().get(), 1);
+    /// ```
     Percent(f32),
 }
 
@@ -403,6 +412,8 @@ impl ReaderThreads {
 
 impl<T: Into<usize>> From<T> for ReaderThreads {
     /// Create a [`ReaderThreads::Number`].
+    ///
+    /// If `value` is `0`, this will return [`ReturnThreads::OnePerThread`].
     fn from(value: T) -> Self {
         let u: usize = value.into();
         if u == 0 {
--- a/database/src/config.rs
+++ b/database/src/config.rs
@@ -172,7 +172,7 @@ impl Default for Config {
 )]
 pub enum SyncMode {
     /// Use [`SyncMode::Fast`] until fully synced,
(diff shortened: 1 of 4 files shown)


hinto-janai:
Ready for review.

database/src/config.rs
```diff
@@ -16,151 +26,355 @@ use crate::{constants::DATABASE_FILENAME, resize::ResizeAlgorithm};
...
+/// Regardless of the variant chosen, dropping [`Env`](crate::Env)
+/// will always cause it to fully sync to disk.
+///
+/// # Sync vs Async
+/// All invariants except [`SyncMode::Fast`] are `synchronous`,
+/// as in the database will wait until the sync is finished before continuing.
+///
+/// `SyncMode::Fast` is `asynchronous`, meaning it will _NOT_
+/// wait until the sync is done before continuing. It will immediately
+/// move onto the next operation even if the database/OS has not responded.
```
Boog900, line 183: This would not be safe, when a write happens a new reader must be able to get the new data. This means we would need the DB to respond to the writer so we know that when we call a write, when that call finishes we can open a new read tx and get this data.

All these config options should be doing is changing how we open the Env

database/src/config.rs
```diff
@@ -16,151 +26,355 @@ use crate::{constants::DATABASE_FILENAME, resize::ResizeAlgorithm};
...
 )]
 pub enum SyncMode {
-    /// Fully sync to disk per transaction.
+    /// Use [`SyncMode::Fast`] until fully synced,
+    /// then use [`SyncMode::Safe`].
+    ///
+    /// TODO: We could not bother with this and just implement
+    /// batching, which was the solution to slow syncs with `Safe`.
+    /// ref: <https://github.com/monero-project/monero/issues/1463>
+    /// solution: <https://github.com/monero-project/monero/pull/1506>
```
Boog900, line 197: Just noting that we will be batching blocks together but I still think this would be a good option to have.

database/src/config.rs
hinto-janai: Sorry, these docs were a little misleading...

This variant will just open LMDB with [`MDB_NOSYNC`](http://www.lmdb.tech/doc/group__mdb__env.html#ga5791dd1adb09123f82dd1f331209e12e) and never call `sync()`. This matches `fastest:async` mode in `monerod`. The readers will still be able to see writes regardless of this, it's referring to _disk_ syncs, not the database data itself so:

```rust
db.put("key", value);
db.get("key"); // <- will get value even with `Fast`, probably from memory directly
```

database/src/config.rs
hinto-janai: There are some open questions on how to implement this: https://github.com/Cuprate/cuprate/issues/78

database/src/config.rs
```diff
@@ -16,151 +26,365 @@ use crate::{constants::DATABASE_FILENAME, resize::ResizeAlgorithm};
...
+///
+/// `SyncMode::Fast` is `asynchronous`, meaning the database will _NOT_
+/// wait until the data is fully synced to disk before continuing.
+/// Note that this doesn't mean the database itself won't be synchronized
+/// between readers/writers, but rather that the data _on disk_ may not
+/// be immediately synchronized after a write.
+///
+/// Something like:
+/// ```rust,ignore
+/// db.get("key", value);
```
Boog900, line 190:
```suggestion
/// db.put("key", value);
```

Boog900 (approved):

PR #79, opened 2024-02-26
database: replace `sanakirja` with `libmdbx`

hinto-janai:
## Part 7
https://github.com/Cuprate/cuprate/pull/77 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/80

- Removes the `sanakirja` backend, documents why
- Adds [`libmdbx`](https://docs.rs/libmdbx) and implements the `cuprate_database` traits (just type checks, no body)
- Actually implements `libmdbx::Error` -> `cuprate_database::RuntimeError`

@Boog900 Some notes:
- `libmdbx-rs` (the bindings) are [MPL-2.0](https://github.com/vorot93/libmdbx-rs/blob/master/LICENSE)
- We need to fork and apply the same DUPKEY changes to `libmdbx-rs`

PR #80, opened 2024-02-27
database: replace `sanakirja` with `redb`

hinto-janai:
## Part 8
https://github.com/Cuprate/cuprate/pull/79 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/81

- Removes the `sanakirja` backend, documents why
- Adds [`redb`](https://docs.rs/redb) and implements the `cuprate_database` traits (just type checks, no body)
- Actually implements `redb::Error` -> `cuprate_database::RuntimeError`

## Traits
The core traits were also changed to account for `redb` abstracting RW operations at the table level (with an associated TX lifetime) compared to `heed` abstracting that at the TX level.

- `trait Database` -> `trait DatabaseRo`, `trait DatabaseRw: DatabaseRo`
- `Env::open_database()` -> `Env::open_db_r{o,w}()`
-  Added `HeedTableR{o,w}` that mimics `redb`'s [`ReadOnlyTable`](https://docs.rs/redb/1.5.0/redb/struct.ReadOnlyTable.html) & [`Table`](https://docs.rs/redb/1.5.0/redb/struct.Table.html)

Now, instead of `trait Database` taking a TX as input every function, the struct that actually implements `trait DatabaseR{o,w}` will need to hold onto the transaction somehow. This is [already the case](https://docs.rs/redb/1.5.0/src/redb/table.rs.html#61) in `redb`'s table types, and is mimicked in our custom `HeedTableR{o,w}` types.

We should still be able to reuse TX's since `HeedTableR{o,w}` only take references:
```rust
let env: Env = todo!();
let mut tx: TxRw = env.tx_rw();

{
    let db: DatabaseRw = env.open_db_rw::<Table>(&mut tx);
    /* do db stuff */
}

{
    // We still have access to `tx`.
    let db2: DatabaseRw = env.open_db_rw::<Table2>(&mut tx);
    /* do db stuff */
}

// Re-used the `tx` for 2 table operations :)
tx.commit();
```

## `SyncMode`
For now, all backends will only support:
- `SyncMode::Safe`
- `SyncMode::Async` (added in this PR)
- `SyncMode::Fast`

as they are easy to implement (just set once in `Env::open`).

Dynamic syncs are more difficulty. `redb` also abstracts it at the TX level, where as `heed` abstracts it at the `Env` level. `heed` does not expose `mdb_env_set_flags` which is needed to change sync modes dynamically (without re-opening the `Env`), so dynamic syncs will be implemented in some later PR.


hinto-janai:
@Boog900 ready.

database/README.md
```diff
@@ -140,10 +142,37 @@ cargo doc
...
+[`sanakirja`](https://docs.rs/sanakirja) was a candidate as a backend, however there were problems with maximum value sizes.
+
+The default maximum value size is [1012 bytes](https://docs.rs/sanakirja/1.4.1/sanakirja/trait.Storable.html) which was too small for our requirements. Using [`sanakirja::Slice`](https://docs.rs/sanakirja/1.4.1/sanakirja/union.Slice.html) and [sanakirja::UnsizedStorage](https://docs.rs/sanakirja/1.4.1/sanakirja/trait.UnsizedStorable.html) was attempted, but there were bugs found when inserting a value in-between `512..=4096` bytes.
+
+As such, it is not implemented.
+
+## `MDBX`
+[`MDBX`](https://erthink.github.io/libmdbx) was a candidate as a backend, however MDBX deprecated the custom key/value comparison functions needed for `cuprate_database`'s duplicate table requirements. It is also quite similar to the main backend LMDB (of which it was originally a fork of).
+
+As such, it is not implemented.
```
Boog900, line 175: It is possible to implement MDBX, just not as easily as LMDB so I would change this to:

```suggestion
[`MDBX`](https://erthink.github.io/libmdbx) was a candidate as a backend, however MDBX deprecated the custom key/value comparison functions, this makes it a bit trickier to implement dup tables. It is also quite similar to the main backend LMDB (of which it was originally a fork of).

As such, it is not implemented (yet).
```

Boog900 (approved):

PR #81, opened 2024-02-29
database: replace `Pod` with `bytemuck`

hinto-janai:
## Part 9
https://github.com/Cuprate/cuprate/pull/80 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/85

- Replaces `cuprate_database::Pod` trait bounds with `bytemuck`'s traits
- Adds `cuprate_database::types` for database schema types


Boog900:
I have been thinking about this a bit and I think we will still need a custom trait for DB encodable types, as we need to support variable length types i.e. [u64] for output indexes. 

The trait could look something like:

```rust
trait DBStorable {
    fn as_bytes(&self) -> &[u8];

    fn from_bytes(bytes: &[u8]) -> &Self;
}
```

With an impl for any type that is `NoUninit + AnyBitPattern`:
```rust
use bytemuck::{bytes_of, from_bytes, AnyBitPattern, NoUninit};


impl<T: NoUninit + AnyBitPattern> DBStorable for T {
    fn as_bytes(&self) -> &[u8] {
        bytes_of(self)
    }

    fn from_bytes(bytes: &[u8]) -> &Self {
        from_bytes(bytes)
    }
}
```

We also need an impl for a slice but because of Rusts orphan Rules we need to do this:
```rust
use ref_cast::RefCast;
use bytemuck::{cast_slice, AnyBitPattern, NoUninit};

#[derive(RefCast, Debug)]
#[repr(transparent)]
struct Slice<T: NoUninit + AnyBitPattern>([T]);

impl<T: NoUninit + AnyBitPattern> Deref for Slice<T> {
    type Target = [T];

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl<T: NoUninit + AnyBitPattern> DBStorable for Slice<T> {
    fn as_bytes(&self) -> &[u8] {
        cast_slice(&self.0)
    }

    fn from_bytes(bytes: &[u8]) -> &Self {
        Slice::ref_cast(cast_slice(bytes))
    }
}

impl<'a, T: NoUninit + AnyBitPattern> From<&'a [T]> for &'a Slice<T> {
    fn from(value: &'a [T]) -> Self {
        RefCast::ref_cast(value)
    }
}
```

now the `Table` trait would look something like:

```rust
trait Table {
    type Key: Key;
    type Value: DBStorable + ?Sized; //  Neither keys types (P|S) need to add ?Sized only the `Value` needs to
}
```
---

Doing this does prevent doing dup tables how monerod does, so we have to go with option 2 that I discussed in the readme here: https://github.com/Cuprate/cuprate/pull/59/commits/41feda484b6619d487dab00492e6c54ab25727d5

What this means is we need to store the output amount for every pre-rct output instead of it only being stored once. I don't think this will be too much extra data and may even make it quicker to get a pre-rct output - we don't need to search the amounts then the amount indexes, instead only doing one search. 

It also means we don't need a custom heed fork.

Using the second method we can only have dup keys with a fixed secondary type (`u64`) as to make looking up the number of entries per primary key work. This is not an issue though.

So our `Key` trait will become:

```rust
pub trait Key: DBStorable {
    const DUPLICATE: bool;

    type Primary: DBStorable;

    fn primary(self) -> Self::Primary;

    fn primary_secondary(self) -> (Self::Primary, u64);
}
```

For non-dup tables Primary and Key is the same, for dup tables `Key` is the primary and secondary (u64) combination one after the other. This means we need a different dup key type per dup key table as bytemuck does not allow generics for `Pod`.

Boog900:
For `heed` you could also do something like:

```rust
struct HeedSerde<T: DBStorable + ?Sized>(PhantomData<T>);

impl<'a, T: DBStorable + ?Sized + 'a> BytesDecode<'a> for HeedSerde<T> {
    type DItem = &'a T;

    fn bytes_decode(bytes: &'a [u8]) -> Result<Self::DItem, BoxedError> {
        Ok(T::from_bytes(bytes))
    }
}

impl<'a, T: DBStorable + ?Sized + 'a> BytesEncode<'a> for HeedSerde<T> {
    type EItem = T;

    fn bytes_encode(item: &'a Self::EItem) -> Result<Cow<'a, [u8]>, BoxedError> {
        Ok(Cow::Borrowed(item.as_bytes()))
    }
}
```

So then a heed DB would be `Database<HeedSerde<Key>, HeedSerde<Value>>` instead of every DB being just bytes

Boog900:
for redb I guess something similar would work?

hinto-janai:
Ready, changes:

- Replaces `cuprate_database::Pod` with `cuprate_databaes::Storable`, implemented automatically by any `T` that implements bytemuck's traits
- Adds `cuprate_database::types` for database schema types (which impl `Storable`)
- Added `src/backend/${BACKEND}/storable.rs` for a compatibility layer so any `T: Storable` can be (de)serialized by the backend's traits automatically (added working tests too)
- Change backend table/database types to use the compatibility layer/struct (`StorableHeed<T>` and `StorableRedb<T>`) instead of bytes. Solely a type compat layer, it only contains `PhantomData<T>`
- Added `Key::CUSTOM_COMPARE`, `Key::compare()`, `Key::new_with_max_secondary()`
- `DatabaseRo::get()` returns a `&Table::Value` instead of `Table::Value`

Misc stuff:
- `Slice<T>` was not implemented as we can `impl Storable for [u8]` directly
- `bytemuck::TransparentWrapper` is used instead of `ref_cast`
- Switch from forked `heed` -> `heed 0.20.0-alpha.9`

database/src/backend/redb/database.rs
```diff
@@ -9,38 +9,44 @@ use crate::{
 };
 
 //---------------------------------------------------------------------------------------------------- DatabaseRo
-impl<T: Table> DatabaseRo<T> for RedbTableRo<'_> {
-    fn get(&self, key: &T::Key) -> Result<Option<T::Value>, RuntimeError> {
+impl<T: Table> DatabaseRo<T> for RedbTableRo<'_, T::Key, T::Value> {
+    fn get(&self, key: &T::Key) -> Result<Option<&T::Value>, RuntimeError> {
```
Boog900, line 13: I just realized this, I don't think we need the `Result<Option<_>,_>` return value only the result as `RuntimeError` contains an error for `KeyNotFound`

database/src/table.rs
```diff
@@ -13,25 +13,31 @@ use crate::{key::Key, pod::Pod};
...
-    // how this table will be stored.
-    //
-    // type KeyComparator: fn(&Self::Key, &Self::Key) -> Ordering;
-    // type ValueComparator: fn(&Self::Value, &Self::Value) -> Ordering;
-
     /// Name of the database table.
     const NAME: &'static str;
 
     /// Whether the table's values are all the same size or not.
     const CONSTANT_SIZE: bool;
```
Boog900, line 20: this can probably be removed now we have put this on the `Storable` trait

Boog900 (changes requested):

Boog900 (approved):

PR #83, opened 2024-02-29
CI: fix deny

Boog900:
I regenerated the file so we don't get depreciated warnings, I think I kept most options the same.

@hinto-janai

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -11,6 +11,9 @@
 
 # Root options
 
+# The graph table configures how the dependency graph is constructed and thus
+# which crates the checks are performed against
+[graph]
 # If 1 or more target triples (and optionally, target_features) are specified,
 # only the specified targets will be checked when running `cargo deny check`.
 # This means, if a particular package is only ever used as a target specific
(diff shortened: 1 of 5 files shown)


deny.toml
```diff
@@ -64,35 +70,18 @@ feature-depth = 1
...
-# The lint level for security vulnerabilities
-vulnerability = "deny"
-# The lint level for unmaintained crates
-unmaintained = "warn"
-# The lint level for crates that have been yanked from their source registry
-yanked = "deny"
-# The lint level for crates with security notices. Note that as of
-# 2019-12-17 there are no security notice advisories in
-# https://github.com/rustsec/advisory-db
-notice = "warn"
```
hinto-janai, line 80: Removed accidentally? Are these the default when running `cargo deny` anyway?

deny.toml
```diff
@@ -103,61 +92,37 @@ ignore = [
 # More documentation for the licenses section can be found here:
 # https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
 [licenses]
-# The lint level for crates which do not have a detectable license
-unlicensed = "deny"
```
hinto-janai, line 107: Keep this, no license basically means [we can't use it.](https://choosealicense.com/no-permission/)

deny.toml
```diff
@@ -103,61 +92,37 @@ ignore = [
...
-# * fsf - The license will be approved if it is FSF Free
-# * osi-only - The license will be approved if it is OSI-approved *AND NOT* FSF
-# * fsf-only - The license will be approved if it is FSF *AND NOT* OSI-approved
-# * neither - This predicate is ignored and the default lint level is used
-allow-osi-fsf-free = "neither"
-# Lint level used when no other predicates are matched
-# 1. License isn't in the allow or deny lists
-# 2. License isn't copyleft
-# 3. License isn't OSI/FSF, or allow-osi-fsf-free = "neither"
-default = "deny"
```
hinto-janai, line 160: Removed accidentally?

deny.toml
Boog900: individual setting has been depreciated, they are all deny now.

deny.toml
Boog900: same reason here as previous comment

hinto-janai (approved):

PR #84, opened 2024-03-02
levin: fragmented messages 

Boog900:


Boog900:
This fixes our fragmented bucket reading and adds support for sending fragmented buckets.

A fragmented bucket is just a message that has been split into parts that can be sent at fixed intervals to hide when data is actually being sent i.e. a Tor connection could every N seconds send a dummy message (a message which will be ignored) and when it needs to send actual data it would split this data up to be the same size as the dummies and send them instead of the dummies every N seconds until the message is sent.

@hinto-janai

Boog900:
I copied our handshake tests and created a new net zone `FragNet` that sends every message as fragments to test compatibility with monerod.

net/levin/src/header.rs
```diff
@@ -16,34 +16,29 @@
...
-        self.0 & START_FRAGMENT != 0
-    }
-    pub fn is_end_fragment(&self) -> bool {
-        self.0 & END_FRAGMENT != 0
+        const START_FRAGMENT = 0b0000_0100;
+        const END_FRAGMENT = 0b0000_1000;
+
+        const DUMMY = Self::START_FRAGMENT.bits() | Self::END_FRAGMENT.bits();
+
+        const _ = !0;
```
hinto-janai, line 41: Docs?

net/levin/src/lib.rs
```diff
@@ -33,16 +33,20 @@
...
+pub use message::LevinMessage;
 
-use std::fmt::Debug;
-
-use bytes::{Buf, Bytes};
-use thiserror::Error;
+use header::Flags;
 
 const MONERO_PROTOCOL_VERSION: u32 = 1;
 const MONERO_LEVIN_SIGNATURE: u64 = 0x0101010101012101;
```
hinto-janai, line 52: I know this isn't a diff but docs would be nice here too.

net/levin/src/lib.rs
```diff
@@ -128,14 +132,14 @@ impl MessageType {
...
     pub fn from_flags_and_have_to_return(
-        flags: header::Flags,
+        flags: Flags,
         have_to_return: bool,
     ) -> Result<Self, BucketError> {
-        if flags.is_request() && have_to_return {
+        if flags.contains(Flags::REQUEST) && have_to_return {
             Ok(MessageType::Request)
-        } else if flags.is_request() {
+        } else if flags.contains(Flags::REQUEST) {
```
hinto-janai, line 140: Compiler probably does it anyway but you could store/reuse the `flags.contains(Flags::REQUEST)`.

net/levin/src/message.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// A message body.
+    /// A levin header will be added to this message before it is sent to the peer.
+    Body(T),
+    /// A full levin bucket.
+    /// This bucket will be sent to the peer directly with no extra information.
+    /// This should only be used to send fragmented messages: [`make_fragmented_messages`]
+    Bucket(Bucket<T::Command>),
+    /// A Dummy message
+    /// A dummy message which the peer will ignore. The dummy message will be the exact size
+    /// (in bytes) of the given `usize` on the wire.
```
hinto-janai, line 24: These won't be spaced out in `cargo docs` btw, it'll be on a single line.

It needs to be like this:
```
/// Line 1
///
/// Line 2
```

net/levin/src/message.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
+/// This represents a dummy message to send to a peer.
+///
+/// The message, including the header, will be the exact size of the given `usize`.
+// This exists because it seems weird to do this:
+// peer.send(1_000);
+// This is a lot clearer:
+// peer.send(Dummy(1_000));
```
hinto-janai, line 46:
```suggestion
/// This represents a dummy message to send to a peer.
///
/// The message, including the header, will be the exact size of the given `usize`.
/// This exists because it seems weird to do this:
/// ```rust,ignore
/// peer.send(1_000);
/// ```
/// This is a lot clearer:
/// ```rust,ignore
/// peer.send(Dummy(1_000));
/// ```
```

net/monero-wire/Cargo.toml
```diff
@@ -6,6 +6,9 @@ license = "MIT"
 authors = ["Boog900"]
 repository = "https://github.com/SyntheticBird45/cuprate/tree/main/net/monero-wire"
 
+[features]
+default = []
+tracing = ["levin-cuprate/tracing"]
```
hinto-janai, line 11: Not related to this PR but we should set all of our crates to `cuprate-*` eventually.

p2p/monero-p2p/tests/fragmented_handshake.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn encode(
+        &mut self,
+        item: LevinMessage<Message>,
+        dst: &mut BytesMut,
+    ) -> Result<(), Self::Error> {
+        match item {
+            LevinMessage::Body(body) => {
+                // 66 is the minimum fragment size.
+                let fragments = make_fragmented_messages(&Protocol::default(), 66, body).unwrap();
```
hinto-janai, line 122: Is this/should this be in a const somewhere?

hinto-janai:
Btw I don't think I can do more than surface-level reviews on diffs until I understand all of the original code (I don't) so passing CI is probably a more helpful signal than me.

net/monero-wire/Cargo.toml
Boog900: The naming convention I have been using is: `*-cuprate` when a crate is not specific to Cuprate and `cuprate-*` when a crate is specific to Cuprate. 

I could rename this to `levin-codec` later, which would probably be clearer. I think this is the only crate which is `*-cuprate`.

p2p/monero-p2p/tests/fragmented_handshake.rs
Boog900: This is just a test and no one should be using fragments this small for real messages, monerod uses 3_000.

Boog900:
yeah that's alright

net/levin/src/header.rs
```diff
@@ -30,12 +30,30 @@ pub struct Flags(u32);
...
+        /// a request or notification.
         const REQUEST = 0b0000_0001;
+        /// The response flags.
+        ///
+        /// Messages with this set are responses to requests.
         const RESPONSE = 0b0000_0010;
 
+        /// The start fragment flag.
+        ///
+        /// Messages with this flag set tell the parser that the next messages util a message
```
hinto-janai, line 45:
```suggestion
        /// Messages with this flag set tell the parser that the next messages until a message
```

hinto-janai (approved):

PR #85, opened 2024-03-03
database: impl trait function bodies for `heed` & `redb`

hinto-janai:
## Part 10
https://github.com/Cuprate/cuprate/pull/81 <- Previous

- Implements `cuprate_database` trait function bodies for the `heed` and `redb` backends


database/src/backend/heed/database.rs
```diff
@@ -26,72 +26,127 @@ use crate::{
...
+                .next()
+                .map(|result| result.map(|value| value.1).map_err(RuntimeError::from))
+        }
+    }
+
+    // TODO
+    match db.range(tx_ro, &range) {
+        Ok(iter) => Ok(Iter::<'_, T> { iter }),
+        Err(e) => Err(RuntimeError::from(e)),
+    }
```
Boog900, line 97: I don't think you need a custom iterator here:

```suggestion
    Ok(db.range(tx_ro, &range)?.map(|res| Ok(res?.1)))
```

database/src/backend/heed/database.rs
```diff
@@ -26,72 +26,127 @@ use crate::{
...
+    fn get_range<'a, Range>(
         &'a self,
-        key: &'a T::Key,
-        amount: usize,
-    ) -> Result<impl Iterator<Item = &'a T::Value>, RuntimeError>
+        range: Range,
+    ) -> Result<impl Iterator<Item = Result<impl Borrow<T::Value> + 'a, RuntimeError>>, RuntimeError>
     where
-        <T as Table>::Value: 'a,
+        Range: RangeBounds<T::Key> + RangeBounds<&'a T::Key> + 'a,
```
Boog900, line 114: just make this RangeBounds<T::Key> IMO, this will only be used on keys with simple types (u64)

hinto-janai: This is due to https://docs.rs/redb/latest/redb/trait.ReadableTable.html#tymethod.range.

- `heed` wants `Range: RangeBounds<T::Key>`
- `redb` wants `Range: RangeBounds<impl Borrow<T::Key>>`

The current `redb` code is `&T::Key`. If we switch our key to `Cow<'_, T::Key>` too, our range has to `impl Borrow<Cow<'_, T::Key>>`.

I guess the input would be:
```rust
let range = Cow::<u64>::Owned(1)..Cow::<u64>::Owned(10);
```

This does create a incompatibility when switching backends since `heed` wants a `u64` and `redb` wants a `Cow<'_, u64>`. We could switch it so `heed` also uses `Cow` internally I guess...?

Boog900: doesn't T::Key impl Borrow for it self? https://doc.rust-lang.org/std/borrow/trait.Borrow.html#impl-Borrow%3CT%3E-for-T

hinto-janai: To be more clear:
- `heed` wants `Range: RangeBounds<T::Key>` (`heed` sees `&'a T` as the key)
- `redb` wants `Range: RangeBounds<Cow<'a, T::Key>> + 'a` (`redb` sees `Cow<'_, T>` as the key)

We could switch `heed` so it sees `Cow<'_, T>` instead of `&'a T` as the key type, but means we're wrapping all key/values in it during serialization (not bad, still gets serialized as T). It means our ranges will be like the above though...

Or we can make `redb` accept `RangeBounds<T::Key>` then in `redb`'s specific `get_range()` we somehow transform:
```rust
RangeBounds<T::Key>
// ...into...
RangeBounds<Cow<'?, T::Key>>
```
is this possible?

hinto-janai:
@Boog900 ready. `cuprate_database` can finally insert/retrieve values...

Additions:
- `trait ValueGuard` to abstract over `redb::AccessGuard` and `Cow<'_, Key>`
- `trait ToOwnedDebug` to allow `Storable` types to be `ToOwned`, `Debug`, and their `Owned` versions to also be `Debug`
- `trait EnvInner` to abstract over `heed` requiring the caller to hold a guard object while using the environment as it uses locks internally 

Changes:
- `redb`'s key/value type from `Table::Key` -> `Cow<'_, Table::Key>` due to unaligned bytes causing UB and/or panicking when casting. Although, any `T` where alignment is `1` should be casted cheaply

Misc:
- Separated `config.rs` -> `config/`
- Added `cuprate_database::config::Backend` (does nothing for now)
- Main traits are documented with invariants
- Made changes to CI

database/src/backend/heed/transaction.rs
```diff
@@ -9,31 +11,26 @@ use crate::{
...
 }
 
 //---------------------------------------------------------------------------------------------------- TxRw
 impl TxRo<'_> for heed::RwTxn<'_> {
-    /// TODO
-    /// # Errors
-    /// TODO
     fn commit(self) -> Result<(), RuntimeError> {
-        todo!()
+        self.commit().map_err(Into::into)
```
Boog900, line 21: and here:
```suggestion
        Ok(self.commit()?)
```

database/src/backend/heed/transaction.rs
```diff
@@ -9,31 +11,26 @@ use crate::{
...
     }
 }
 
 impl TxRw<'_> for heed::RwTxn<'_> {
-    /// TODO
-    /// # Errors
-    /// TODO
     fn commit(self) -> Result<(), RuntimeError> {
-        todo!()
+        self.commit().map_err(Into::into)
```
Boog900, line 27:
```suggestion
        Ok(self.commit()?)
```

database/src/backend/redb/database.rs
```diff
@@ -1,62 +1,201 @@
...
+#[inline]
+fn get<'a, T: Table + 'static>(
+    db: &'a impl redb::ReadableTable<StorableRedb<T::Key>, StorableRedb<T::Value>>,
+    key: &'a T::Key,
+) -> Result<impl ValueGuard<T::Value> + 'a, RuntimeError> {
+    match db.get(Cow::Borrowed(key)) {
+        Ok(Some(redb_guard)) => Ok(redb_guard),
+        Ok(None) => Err(RuntimeError::KeyNotFound),
+        Err(e) => Err(RuntimeError::from(e)),
+    }
```
Boog900, line 39:
```suggestion
    db.get(Cow::Borrowed(key))?.ok_or(RuntimeError::KeyNotFound)
```

database/src/backend/redb/database.rs
```diff
@@ -1,62 +1,201 @@
...
+        }
+
+        // FIXME: we must match and re-wrap since this function wants `&Cow` not `Cow`.
+        fn end_bound(&self) -> Bound<&Cow<'a, K>> {
+            match &self.end_bound {
+                Bound::Included(t) => Bound::Included(t),
+                Bound::Excluded(t) => Bound::Excluded(t),
+                Bound::Unbounded => Bound::Unbounded,
+            }
+        }
```
Boog900, line 113:
```suggestion
        fn start_bound(&self) -> Bound<&Cow<'a, K>> {
            self.start_bound.as_ref()
        }

        fn end_bound(&self) -> Bound<&Cow<'a, K>> {
            self.end_bound.as_ref()
        }
```

database/src/backend/redb/database.rs
```diff
@@ -1,62 +1,201 @@
...
-        todo!()
+impl<'tx, T: Table + 'static> DatabaseRo<'tx, T> for RedbTableRo<'tx, T::Key, T::Value> {
+    #[inline]
+    fn get<'a>(&'a self, key: &'a T::Key) -> Result<impl ValueGuard<T::Value> + 'a, RuntimeError> {
+        get::<T>(self, key)
     }
 
-    fn get_range<'a>(
+    #[inline]
+    #[allow(clippy::unnecessary_wraps, clippy::trait_duplication_in_bounds)]
```
Boog900, line 145: I don't think this is needed anymore (`allow`)

database/src/backend/redb/database.rs
```diff
@@ -1,62 +1,201 @@
...
-        todo!()
+impl<'tx, T: Table + 'static> DatabaseRo<'tx, T> for RedbTableRw<'_, 'tx, T::Key, T::Value> {
+    #[inline]
+    fn get<'a>(&'a self, key: &'a T::Key) -> Result<impl ValueGuard<T::Value> + 'a, RuntimeError> {
+        get::<T>(self, key)
     }
 
-    fn get_range<'a>(
+    #[inline]
+    #[allow(clippy::unnecessary_wraps, clippy::trait_duplication_in_bounds)]
```
Boog900, line 168: same here

database/src/backend/redb/env.rs
```diff
@@ -61,55 +67,124 @@ impl Env for ConcreteEnv {
...
+        tx_rw.open_table(table)?;
+
+        // TestTable2
+        let table: redb::TableDefinition<
+            'static,
+            StorableRedb<<TestTable2 as Table>::Key>,
+            StorableRedb<<TestTable2 as Table>::Value>,
+        > = redb::TableDefinition::new(TestTable2::NAME);
+        tx_rw.open_table(table)?;
+
```
Boog900, line 110: same here with heed about seperating this into a function at some point so we don't have to repeate this for each table

.github/workflows/ci.yml
```diff
@@ -96,22 +96,16 @@ jobs:
...
         install: mingw-w64-x86_64-toolchain mingw-w64-x86_64-boost msys2-runtime-devel git mingw-w64-x86_64-cmake mingw-w64-x86_64-ninja
 
-    - name: Update Rust (UNIX)
-      if: matrix.os != 'windows-latest'
-      run: rustup update
-
-    - name: Switch target (Windows)
-      if: matrix.os == 'windows-latest'
-      run: rustup toolchain install stable-x86_64-pc-windows-gnu -c clippy && rustup set default-host x86_64-pc-windows-gnu && rustup default stable-x86_64-pc-windows-gnu
-
```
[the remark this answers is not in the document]
Boog900: hmm weird, I don't think these lines caused that error. 

Maybe this got cached? and thats why `windows-gnu` was already installed

database/README.md
```diff
@@ -66,21 +66,23 @@ Note that `lib.rs/mod.rs` files are purely for re-exporting/visibility/lints, an
...
+| `free.rs`           | General free functions (related to the database)
+| `key.rs`            | Abstracted database keys; `trait Key`
+| `resize.rs`         | Database resizing algorithms
+| `storable.rs`       | Data (de)serialization; `trait Storable`
+| `table.rs`          | Database table abstraction; `trait Table`
+| `tables.rs`         | All the table definitions used by `cuprate-database`
+| `to_owned_debug.rs` | Borrowed/owned data abstraction; `trait ToOwnedDebug`
+| `transaction.rs`    | Database transaction abstraction; `trait TxR{o,w}`
+| `types.rs`          | Database table schema types
+| `value_guard.rs`    | Database table schema types
```
Boog900, line 85: was purposed missed for `value_guard`?

database/src/backend/heed/database.rs
```diff
@@ -26,72 +32,113 @@ use crate::{
...
+fn get<'a, T: Table>(
+    db: &'_ HeedDb<T::Key, T::Value>,
+    tx_ro: &'a heed::RoTxn<'_>,
+    key: &T::Key,
+) -> Result<impl ValueGuard<T::Value> + 'a, RuntimeError> {
+    match db.get(tx_ro, key) {
+        Ok(Some(cow)) => Ok(Cow::Borrowed(cow)),
+        Ok(None) => Err(RuntimeError::KeyNotFound),
+        Err(e) => Err(e.into()),
+    }
```
Boog900, line 69:
```suggestion
    db.get(tx_ro, key)?
        .map(Cow::Borrowed)
        .ok_or(RuntimeError::KeyNotFound)
```

Boog900 (changes requested):
Looks good! left some comments

.github/workflows/ci.yml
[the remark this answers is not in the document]
hinto-janai: Should this be added back? CI works without it.

database/src/backend/heed/env.rs
```diff
@@ -77,45 +94,133 @@ impl Drop for ConcreteEnv {
...
+        ) -> Result<(), InitError> {
+            DatabaseOpenOptions::new(env)
+                .name(<T as Table>::NAME)
+                .types::<StorableHeed<<T as Table>::Key>, StorableHeed<<T as Table>::Value>>()
+                .create(tx_rw)?;
+            Ok(())
+        }
+
+        create_table::<TestTable>(&env, &mut tx_rw)?;
+        create_table::<TestTable2>(&env, &mut tx_rw)?;
```
hinto-janai, line 196: `heed`'s `create_table()` (local function in `Env::open`)

database/src/backend/redb/env.rs
```diff
@@ -61,55 +67,123 @@ impl Env for ConcreteEnv {
...
+            tx_rw.open_table(table)?;
+            Ok(())
+        }
 
-    #[cold]
-    #[inline(never)] // called once in [`Env::open`]?`
-    fn create_tables<T: Table>(&self, tx_rw: &mut Self::TxRw<'_>) -> Result<(), RuntimeError> {
-        todo!()
+        create_table::<TestTable>(&tx_rw)?;
+        create_table::<TestTable2>(&tx_rw)?;
```
hinto-janai, line 109: `redb`'s `create_table()` (local function in `Env::open`)

hinto-janai (commented):
Made changes and added the `create_table()` functions.

.github/workflows/ci.yml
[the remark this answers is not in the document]
Boog900: If CI works then this can stay removed

database/src/database.rs
```diff
@@ -1,54 +1,74 @@
...
-    /// TODO
-    //
-    // TODO: (Iterators + ?Sized + lifetimes) == bad time
-    // fix this later.
-    fn get_range<'a>(
+    /// Each key in the `range` has the potential to error, for example,
+    /// if a particular key in the `range` does not exist,
+    /// [`RuntimeError::KeyNotFound`] wrapped in [`Err`] will be returned
+    /// from the iterator.
+    #[allow(clippy::trait_duplication_in_bounds)]
```
Boog900, line 44:
```suggestion
```

Boog900 (changes requested):
just one clippy allow thats isn't needed anymore

hinto-janai:
Fixed, also removed some documentation about our backend's seeing all types as `[u8]` for serde, which isn't true anymore.

They see `Storable{Heed,Redb}` as the implementer, then either `&T` or `Cow<'_, T>` as the type.

Boog900 (approved):

PR #73, opened 2024-02-24
P2P: Peer set

Boog900:

PR #89, opened 2024-03-13
P2P Address book & Handshake changes

Boog900:
This contains changes I made to the address book in #73 

I will add more details here when this is ready


Boog900:
@hinto-janai This is ready

This PR contains a few things:

- Address book bans are now kept in a [DelayQueue](https://docs.rs/tokio-util/latest/tokio_util/time/delay_queue/struct.DelayQueue.html)
- When getting peers from the address book the peer is removed.
   This makes it easier to get multiple random different peers one after the other as the book doesn't have to keep track of peers given out.
- Ping inbound connections if they specify a reachable address so we know if they are reachable or not.
  This caused an issue with out fragmented handshakes tests:
- Fix fragmented buckets when only 1 bucket is needed, the size of the bucket was being set as the whole bucket not just the size of the body
- Fix empty p2p messages, monerod/ levin still adds an epee header to the body so we need to as well.

net/epee-encoding/src/macros.rs
```diff
@@ -35,7 +35,7 @@ macro_rules! try_right_then_left {
 macro_rules! epee_object {
     (
         $obj:ident,
-        $($field: ident $(($alt_name: literal))?: $ty:ty $(as $ty_as:ty )? $(= $default:expr)?  $(=> $read_fn:expr, $write_fn:expr, $should_write_fn:expr)?, )+
+        $($field: ident $(($alt_name: literal))?: $ty:ty $(as $ty_as:ty )? $(= $default:expr)?  $(=> $read_fn:expr, $write_fn:expr, $should_write_fn:expr)?, )*
```
hinto-janai, line 38: Split into multiple lines? Maybe some `// comments` explaining what each input is too.

p2p/address-book/src/peer_list.rs
```diff
@@ -78,37 +79,36 @@ impl<Z: NetworkZone> PeerList<Z> {
...
+            self.remove_peer(&peer)
         } else {
             let len = self.len();
             if len == 0 {
                 None
             } else {
                 let n = r.gen_range(0..len);
 
-                self.peers.values().nth(n)
+                let (&key, _) = self.peers.get_index(n).unwrap();
```
hinto-janai, line 110: Does this work?
```suggestion
                let (&key, _) = self.peers[n];
```

p2p/monero-p2p/src/client/handshaker.rs
```diff
@@ -56,28 +84,42 @@ pub enum HandshakeError {
     IO(#[from] std::io::Error),
 }
 
+/// A request to complete a handshake.
 pub struct DoHandshakeRequest<Z: NetworkZone> {
+    /// The [`InternalPeerID`] of the peer we are handshaking with.
```
hinto-janai, line 89: Hmm I thought clippy would complain on `NonCamelCase` I guess not

p2p/monero-p2p/src/client/handshaker.rs
```diff
@@ -143,6 +185,45 @@ where
...
+            "Peer did not send correct response for ping.",
+        ))?;
+    }
+
+    tracing::debug!("Connection closed before ping response.");
+    Err(BucketError::IO(std::io::Error::new(
+        std::io::ErrorKind::ConnectionAborted,
+        "The peer stream returned None",
+    )))?
+}
```
hinto-janai, line 225: nit: `return` instead of `?`

More clear imo - not a big deal though

p2p/monero-p2p/src/client/handshaker.rs
```diff
@@ -167,11 +248,13 @@ where
         permit,
     } = req;
 
+    // A list of protocol messages the peer has sent during the handshake for us to handle after the handshake.
+    // see: MAX_EAGER_PROTOCOL_MESSAGES
```
hinto-janai, line 252: Allows IDE jump-to-definition
```suggestion
    // see: [`MAX_EAGER_PROTOCOL_MESSAGES`]
```

p2p/monero-p2p/src/network_zones/clear.rs
```diff
@@ -16,6 +16,10 @@ use crate::{NetZoneAddress, NetworkZone};
 impl NetZoneAddress for SocketAddr {
     type BanID = IpAddr;
 
+    fn set_port(&mut self, port: u16) {
+        self.set_port(port)
+    }
```
hinto-janai, line 21: This seems infinitely recursive?

p2p/address-book/src/book.rs
```diff
@@ -220,25 +206,28 @@ impl<Z: NetworkZone> AddressBook<Z> {
...
+        let unban_at = Instant::now() + time;
+
+        self.banned_peers_queue.insert_at(addr.ban_id(), unban_at);
+        self.banned_peers.insert(addr.ban_id(), unban_at);
     }
 
     /// adds a peer to the gray list.
     fn add_peer_to_gray_list(&mut self, mut peer: ZoneSpecificPeerListEntryBase<Z::Addr>) {
         if self.white_list.contains_peer(&peer.adr) {
+            tracing::trace!("Peer {} is already in white list skipping.", peer.adr);
```
hinto-janai, line 218: nit: should be `addr` like everywhere else

p2p/address-book/src/peer_list.rs
Boog900: `[n]` just returns the value, although we could get the key from the value it would require copying the whole value.

p2p/monero-p2p/src/network_zones/clear.rs
Boog900: hmm yeah, this does work though, Rust must resolve impls before trait impls 

I'll change this to `SocketAddr::set_port(self, port)`

p2p/address-book/src/book.rs
Boog900: this is how Monero's p2p network encodes the field name in this message, I could change it but I don't think it's too bad

net/epee-encoding/src/macros.rs
Boog900: I didn't split this into multiple lines but I did add docs to this macro.

hinto-janai (approved):

PR #90, opened 2024-03-14
Keep pruning seeds decompressed

### LINKED ISSUE #37: Pruninng Seeds
When doing pruning calculations Cuprate keeps the seed in a compressed form and works off that. A more idiomatic way IMO is to have a CompressedPruningSeed struct which is just a wrapper around u32 and a PruningSeed which has the pruning seed in decompressed form.

So instead of:
https://github.com/Cuprate/cuprate/blob/343e979e8247b36b648b263af7038f20343a6273/common/src/pruning.rs#L53

We would have:
```rust
pub struct CompressedPruningSeed(u32);

pub struct PruningSeed {
    stripe: u32,
    log_stripes: u32
}
```

This would mean that we could enforce that `PruningSeed` always holds a valid seed and we don't have to account for no pruning in `PruningSeed`.

Boog900:
closes #37


Boog900:
@hinto-janai this is ready (if CI passes)

pruning/src/lib.rs
```diff
@@ -23,9 +23,11 @@ use std::cmp::Ordering;
 use thiserror::Error;
 
 pub const CRYPTONOTE_MAX_BLOCK_NUMBER: u64 = 500000000;
```
hinto-janai, line 25: Is this the max block height?

pruning/src/lib.rs
```diff
@@ -153,43 +325,36 @@ impl PruningSeed {
...
 
             // amt_of_cycles * blocks in a cycle + how many blocks through a cycles until the seed starts storing blocks
             let calculated_height = cycles_start
-                * (CRYPTONOTE_PRUNING_STRIPE_SIZE << seed_log_stripes)
-                + (seed_stripe as u64 - 1) * CRYPTONOTE_PRUNING_STRIPE_SIZE;
+                * (CRYPTONOTE_PRUNING_STRIPE_SIZE << self.log_stripes)
+                + (self.stripe as u64 - 1) * CRYPTONOTE_PRUNING_STRIPE_SIZE;
             if calculated_height + CRYPTONOTE_PRUNING_TIP_BLOCKS > blockchain_height {
                 // if our calculated height is greater than the amount of tip blocks the the start of the tip blocks will be the next un-pruned
                 return Ok(blockchain_height.saturating_sub(CRYPTONOTE_PRUNING_TIP_BLOCKS));
```
hinto-janai, line 360: Could be 1 expression:

```rust
if calculated_height + CRYPTONOTE_PRUNING_TIP_BLOCKS > blockchain_height {
    // if our calculated height is greater than the amount of tip blocks
    // then the start of the tip blocks will be the next un-pruned
    Ok(blockchain_height.saturating_sub(CRYPTONOTE_PRUNING_TIP_BLOCKS))
} else if calculated_height < block_height {
    Err(PruningError::CalculatedHeightSmallerThanEnteredBlock)
} else {
    Ok(calculated_height)
}

File: pruning/src/lib.rs [hunk 3]
```diff
@@ -153,43 +325,36 @@ impl PruningSeed {
         } else if blockchain_height > CRYPTONOTE_MAX_BLOCK_NUMBER {
             Err(PruningError::BlockChainHeightTooLarge)
         } else {
```
hinto-janai, line 327: Could early return here to prevent this `else` block being indented:
```rust
if block_height > CRYPTONOTE_MAX_BLOCK_NUMBER || block_height > blockchain_height {
    return Err(PruningError::BlockHeightTooLarge);
}

if blockchain_height > CRYPTONOTE_MAX_BLOCK_NUMBER {
    return Err(PruningError::BlockChainHeightTooLarge);
}

// rest of function
```

pruning/src/lib.rs
```diff
@@ -219,60 +384,28 @@ impl PruningSeed {
...
             return Ok(block_height);
         }
 
         // We can get the end of our "non-pruning" cycle by getting the next stripe's after us first un-pruned block height
         // so we calculate the next un-pruned block for the next stripe and return it as our next pruned block
-        let next_stripe = (1 + seed_log_stripes) & ((1 << seed_log_stripes) - 1);
-        let seed = PruningSeed::new(next_stripe, seed_log_stripes)?;
+        let next_stripe = (1 + self.log_stripes) & ((1 << self.log_stripes) - 1);
+        let seed = DecompressedPruningSeed::new(next_stripe, self.log_stripes)
+            .expect("We just made sure this stipe is in range for this log_stripe");
```
hinto-janai, line 404:
```suggestion
            .expect("We just made sure this stripe is in range for this log_stripe");
```

pruning/src/lib.rs
Boog900: yes, I'll rename this to `CRYPTONOTE_MAX_BLOCK_HEIGHT` to make it more clear

hinto-janai (approved):

PR #91, opened 2024-03-14
database: implement tables and types

hinto-janai:
## Part 11
https://github.com/Cuprate/cuprate/pull/85 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/92

- Implements the tables and types from https://github.com/Cuprate/cuprate/pull/59's in `cuprate_database::{tables, types}`
- Implements `cuprate_database::{Table, Storable}` on them
- Adds a test for each table in `src/backend/tests.rs`
- Makes `Env::open` create all tables

`/// docs` will be done in a later PR.


database/src/backend/heed/env.rs
```diff
@@ -171,29 +171,43 @@ impl Env for ConcreteEnv {
...
+        create_table::<KeyImages>(&env, &mut tx_rw)?;
+        create_table::<NumOutputs>(&env, &mut tx_rw)?;
+        create_table::<Outputs>(&env, &mut tx_rw)?;
+        create_table::<PrunableHashes>(&env, &mut tx_rw)?;
+        create_table::<PrunableTxBlobs>(&env, &mut tx_rw)?;
+        create_table::<PrunedTxBlobs>(&env, &mut tx_rw)?;
+        create_table::<RctOutputs>(&env, &mut tx_rw)?;
+        create_table::<TxHeights>(&env, &mut tx_rw)?;
+        create_table::<TxIds>(&env, &mut tx_rw)?;
+        create_table::<TxUnlockTime>(&env, &mut tx_rw)?;
```
hinto-janai, line 210: Is there any type reflection magic we can do here? Something like:

```rust
for table in crate::tables::* {
    create_table::<table>();
}
```

database/src/tables.rs
```diff
@@ -74,16 +77,71 @@ macro_rules! tables {
...
+    NumOutputs,
+    Amount => AmountIndex,
+
+    /// TODO
+    PrunedTxBlobs,
+    TxId => PrunedBlob,
+
+    /// TODO
+    Outputs,
+    Amount => Output, // FIXME: `Amount | AmountIndex` key
```
hinto-janai, line 120: This is `u64 | u64` so the type key is `u64` but what should the alias name be?

database/src/types.rs
```diff
@@ -49,96 +45,290 @@ use bytemuck::{AnyBitPattern, NoUninit, Pod, Zeroable};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-//---------------------------------------------------------------------------------------------------- TestType
-/// TEST
+//---------------------------------------------------------------------------------------------------- Aliases
+// TODO: document these, why they exist, and their purpose.
+//
+// TODO: should we go as far as making transparent
+// wrappers for all of these to be more typesafe?
```
hinto-janai, line 52: Stuff like `Amount`, `KeyImage` are just type aliases for now.

I think we should do something like `KeyImage(pub [u8; 32])` instead? Returning newtypes would be a lot more type safe.

database/src/types.rs
```diff
@@ -49,96 +45,290 @@ use bytemuck::{AnyBitPattern, NoUninit, Pod, Zeroable};
...
+    pub cumulative_difficulty: u64,
+    /// TODO
+    pub block_hash: [u8; 32],
+    /// TODO
+    pub cumulative_rct_outs: u32,
+    /// TODO
+    // TODO: get rid of or use this padding.
+    // Without this, `BlockInfoV2` is 68 bytes,
+    // which isn't aligned to 8.
+    pub _pad: [u8; 4],
```
hinto-janai, line 190: We have 4 bytes to play with here.

Could do `cumulative_rct_outs: u64` if we don't want explicit padding.

database/src/backend/heed/env.rs
Boog900: I don't think so but you might be able to make `create_table` a closure so you don't need to take in `env` and `tx_rw` on each call

database/src/tables.rs
Boog900: `PreRCTOutputID`?

 By `u64 | u64` I meant a byte concatenation so a [u8; 16] but the key should be `(u64, u64)` to Rust

database/src/types.rs
Boog900: I don't think this is something to worry about too much, if we are adding data to the wrong tables I think we would find out pretty quick right?

database/src/types.rs
Boog900: yeah I think just do `cumulative_rct_outs: u64`

database/src/backend/heed/env.rs
hinto-janai: Can you do generic closures? (`T: Table`)

database/src/tables.rs
hinto-janai: Ah, I see. It will have to be a named struct since `Pod` isn't implemented on tuples.

hinto-janai:
@Boog900 ready.

database/src/backend/heed/env.rs
Boog900: doesn't look like it :(

database/src/types.rs
```diff
@@ -49,96 +45,319 @@ use bytemuck::{AnyBitPattern, NoUninit, Pod, Zeroable};
...
+    /// TODO
+    pub cumulative_difficulty: u64,
+    /// TODO
     ///
-    /// TODO: this is basically leeway to
-    /// add more things to our structs too,
-    /// because otherwise this space is wasted.
-    pub _pad: [u8; 7],
+    /// TODO: note that this is originally u32,
+    /// but is u64 here for padding reasons.
```
Boog900, line 218: an explanation here isn't really needed IMO so this TODO could be removed

database/src/types.rs
```diff
@@ -49,96 +45,319 @@ use bytemuck::{AnyBitPattern, NoUninit, Pod, Zeroable};
...
+    pub timestamp: u64,
+    /// TODO
+    pub total_generated_coins: u64,
+    /// TODO
+    pub weight: u64,
+    // Maintain 8 byte alignment.
+    /// TODO
+    pub cumulative_difficulty_low: u64,
+    /// TODO
+    pub cumulative_difficulty_high: u64,
```
Boog900, line 267: For the next PR where a maximum alignment isn't needed this can be changed back into a u128

Boog900 (approved):

database/src/backend/heed/env.rs
SyntheticBird45: Only way would be Iterator over `dyn Table` but it needs `[Box<T: Table>]` so really not worth it

PR #92, opened 2024-03-18
database: return owned `T` in `Storable`

hinto-janai:
## Part 12

https://github.com/Cuprate/cuprate/pull/91 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/93

- `Storable::from_bytes(&[u8]) -> &T` -> `Storable::from_bytes(&[u8]) -> T`
- Adds `struct StorableVec<T>(pub Vec<T>)` for slice types
- Fixes `DatabaseRo::{get, get_range}()` to return `T` directly
- Removes `trait ValueGuard` as we return an owned `T` directly now
- Removes `Storable::ALIGN` as we don't care about alignment anymore


hinto-janai:
@Boog900 there's communication issues with Matrix.

Your reply after my:
```
both heed and redb will -> Cow<'_, T::Value> [...]
```
is encrypted, don't think you received some of my messages either. Email or post here maybe?

Boog900:
Weird, I did use my laptop to message you when the first issue happend but since then I am back on my computer so it should be fine again.

Here was my message:



if we were to change the table definition in the table macro to `pub struct [<$table:camel>]<'a>(PhantomData<&'a ()>);`

Then on impl Table we can use this lifetime for the Value if needed.

I have made some VERY rough code just to test if this would work and it _seems_ to:

```rust
use core::marker::PhantomData;
use core::fmt::Debug;
use std::borrow::Cow;


trait Table {
    type T: Debug;
    
    fn new_t() -> Self::T;
}

struct TestTable<'a>(PhantomData<&'a ()>);

impl<'a> Table for TestTable<'a> {
    // Cow used in place of StorableSlice
    type T = Cow<'a, [u8]>;
    
    fn new_t() -> Cow<'static, [u8]> {
        Cow::Owned(vec![3, 2, 1])
    }
}

fn test_put<T: Table>(t: &T::T) {
    println!("{:?}", t)
}

fn test_get<T: Table>() -> T::T {
    T::new_t()
}

fn main() {
    let t = [1_u8,2,3];
    test_put::<TestTable>(&Cow::Borrowed(t.as_slice()));
    let tt = test_get::<TestTable>();
    println!("{:?}", tt)
}
```

hinto-janai:
It does seem to work but what about the special impl for `[u8]` and `Cow` as the returned type? It would allow us to just pass `[u8]` along, no alloc/cast.

Most types wouldn't have to copy inside `heed` either since we can return `Cow::Borrowed(casted_ref)`. The only type that would alloc is non `u8` slices. Are there more slice types than `[u8]` and `[u64]` we'll be using?

Boog900:
\[u8\] has no alignment requirement https://doc.rust-lang.org/reference/type-layout.html#array-layout so I think whats happening is the [u8] is throwing off the key for that table which means the key will need to be special cased.

Plus this PR also allows us to remove the ValueGuard and Cow on return.

Even if we don't copy in `heed` we will end up copying data anyway as it gets returned from the service. 

> Are there more slice types than [u8] and [u64] we'll be using

No

Boog900:
I tried to find a way to do this nicely with the current traits and I couldn't, I think what we are going to need to do is add a type like redb's [SelfType](https://docs.rs/redb/latest/redb/trait.RedbValue.html#associatedtype.SelfType) on our Storable trait and use this instead of self in the `as_bytes`/`from_bytes` method.

We would also need to change a lot of the signatures to now use <T::Value as Storable>::SelfType and <T::Key as Storable>::SelfType

hinto-janai:
Surely impl for `[u8]` and returning `Cow` is simpler at that point...?

Boog900:
it's the u64, key, that needs the special impl 

I still think this PR is a good idea just to get rid of the 2 levels of indirection we have we returning values.

If we were to remove the StorableSlice completely and go with the plan of having a  `StorableVec` like:
```rust
struct StorableVec(Vec<T>)
```

and impl `RefCast` I think this would work .... in our messages you said that if we had a &[u8] we would have to collect into a Vec first but I can't think of a situation when adding data to the database through the service where you would only get a slice.

hinto-janai:
`StorableVec` would be a solution but any use-case that doesn't use `service` will suffer. I'm okay with doing this but just saying.

> so I think whats happening is the [u8] is throwing off the key for that table which means the key will need to be special cased

Btw I believe you are correct here - after testing no `bytemuck` on `[u8]` it still panics on the key.

Boog900:
> any use-case that doesn't use service will suffer

I don't think these use cases will need to maximizes performance, so this should be ok

database/src/database.rs
```diff
@@ -22,19 +21,18 @@ use crate::{
...
-    /// This returns a guard to the value, not the value itself.
-    /// See [`ValueGuard`] for more info.
+    /// The returned value is _owned_.
     ///
     /// # Errors
     /// This will return [`RuntimeError::KeyNotFound`] wrapped in [`Err`] if `key` does not exist.
     ///
     /// It will return other [`RuntimeError`]'s on things like IO errors as well.
-    fn get<'a>(&'a self, key: &'a T::Key) -> Result<impl ValueGuard<T::Value> + 'a, RuntimeError>;
+    fn get(&self, key: &T::Key) -> Result<T::Value, RuntimeError>;
```
hinto-janai, line 30: No more lifetimes needed, returned value is owned.

database/src/database.rs
```diff
@@ -43,11 +41,8 @@ pub trait DatabaseRo<'tx, T: Table> {
     /// from the iterator.
     fn get_range<'a, Range>(
         &'a self,
-        range: &'a Range,
-    ) -> Result<
-        impl Iterator<Item = Result<impl ValueGuard<T::Value>, RuntimeError>> + 'a,
-        RuntimeError,
-    >
+        range: Range,
+    ) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
```
hinto-janai, line 45: For `get_range()`, the iterator itself is tied to `self` and `range` but the values actually yielded from the iterator are fully owned, which is tested...

database/src/backend/tests.rs
```diff
@@ -200,31 +200,47 @@ fn db_read_write() {
...
+    // Assert count is correct.
+    assert_eq!(100, table.get_range(range.clone()).unwrap().count());
+
+    // Assert each returned value from the iterator is owned.
+    {
+        let mut iter = table.get_range(range.clone()).unwrap();
+        let value: Output = iter.next().unwrap().unwrap(); // 1. take value out
+        drop(iter); // 2. drop the `impl Iterator + 'a`
+        assert_same(value); // 3. assert even without the iterator, the value is alive
+    }
```
hinto-janai, line 234: ...here.

hinto-janai (commented):
@Boog900 ready, added changes to the main post.

database/src/storable.rs
```diff
@@ -123,46 +96,27 @@ pub trait Storable: ToOwnedDebug {
...
-    /// `bytes` must be perfectly aligned for `Self`
-    /// or else this function may cause UB.
-    ///
-    /// This function _may_ panic if `bytes` isn't aligned.
+    /// Create an owned [`Self`] from bytes.
     ///
     /// # Blanket implementation
     /// The blanket implementation that covers all types used
     /// by `cuprate_database` will simply cast `bytes` into `Self`,
     /// with no copying.
```
Boog900, line 112: The docs here need to be updated as we are now copying

database/src/tables.rs
```diff
@@ -67,6 +67,7 @@ macro_rules! tables {
             impl private::Sealed for [<$table:camel>] {}
 
             // Table trait impl.
+            #[allow(single_use_lifetimes)]
             impl Table for [<$table:camel>] {
```
Boog900, line 71: is this allow needed?

Boog900 (changes requested):
I think the readme needs to be updated to remove references to value guard as well

hinto-janai:
Will we also need to add something like `Storable{Arc,Box,Bytes}<T>(pub {Arc,Box,Bytes}<T>)`?

Boog900:
Yeah we will need a StorableBytes but I don't think we need any of the others

database/src/storable.rs
```diff
@@ -63,35 +65,6 @@ use crate::ToOwnedDebug;
 /// Most likely, the bytes are little-endian, however
 /// that cannot be relied upon when using this trait.
 pub trait Storable: ToOwnedDebug {
```
Boog900, line 67: Do we still need `ToOwnedDebug`?

hinto-janai:
- Replaced `ToOwnedDebug` with regular `Debug`
- Added `StorableBytes(pub bytes::Bytes)`

Boog900 (approved):

PR #93, opened 2024-03-19
database: use `rayon` for `service`'s reader thread-pool

hinto-janai:
## Part 13
https://github.com/Cuprate/cuprate/pull/92 <- Previous

- Replaces `cuprate_database::service`'s custom `DatabaseReader` thread-pool with a rayon backed one
- Implements `tower::Service` using `rayon::spawn` instead of a channel message


database/src/service/read.rs
```diff
@@ -8,6 +8,8 @@ use std::{
 
 use crossbeam::channel::Receiver;
 
+use futures::channel::oneshot;
```
hinto-janai, line 11: `InfallibleOneshotReceiver` uses `futures`'s oneshot, not `tokio` - on purpose?

database/src/service/read.rs
```diff
@@ -47,145 +49,163 @@ type ResponseSend = tokio::sync::oneshot::Sender<ResponseResult>;
...
-    /// Sender channel to the database read thread-pool.
+    /// Handle to the custom `rayon` DB reader thread-pool.
+    ///
+    /// Requests are "spawn"ed in this thread-pool, and responses
+    /// are returned via a channel we (the caller) provide.
+    pub(super) pool: Arc<rayon::ThreadPool>,
+
+    /// Access to the database.
+    pub(super) db: Arc<ConcreteEnv>,
+}
```
hinto-janai, line 62: Each reader handle will now have to `Arc::clone(&self.db)` and pass it to the actual `rayon` thread as they don't hold any state themselves.

This means we're adding an `Arc::clone` each request.

database/src/service/read.rs
```diff
@@ -47,145 +49,163 @@ type ResponseSend = tokio::sync::oneshot::Sender<ResponseResult>;
...
     type Error = RuntimeError;
-    type Future = ResponseRecv;
+    type Future = ResponseReceiver;
 
     #[inline]
     fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        todo!()
+        // Threadpool is always ready as long as this handle is alive.
+        Poll::Ready(Ok(()))
     }
```
hinto-janai, line 107: Should we do something here when the threadpool is at capacity?

database/src/service/write.rs
```diff
@@ -48,22 +50,54 @@ pub struct DatabaseWriteHandle {
...
     #[inline]
-    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        todo!()
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // If there are previous write requests not fulfilled, wait.
+        if self.sender.is_empty() {
+            Poll::Ready(Ok(()))
+        } else {
+            Poll::Pending
+        }
```
hinto-janai, line 89: Not sure if this is needed.

database/src/service/read.rs
```diff
@@ -47,145 +49,163 @@ type ResponseSend = tokio::sync::oneshot::Sender<ResponseResult>;
...
+        // Spawn the request in the rayon DB thread-pool.
+        //
+        // Note that this uses `install()` instead of `spawn()`
+        // such that any `rayon` parallel code that runs within
+        // the passed closure uses the same `rayon` threadpool.
+        //
+        // INVARIANT:
+        // The below `DatabaseReader` function impl block relies on this behavior.
+        self.pool
+            .install(move || db_reader.map_request(request, response_sender));
```
hinto-janai, line 127: Is this correct? `par_*()` code in this closure will use the same `self.pool`?

hinto-janai (commented):
@Boog900 ready.

database/src/service/write.rs
hinto-janai: The channel itself acts as a backlog since it's unbounded, so this could be removed?

database/src/service/read.rs
Boog900: This should be ok

database/src/service/write.rs
Boog900: Yeah I don't think this is needed for the write service.

Plus when you return `Poll::Pending` from a future/ function returning `Poll` you MUST ensure the `Context` is woken up when the future becomes ready, here you drop the context so when the channle is full any write requests would just wait forever even if the channel is no longer empty

database/src/service/read.rs
Boog900: Yeah we do need some sort of backpressure here to make sure we don't get overloaded with requests.

I think the best way to do this is to have a tokio semaphore and when `poll_ready` is called we acquire a permit, when the permit is acquired we return Poll::Ready.

We then pass this permit into the rayon pool so when the task is completed the permit is dropped.

The amount of permits should probably be the amount of threads we have in the pool

database/src/service/read.rs
```diff
@@ -47,145 +49,163 @@ type ResponseSend = tokio::sync::oneshot::Sender<ResponseResult>;
...
-/// Database reader thread.
+/// Database reader thread state.
 ///
-/// This struct essentially represents a thread.
+/// This struct represents (some of the) state every
+/// `DatabaseReadHandle` rayon thread will have access to.
 ///
-/// Each reader thread is spawned with access to this struct (self).
+/// Could just be function inputs, although this allows `self`
+/// and a `Drop` impl such that each request has the same cleanup routine.
```
Boog900, line 140: IMO this would probably be clearer as just function inputs, instead of the extra struct

database/src/service/read.rs
Boog900: no, is tokio's better?

hinto-janai: Not sure, `futures` is fine though

database/src/service/read.rs
hinto-janai: Where would we store the `SemaphorePermit<'_>` in-between `poll_ready()` and `call()`?

hinto-janai: Ah, like this?
```rust
pub struct DatabaseReadHandle {
    pool: Arc<rayon::ThreadPool>,
    semaphore: Arc<Semaphore>,
    permit: Option<OwnedSemaphorePermit>,
    env: Arc<ConcreteEnv>,
}
```
Would it be something like this in `call()`?
```rust
let permit = selt.permit
    .take()
    .expect("we should have acquired a permit in poll_ready()");

/* pass to rayon pool */
```

Boog900: Yep

> Although OwnedSemaphorePermit is !Clone so we'd have to manually impl and set it as None.

That's ok

Boog900: you will need this though: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.PollSemaphore.html#

database/src/service/read.rs
```diff
@@ -47,145 +52,211 @@ type ResponseSend = tokio::sync::oneshot::Sender<ResponseResult>;
...
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Some(permit)) => {
+                self.permit = Some(permit);
+                Poll::Ready(Ok(()))
+            }
+
+            // `self` itself owns the backing semaphore,
+            // so it can't be closed.
+            Poll::Ready(None) => unreachable!(),
+        }
```
Boog900, line 170: the `ready!` macro would probably be good here (need to import it though):
```suggestion
        let Some(permit) = ready!(self.semaphore.poll_acquire(cx)) else {
            // `self` itself owns the backing semaphore,
            // so it can't be closed.
            unreachable!()
        };
        
        self.permit = Some(permit);
        Poll::Ready(Ok(()))
        
```

Boog900 (approved):

PR #94, opened 2024-03-22
Implement `types/` 

hinto-janai:
## Part 14
https://github.com/Cuprate/cuprate/pull/93 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/95

- Implements a top level `cuprate-types` crate for various shared data types


hinto-janai:
Right, I was wondering how you wanted to do it for these types - I do think moving stuff to a separate crate makes the most sense, how does `cuprate-types` sound? (since it'll have more than just traits)

We could actually move all of `cuprate_database::types` inside there too, no?

Also, what should the `Request/Response` variants be? Are the current ones appropriate?

Boog900:
> how does cuprate-types sound? (since it'll have more than just traits)

Yeah that sounds good.

> We could actually move all of cuprate_database::types inside there too, no?

I'm not sure about this, these types will only be used in the DB and not in the service interface, so they can probably stay there.

> Also, what should the Request/Response variants be? Are the current ones appropriate?

Yeah, we will need to add more but the current ones are all the consensus code needs.

For the write request we will just need a write block request which takes in [`VerifiedBlockInformation`]

SyntheticBird45:
> I'm not sure about this, these types will only be used in the DB and not in the service interface, so they can probably stay there.

Agree on this. No need to declare on higher crates what stay in specific impl

Boog900:
I think it would be best for `types/` to keep hfs as `u8`s so we don't need to pull all the hf stuff out of consensus.

We will NEED to do this for votes anyway as going from u8 -> `HardForks` for votes means you may lose the actual vote as it will just go to the highest known hf. 

Sorry if you spent a while copying all that over

hinto-janai:
Should there be 2 `ExtendedBlockHeader` types? One with `u8` and one with `HardFork`?

`enum`s are `!Pod` so we need to represent it as `u8` but it seems bad to lose all that type information. Maybe the type stored isn't `ExtendedBlockHeader` but some builder/middle type that can `Into<ExtendedBlockHeader>`?

hinto-janai:
> For the write request we will just need a write block request which takes in VerifiedBlockInformation

Should `VerifiedBlockInformation` be moved into `types/`? It pulls in a lot of other types.

What specifically should be in `types/`? Should it even exist? 

It would be nice to have all types in one place so there is no unnecessary mapping between `consensus/` and `database/` types. This PR originally was going to:
 - impl `Pod` and fix layout where needed on `consensus/` types 
 - keep `Request/Response` in `database/`
 - use and re-export `consensus/` types in `database/`

Although if a type is used in at least 2 places I think that's enough reason to have `types/` and move it there.

Boog900:
> Should there be 2 ExtendedBlockHeader types? One with u8 and one with HardFork?

No I'll change the one in consensus to use the one in `types` in another PR.

> enums are !Pod so we need to represent it as u8 but it seems bad to lose all that type information. Maybe the type stored isn't ExtendedBlockHeader but some builder/middle type that can Into<ExtendedBlockHeader>?

We wont be storing `ExtendedBlockHeader` so it shouldn't need to impl Pod. We will construct it from multiple tables data.

> Should VerifiedBlockInformation be moved into types/? It pulls in a lot of other types.

here is a minimal version to use in `types`:

```rust
pub struct VerifiedBlockInformation {
    pub block: Block,
    pub hf_vote: u8,
    pub txs: Vec<Arc<TransactionVerificationData>>,
    pub block_hash: [u8; 32],
    pub pow_hash: [u8; 32],
    pub height: u64,
    pub generated_coins: u64,
    pub weight: usize,
    pub long_term_weight: usize,
}

pub struct TransactionVerificationData {
    pub tx: Transaction,
    pub tx_blob: Vec<u8>,
    pub tx_weight: usize,
    pub fee: u64,
    pub tx_hash: [u8; 32],
}

```
> What specifically should be in types/? Should it even exist?

Right now just stuff in the Read/Write request enums for the DB (including the enums)

types/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dependencies]
+borsh            = { workspace = true, optional = true }
+cfg-if           = { workspace = true }
+curve25519-dalek = { workspace = true }
+monero-serai     = { workspace = true }
+paste            = { workspace = true }
+thiserror        = { workspace = true }
+serde            = { workspace = true, optional = true }
+strum            = { workspace = true, features = ["derive"], optional = true }
```
Boog900, line 23: what are your plans for this? 

Although I don't want to worry about it too much now, I think a goal should be to have as little dependencies as possible (within reason)

Boog900: Also I don't think all the dependencies defined here are actually used

hinto-janai: Nothing in particular, mostly just implementing now for future proofing if it's ever needed, should it be removed?

hinto-janai: Or feature gated?

types/src/types.rs
```diff
@@ -0,0 +1,105 @@
...
+//---------------------------------------------------------------------------------------------------- VerifiedBlockInformation
+/// TODO
+#[derive(Clone, Debug, PartialEq, Eq)]
+// #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))] // FIXME: monero_serai
+// #[cfg_attr(feature = "borsh", derive(BorshSerialize, BorshDeserialize))]
+pub struct VerifiedBlockInformation {
+    /// TODO
+    pub block: Block,
+    /// TODO
+    pub hf_vote: u8,
```
Boog900, line 66: hf_vote can be removed actually it is stored in `Block`

Cargo.toml
```diff
@@ -62,6 +64,7 @@ rayon                 = { version = "1.9.0", default-features = false }
 serde_bytes           = { version = "0.11.12", default-features = false }
 serde_json            = { version = "1.0.108", default-features = false }
 serde                 = { version = "1.0.190", default-features = false }
+strum                 = { version = "0.26.2", default-features = false }
```
Boog900, line 67:
```suggestion
```

hinto-janai: oops...

Boog900 (approved):

PR #96, opened 2024-03-25
database: add more `fn` to `trait DatabaseR{o,w}`

hinto-janai:
## Part 16
https://github.com/Cuprate/cuprate/pull/95 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/97

This add various functions to `trait DatabaseR{o,w}` that are supported by both backends `heed` & `redb` - more specifically, APIs that are `BTreeMap`-like are added since both are based on B-trees.

All methods were added to the current tests in `src/backend/tests.rs`.

```rust
// Read methods.

fn iter(
    &self,
) -> Result<impl Iterator<Item = Result<(T::Key, T::Value), RuntimeError>> + '_, RuntimeError>;

fn keys(
    &self,
) -> Result<impl Iterator<Item = Result<T::Key, RuntimeError>> + '_, RuntimeError>;  

fn values(
    &self,
) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + '_, RuntimeError>; 

fn contains(&self, key: &T::Key) -> Result<bool, RuntimeError>;

fn len(&self) -> Result<u64, RuntimeError>; 

fn first(&self) -> Result<(T::Key, T::Value), RuntimeError>; 

fn last(&self) -> Result<(T::Key, T::Value), RuntimeError>; 

fn is_empty(&self) -> Result<bool, RuntimeError>; 
```
```rust
// Write methods.

fn pop_first(&mut self) -> Result<(T::Key, T::Value), RuntimeError>;

fn pop_last(&mut self) -> Result<(T::Key, T::Value), RuntimeError>;
```
```rust
// trait EnvInner

fn clear_db<T: Table>(&self, tx_rw: &mut impl TxRw<'_>) -> Result<(), RuntimeError>;
```

### DIFF
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -69,11 +69,11 @@ fn get_range<'a, T: Table, Range>(
     db: &'a HeedDb<T::Key, T::Value>,
     tx_ro: &'a heed::RoTxn<'_>,
     range: Range,
-) -> Result<impl Iterator<Item = Result<(T::Key, T::Value), RuntimeError>> + 'a, RuntimeError>
+) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
 where
     Range: RangeBounds<T::Key> + 'a,
 {
-    Ok(db.range(tx_ro, &range)?.map(|res| Ok(res?)))
+    Ok(db.range(tx_ro, &range)?.map(|res| Ok(res?.1)))
 }
 
 /// Shared [`DatabaseRo::iter()`].
@@ -150,7 +150,7 @@ impl<T: Table> DatabaseRo<T> for HeedTableRo<'_, T> {
     fn get_range<'a, Range>(
         &'a self,
         range: Range,
-    ) -> Result<impl Iterator<Item = Result<(T::Key, T::Value), RuntimeError>> + 'a, RuntimeError>
+    ) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
     where
         Range: RangeBounds<T::Key> + 'a,
     {
@@ -211,7 +211,7 @@ impl<T: Table> DatabaseRo<T> for HeedTableRw<'_, '_, T> {
     fn get_range<'a, Range>(
         &'a self,
         range: Range,
-    ) -> Result<impl Iterator<Item = Result<(T::Key, T::Value), RuntimeError>> + 'a, RuntimeError>
+    ) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
     where
         Range: RangeBounds<T::Key> + 'a,
     {
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -240,6 +240,7 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
     {
         let mut iter = self.db.iter_mut(self.tx_rw)?;
 
+        // FIXME: is this optimal?
         loop {
             let Some(result) = iter.next() else {
                 return Ok(());
@@ -254,6 +255,7 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
                 // We are deleting the value and never accessing
                 // it again so this should be safe.
                 unsafe {
+                    // TODO: is this actually safe?
                     iter.del_current()?;
                 }
             }
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -279,36 +279,6 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
     }
 
     #[inline]
-    fn retain<P>(&mut self, mut predicate: P) -> Result<(), RuntimeError>
-    where
-        P: FnMut(T::Key, T::Value) -> bool,
-    {
-        let mut iter = self.db.iter_mut(self.tx_rw)?;
-
-        // FIXME: is this optimal?
-        loop {
-            let Some(result) = iter.next() else {
-                return Ok(());
-            };
-
-            let (key, value) = result?;
-
-            if !predicate(key, value) {
-                // SAFETY:
-                // It is undefined behavior to keep a reference of
-                // a value from this database while modifying it.
-                // We are deleting the value and never accessing
-                // it again so this should be safe.
-                unsafe {
-                    iter.del_current()?;
-                }
-            }
-        }
-
-        Ok(())
-    }
-
-    #[inline]
     fn pop_first(&mut self) -> Result<(T::Key, T::Value), RuntimeError> {
         // Get the first value first...
         let Some(first) = self.db.first(self.tx_rw)? else {
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -274,11 +274,6 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
     }
 
     #[inline]
-    fn clear(&mut self) -> Result<(), RuntimeError> {
-        Ok(self.db.clear(self.tx_rw)?)
-    }
-
-    #[inline]
     fn pop_first(&mut self) -> Result<(T::Key, T::Value), RuntimeError> {
         // Get the first value first...
         let Some(first) = self.db.first(self.tx_rw)? else {
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -53,7 +53,7 @@ pub(super) struct HeedTableRw<'env, 'tx, T: Table> {
 // call the functions since the database is held by value, so
 // just use these generic functions that both can call instead.
 
-/// Shared generic `get()` between `HeedTableR{o,w}`.
+/// Shared [`DatabaseRo::get()`].
 #[inline]
 fn get<T: Table>(
     db: &HeedDb<T::Key, T::Value>,
@@ -63,7 +63,7 @@ fn get<T: Table>(
     db.get(tx_ro, key)?.ok_or(RuntimeError::KeyNotFound)
 }
 
-/// Shared generic `get_range()` between `HeedTableR{o,w}`.
+/// Shared [`DatabaseRo::get_range()`].
 #[inline]
 fn get_range<'a, T: Table, Range>(
     db: &'a HeedDb<T::Key, T::Value>,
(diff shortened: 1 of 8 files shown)


hinto-janai:
@Boog900 I think you said `clear()` wasn't necessarily needed but I thought these extra functions would be rather useful for other users/applications, even if `cuprate_database` itself doesn't use them.

database/src/backend/heed/database.rs
```diff
@@ -87,12 +150,54 @@ impl<T: Table> DatabaseRo<T> for HeedTableRo<'_, T> {
     fn get_range<'a, Range>(
         &'a self,
         range: Range,
-    ) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
+    ) -> Result<impl Iterator<Item = Result<(T::Key, T::Value), RuntimeError>> + 'a, RuntimeError>
```
hinto-janai, line 153: Something to note about including the `T::Key` by default is that `redb::AccessValue::access()` on the key will add just the slight bit more overhead.

Could have a `get_range_keys()` and `get_range_values()` I guess.

database/src/backend/heed/database.rs
```diff
@@ -125,6 +272,88 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
...
+            if !predicate(key, value) {
+                // SAFETY:
+                // It is undefined behavior to keep a reference of
+                // a value from this database while modifying it.
+                // We are deleting the value and never accessing
+                // it again so this should be safe.
+                unsafe {
+                    iter.del_current()?;
+                }
+            }
```
hinto-janai, line 305: Pretty sure this is always safe since we copy `(key, value)` anyway.

database/src/backend/heed/database.rs
```diff
@@ -125,6 +272,88 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
...
+        let mut iter = self.db.iter_mut(self.tx_rw)?;
+
+        // SAFETY:
+        // It is undefined behavior to keep a reference of
+        // a value from this database while modifying it.
+        // We are deleting the value and never accessing
+        // the iterator again so this should be safe.
+        unsafe {
+            iter.del_current()?;
+        }
```
hinto-janai, line 333: ...same here...

database/src/backend/heed/database.rs
```diff
@@ -125,6 +272,88 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
...
+        let mut iter = self.db.rev_iter_mut(self.tx_rw)?;
+
+        // SAFETY:
+        // It is undefined behavior to keep a reference of
+        // a value from this database while modifying it.
+        // We are deleting the value and never accessing
+        // the iterator again so this should be safe.
+        unsafe {
+            iter.del_current()?;
+        }
```
hinto-janai, line 353: ...and here.

database/src/backend/redb/database.rs
```diff
@@ -78,29 +187,105 @@ impl<T: Table + 'static> DatabaseRo<T> for RedbTableRw<'_, T::Key, T::Value> {
...
+        // Deleting and re-creating the table outright
+        // would be faster, although we don't have direct
+        // access to the write transaction, which implements
+        // this functionality in `redb::WriteTransaction::delete_table`.
+
+        // Pop the last `(key, value)` until there are none.
+        while redb::Table::pop_last(self)?.is_some() {}
+
         Ok(())
     }
```
hinto-janai, line 267: I think we could create an intermediate `RedbTable` type like our `HeedTable` type such that it holds onto a reference of the transaction such that this would be possible.

database/src/database.rs
```diff
@@ -39,12 +46,73 @@ pub trait DatabaseRo<T: Table> {
...
+    ///
+    /// # Errors
+    /// TODO
+    fn contains(&self, key: &T::Key) -> Result<bool, RuntimeError> {
+        match self.get(key) {
+            Ok(_) => Ok(true),
+            Err(RuntimeError::KeyNotFound) => Ok(false),
+            Err(e) => Err(e),
+        }
+    }
```
hinto-janai, line 91: I think `RuntimeError::KeyNotFound` shouldn't exist and we should return `Option<(T::Key, T::Value)>` like all other APIs do. On stuff like `get()` too.

database/src/database.rs
```diff
@@ -65,4 +133,35 @@ pub trait DatabaseRw<T: Table>: DatabaseRo<T> {
...
+    /// - `true == keep`
+    /// - `false == remove`
+    ///
+    /// # Errors
+    /// TODO
+    ///
+    /// TODO: Document early return on error.
+    fn retain<P>(&mut self, predicate: P) -> Result<(), RuntimeError>
+    where
+        P: FnMut(T::Key, T::Value) -> bool;
```
hinto-janai, line 154: What should be the behavior on database errors mid-`retain()`?

Currently it just early returns, so all removes up until that point are still `commit()`-able. I think this makes the most sense since you can `abort()` if there's an error?

database/src/backend/heed/database.rs
Boog900: Do we need to include the key?

database/src/database.rs
Boog900: With the service interface this be an error any way, so we would have to map the option to an error. IMO I don't think this is worth it

database/src/backend/heed/database.rs
```diff
@@ -125,6 +272,88 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
...
     }
+
+    #[inline]
+    fn clear(&mut self) -> Result<(), RuntimeError> {
+        Ok(self.db.clear(self.tx_rw)?)
+    }
+
+    #[inline]
+    fn retain<P>(&mut self, mut predicate: P) -> Result<(), RuntimeError>
+    where
```
Boog900, line 283: I don't really think this method will have any use, IMO this one should be removed.

database/src/backend/heed/database.rs
Boog900: Yeah as long as the txs are not sync (they aren't)

database/src/backend/redb/database.rs
Boog900: Yeah I think that would be better

Boog900 (changes requested):

database/src/backend/heed/database.rs
hinto-janai: Not necessarily but most `get_range()` APIs return `(key, value)` (`heed`, `redb`, `BTreeMap`, etc).

I'm certain it would have use cases, if the overhead is a problem I don't think adding `get_range_{keys,values}()` would be that bad.

database/src/database.rs
hinto-janai: The main reason is because `?` no longer works for callers, aka, any non-`service` user.

You can no longer use `?` because the error might be a non-existent key (not really a hard error) or an _actual_ database error, so all callers must:
```rust
fn get_block(db: &Database) -> Result<Option<Block>, Error> {
    match db.get(&key) {
        Ok(v) => Ok(Some(v)),
        Err(RuntimeError::KeyNotFound) => Ok(None),
        Err(e) => Err(e),
    }
}
```
`Option` is probably used in most APIs because it represents the problem more accurately and allows `?`:
```rust
fn get_block(db: &Database) -> Result<Option<Block>, Error> {
    Ok(db.get(&key)?)
}
```
With `Option`, yes `cuprate_database::service` will have to map it but taking away `?` from non-`service` users and forcing `match` at all call sites is also quite bad. Although, I guess it does make usage within Cuprate easier.

database/src/backend/redb/database.rs
hinto-janai: Actually, this is hard to impl actually since `redb` doesn't allow deleting tables that are open, [it returns `TableAlreadyOpen`](https://docs.rs/redb/2.0.0/src/redb/transactions.rs.html#380-382).

If our custom `RedbTable` itself holds onto the table we want to delete, we have to do some swapping like so:

1. swap our `self.db` table with a temporary table
2. delete the target table
3. create the target table, set `self.db` to it again
4. delete the temporary table

Only issue is that this is not `Drop`-safe, as in if we early return or panic, our `self.db` is set to the invalid temporary table. Creating a special struct and doing `fn drop() { /* reset self.db */ }` is not good either since that can error too.

I think there are only 2 safe alternatives:
1. Use `redb::Table::retain(self, false)` which hopefully compiles down efficiently since the predicate is known
2. Just remove `DatabaseRw::clear()` from this PR

database/src/backend/heed/database.rs
Boog900: I'm unsure, the only tables `get_range` should really be called are tables which have indexes for keys i.e. integers with no gaps.

It doesn't really make sense to use this on a table which has hashes for keys.

So the keys would already be known by the caller

database/src/backend/redb/database.rs
Boog900: could we move this to env? then we would be able to spawn a tx to remove a table

database/src/database.rs
Boog900: I would rather keep it an error.

We will still need the error variant for `KeyNotFound` for the service and we will be wrapping the return in an option to then "unwrap" it. 

Up to you though.

database/src/backend/heed/database.rs
hinto-janai: Reverted back to only values being returned.

database/src/backend/redb/database.rs
hinto-janai: Sure, moved to `EnvInner::clear_db()`.

database/src/database.rs
hinto-janai: Will keep it as is since both have tradeoffs - no `Option`.

database/src/database.rs
```diff
@@ -29,7 +29,14 @@ pub trait DatabaseRo<T: Table> {
...
 
-    /// Get an iterator of values corresponding to a range of keys.
+    /// Get an iterator of value's corresponding to a range of keys.
+    ///
+    /// For example:
+    /// ```rust,ignore
+    /// // This will return all 100 tuples of `(key, value)` where
+    /// // `key` is `0..100` and `value` is the corresponding value.
+    /// self.get_range(0..100);
+    /// ```
```
Boog900, line 39: I think this comment needs to be updated

database/src/backend/redb/env.rs
```diff
@@ -202,6 +202,29 @@ where
...
+        // which will panic in the `From` impl, as:
+        //
+        // 1. Only 1 `redb::WriteTransaction` can exist at a time
+        // 2. We have exclusive access to it
+        // 3. So it's not being used to open a table since that needs `&tx_rw`
+        //
+        // Reader-open tables do not affect this, if they're open the below is still OK.
+        redb::WriteTransaction::delete_table(tx_rw, table)?;
+
+        Ok(())
```
Boog900, line 226: I think we will have to recreate the table after deleting it right?

Boog900 (changes requested):

database/src/backend/redb/env.rs
hinto-janai: Right, write-tables will be automatically created if they don't exist but `open_db_ro()` would panic after this... my bad.

This function isn't atomic I guess, if `delete_table()` succeeds but `open_table()` errors and returns early, the table is left uncreated. We could `.unwrap()` the table creation since it'll eventually panic anyway?

Boog900: >This function isn't atomic I guess

This is probably fine

Boog900 (approved):

PR #97, opened 2024-03-26
database: add `redb-memory` backend

hinto-janai:
## Part 17
https://github.com/Cuprate/cuprate/pull/96 <- Previous

- Adds a new `redb-memory` feature flag that uses the same code from `src/backend/redb`, except that it uses `redb::backend::InMemoryBackend` instead which resides completely in memory instead of on disk

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,7 @@ jobs:
       run: |
         cargo test --all-features --workspace
         cargo test --package cuprate-database --no-default-features --features redb --features service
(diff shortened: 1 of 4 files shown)


SyntheticBird45:
What would be the use case exactly ?

hinto-janai:
Solely for debugging and as a benchmark control group I guess.

It's only a few lines to add but it's not 100% necessary.

Boog900 (approved):
Yeah this is small enough and could be useful, would be cool to have all back ends configurable at runtime

PR #101, opened 2024-04-01
database: impl `service::{Request,Response}` mappings

hinto-janai:
## Part 19
https://github.com/Cuprate/cuprate/pull/98 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/102

- Updates to `cuprate_types::service::{Request,Response}` types
- Adds `fn` signatures for the `service` reader/writer threads (impl after https://github.com/Cuprate/cuprate/pull/102 is merged)
- Implements `service` reader/writer thread request/response/fn mapping logic
- Makes `Env::resize_map()` return the new memory map size


database/src/service/write.rs
```diff
@@ -141,10 +146,39 @@ impl DatabaseWriter {
...
+
+                        request_retry_counter += 1;
+                        self.resize_map();
+                        continue;
+                    }
+                    Err(error) => response_sender.send(Err(error)).unwrap(),
+                }
+
+                break;
             }
```
hinto-janai, line 182: @Boog900 How do you think we should do retries on `WriteRequest`'s?

The current method of _reacting_ to resize errors instead of _proactively_ resizing comes with the downside in that we must `clone()` the data we use since we may not know if a resize is needed and the next `loop` iteration needs the data.

I propose we keep `RuntimeError::ResizeNeeded` but `DatabaseWriter` keeps track of the memory map byte size in some local `usize` counter, which we do book-keeping on each request:

1. `WriteRequest` received
2. Calculate how many bytes it'll write (with some buffer in-case?)
3. Check if our `current_memory_map_bytes: usize` needs to be resized
4. Resize if needed
5. Call write function
6. Panic if it errors with `ResizeNeeded` as we just handled that

Similar to: https://github.com/monero-project/monero/blob/c8214782fb2a769c57382a999eaf099691c836e7/src/blockchain_db/lmdb/db_lmdb.cpp#L665

hinto-janai: Nvm... ignore this.

Our `DatabaseRw::put` takes `&Key` not `Key`, so our write function will just take a reference, and we can be reactive when handling resize errors.

hinto-janai (commented):
@Boog900 As `add_block()` will take `VerifiedBlockInformation` by value so it can just take the inner data instead of cloning, the resize retry problem here is fixed using https://github.com/Cuprate/cuprate/pull/101#discussion_r1545879735.

Another option is just setting the memory map to a large value at startup (e.g. `16TB`) such that this error never occurs - I'm not sure what the overhead in doing this is though.

Boog900:
@hinto-janai Why does add block need to take by value? it should be able to just reference the data

Boog900:
I looked into setting the memory map size high and it supposedly works on linux but I found a few people mentioning problems on windows/ other platforms, maybe these have been fixed but I would rather not do this if we don't have to.

hinto-janai:
> Why does add block need to take by value? it should be able to just reference the data

Mostly because block blobs is `StorableVec`, which is just `StorableVec(Vec<T>)` so we can't pass a `&Vec<T>` here: https://github.com/hinto-janai/cuprate/blob/1a11938701df3be9165e45078b94f0d44b212d22/database/src/ops/block.rs#L90-L93

I guess we could clone it inside `add_block()` but that means callers don't get to decide when to clone, i.e. even if they don't need the `VerifiedBlockInformation` after `add_block()` it'll still get cloned.

Boog900:
> Mostly because block blobs is StorableVec, which is just StorableVec(Vec<T>)

you can ref cast: https://docs.rs/bytemuck/latest/src/bytemuck/transparent.rs.html#142-155

hinto-janai:
Right... I forgot, will make `add_block()` take by ref.

I'll revert it so `RuntimeError::ResizeNeeded` is handled reactively too. Does it make sense to retry writes a 3rd time? If the 2nd try did not resize correctly something is surely wrong and we should panic, right?

Boog900:
I don't think there is anything wrong with trying 3 times though, better to go to high than to low IMO

database/src/service/read.rs
```diff
@@ -199,30 +200,45 @@ impl tower::Service<ReadRequest> for DatabaseReadHandle {
...
+        R::BlockExtendedHeaderInRange(range) => block_extended_header_in_range(&env, range),
+        R::ChainHeight => chain_height(&env),
+        R::GeneratedCoins => generated_coins(&env),
+        R::Outputs(map) => outputs(&env, map),
+        R::NumberOutputsWithAmount(vec) => number_outputs_with_amount(&env, vec),
+        R::CheckKIsNotSpent(set) => check_k_is_not_spent(&env, set),
+        R::BlockBatchInRange(range) => block_batch_in_range(&env, range),
+    };
+
+    // TODO: what do we do if this errors?
```
SyntheticBird45, line 227: replace by `expect("msg")` if you feel like it. It should never happen

database/src/service/read.rs
```diff
@@ -199,30 +200,45 @@ impl tower::Service<ReadRequest> for DatabaseReadHandle {
...
+    response_sender.send(response).unwrap();
 
     /* TODO: post-request handling, run some code for each request? */
```
SyntheticBird45, line 230: Maybe some performance counter ?
- an atomic counter of database requested handled since uptime
- some timer for measuring how much time the request take ? Maybe tracing would be enough for this, but I remember they postpone some of their writing.

hinto-janai: We could instrument manually and make `cuprate_database::service::init()` return some shared state like:
```rust
use std::time::Instant;

struct DatabaseStats {
    read_requests: SomeList<Instant>, // .len() == total requests
    write_requests: SomeList<Instant>,
    init_instant: Instant,
    other_stats: [...],
}

pub fn init(config: Config) ->
    Result<(
        DatabaseReadHandle,
        DatabaseWriteHandle,
        Arc<DatabaseStats> // todo: make this read-only
    ), InitError>;
```
but I think `tracing` will mostly cover everything and be better? I'm not sure.

database/src/service/read.rs
Boog900: no the error should just be ignored, if for example a p2p peer requests some data and then disconnects the oneshot receiver will be dropped and this will panic

database/src/service/read.rs
Boog900: yeah tracing should handle this

hinto-janai:
Sorry, should be ready now.

database/src/env.rs
```diff
@@ -113,7 +113,7 @@ pub trait Env: Sized {
     /// This function _must_ be re-implemented if [`Env::MANUAL_RESIZE`] is `true`.
     ///
     /// Otherwise, this function will panic with `unreachable!()`.
-    fn resize_map(&self, resize_algorithm: Option<ResizeAlgorithm>) {
+    fn resize_map(&self, resize_algorithm: Option<ResizeAlgorithm>) -> NonZeroUsize {
```
SyntheticBird45, line 116: We should add to the fn's doc that it return the new size and not the old one.

database/src/service/write.rs
```diff
@@ -9,16 +9,16 @@ use std::{
...
 
+use crate::{error::RuntimeError, ConcreteEnv, Env};
+
 //---------------------------------------------------------------------------------------------------- Constants
 /// Name of the writer thread.
-const WRITER_THREAD_NAME: &str = "cuprate_helper::service::read::DatabaseWriter";
+const WRITER_THREAD_NAME: &str = concat!(module_path!(), "::DatabaseWriter");
 
 //---------------------------------------------------------------------------------------------------- Types
 /// The actual type of the response.
```
SyntheticBird45, line 24: I see we've redefined these types in write.rs and read.rs. Just to be sure, are we not writing these as `pub (self) type` for keeping only mod import in `mod.rs` ?

SyntheticBird45 (approved):

database/src/service/write.rs
hinto-janai: Wow good catch they are the same... moved into `types.rs` so both can share.

database/src/service/read.rs
```diff
@@ -193,28 +175,38 @@ impl tower::Service<ReadRequest> for DatabaseReadHandle {
...
+    let response = match request {
+        R::BlockExtendedHeader(block) => block_extended_header(&env, block),
+        R::BlockHash(block) => block_hash(&env, block),
+        R::BlockExtendedHeaderInRange(range) => block_extended_header_in_range(&env, range),
+        R::ChainHeight => chain_height(&env),
+        R::GeneratedCoins => generated_coins(&env),
+        R::Outputs(map) => outputs(&env, map),
+        R::NumberOutputsWithAmount(vec) => number_outputs_with_amount(&env, vec),
+        R::CheckKIsNotSpent(set) => check_k_is_not_spent(&env, set),
+        R::BlockBatchInRange(range) => block_batch_in_range(&env, range),
```
Boog900, line 204: mentioned this in the other PR but `BlockBatchInRange` doesn't need to exist, it was a hack to get the RPC scanner to work, missed this when in the `types` PR.

database/README.md
```diff
@@ -108,6 +108,7 @@ This folder contains the `cupate_database::service` module.
 | `request.rs`   | Read/write `Request`s to the database
 | `response.rs`  | Read/write `Response`'s from the database
```
Boog900, line 109: I think these need to be removed.

```suggestion
```

Boog900 (approved):

Issue #82, opened 2024-02-29
Support for automated verification tools


SyntheticBird45:
## Support for automated verification tools

While Cuprate is almost entirely written in safe Rust, it is still a huge codebase. Such size could lead us to have introduced, without knowing, the very far corner cases of safe Rust, that could lead to [memory unsafety and undefined behavior](https://github.com/Speykious/cve-rs). 

It might be beneficial to add support for the following automated verifiers:
- [Miri](https://github.com/rust-lang/miri). Rust's official MIR verifier, provide example of CI integration
- [Kani](https://github.com/model-checking/kani). The bit-precise model checker for Rust. Focused on memory safety and assertions check
- [Loom](https://github.com/tokio-rs/loom). Tokio's concurrency verifier. Check all possible interleaving for race conditions

Honorable mention: [Proptest](https://github.com/proptest-rs), fuzzing of properties framework. Can be used directly as a `dev.dependencies` for `#[test]`

If this is adopted with [reproducible builds](https://github.com/Cuprate/cuprate/issues/39), this will make Cuprate able to deliver deterministic and fully verified binaries.

PR #102, opened 2024-04-01
database: implement `ops/`

hinto-janai:
## Part 20
https://github.com/Cuprate/cuprate/pull/101 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/104

- Implements the entire `cuprate_database::ops` module; abstracted Monero-related database functions


hinto-janai:
@Boog900 Re: https://github.com/Cuprate/cuprate/pull/35#discussion_r1486317218

Do you still want these to be free functions? I think a pre-implemented superset trait of our current `trait DatabaseR{o,w}` would make more sense - the `self` in each trait would be an `impl DatabaseR{o,w}` instead of having to specify the generic on each free function e.g:

```rust
// free functions
fn get_block<T: Table, D: DatabaseRo<T>>(db: &D, block: u64) -> Block;
fn get_block_from_height<T: Table, D: DatabaseRw<T>>(db: &D, height: u64) -> Block;
fn pop_block<T: Table, D: DatabaseRw<T>>(db: &D, block: u64) -> Block;

// caller
let db = env_inner.open_table_ro::<Blocks>(&tx_ro);
let block = get_block<Blocks, /* we'd have to name a backend specific type here? */>(&db, 0);
```
```rust
// super-set traits (all fn's pre-implemented)
trait MoneroRo<T: Table>: DatabaseRo<T> {
    fn get_block(&self, block: u64) -> Block;
    fn get_block_from_height(&self, height: u64) -> Block;
}

trait MoneroRw<T: Table>: DatabaseRw<T> {
    fn pop_block(&self, block: u64) -> Block;
}

// caller
let db = env_inner.open_table_ro::<Blocks>(&tx_ro);
let block = db.get_block(0);

**Comment by Boog900** [comment]
Yeah, the caller shouldn't need to specify the generics rust should be able to infer them (I think). I think these make more sense as functions as they shouldn't be re-implemented, I don't think specifying the generics on each function is too bad.

**Comment by hinto-janai** [comment]
> the caller shouldn't need to specify the generics rust should be able to infer them

Okay with free functions too but it's worth noting in any call-site where the type must be specified it will essentially tie you to a specific backend - making it impossible to call from `cuprate_database`'s perspective.

We'll see if there are inference issues when all these functions are done, if there are, we could use a trait? The bodies will be the same, so it'll be easy to switch.

File: database/src/ops/block.rs [hunk 1]
```diff
@@ -1,89 +1,398 @@
...
+    height: BlockHeight,
+    block: &BlockInfoLatest,
+) -> Result<(), RuntimeError>
+where
+    Ro: TxRo<'env>,
+    Rw: TxRw<'env>,
+    Env: EnvInner<'env, Ro, Rw>,
+{
+    add_block_v3(env, tx_rw, height, block)
 }
```
hinto-janai, line 31: There's a function per `BlockInfoV*` type and a single `add_block()` that forwards to the latest type, `BlockInfoV3` in this case.

This is the case for `add_block()` as well - is this ok?

database/src/ops/block.rs
```diff
@@ -1,89 +1,398 @@
...
+    Ro: TxRo<'env>,
+    Rw: TxRw<'env>,
+    Env: EnvInner<'env, Ro, Rw>,
+{
+    let mut table = env.open_db_rw::<BlockInfoV3s>(tx_rw)?;
+    for (height, block) in height_and_blocks {
+        table.put(height, block)?;
+    }
+    Ok(())
+}
```
hinto-janai, line 172: These bulk functions are needed right?

Otherwise calling `get_block()` in a loop would mean opening+closing the table in a loop, where as in this function it's done only once.

database/src/ops/block.rs
```diff
@@ -1,89 +1,398 @@
...
+    tx_rw: &mut Rw,
+) -> Result<(BlockHeight, BlockInfoLatest), RuntimeError>
+where
+    Ro: TxRo<'env>,
+    Rw: TxRw<'env>,
+    Env: EnvInner<'env, Ro, Rw>,
+{
+    todo!(); // remove related info?
+    env.open_db_rw::<BlockInfoV3s>(tx_rw)?.pop_last()
+}
```
hinto-janai, line 190: Not exactly sure what else to remove in our tables.

database/src/ops/block.rs
```diff
@@ -1,89 +1,398 @@
...
-pub fn get_top_block() {
+pub fn get_block_cumulative_rct_outputs() {
     todo!()
 }
 
 /// TODO
-pub fn get_top_block_hash() {
+pub fn get_block_from_height() {
     todo!()
 }
```
hinto-janai, line 398: These functions that are basically `get_block()` -> return some field - should they exist?

There is already `get_block()`, users can just `get_block()?.field` right?

hinto-janai (commented):
@Boog900 Some questions.

database/src/ops/block.rs
Boog900: yeah I would say so but I don't think `add_blocks_v3` needs to be a bulk function just `add_blocks`

database/src/ops/block.rs
Boog900: This should remove everything in the block, all block info, all txs in the blocks and all tx info from the DB

SyntheticBird45:
> Adding block info should just be 1 function which routes to the different tables for example.

I agree. API should be descriptive.

SyntheticBird45:
I believe monerod made fine-grained database API for various reasons. The first I see is that the almost monolithic operations all required slight changes compared to their predecessors. At each hard fork a need for new operations arise, not divergent enough to rethink the entire API, but not close enough either to justify being in the same function. The second is that some of these APIs were used once or twice accross the codebase, for very specific purposes, and it just didn't make sense to put that needed routines under another umbrella.

I think Cuprate abstractions and architecture alleviate this behavior. So we can put more routines under the same calls.

Boog900:
A lot of monerod's database oddities are from when the DB was in RAM, so these functions would have been a lot cheaper.

database/src/ops/block.rs
hinto-janai: Which field is this? Something like:

```rust
pub fn add_block<'env, Ro, Rw, Env>(
    env: &Env,
    tx_rw: &mut Rw,
    block: cuprate_consensus::block::VerifiedBlockInformation
) -> Result<(), RuntimeError>
where
    Ro: TxRo<'env>,
    Rw: TxRw<'env>,
    Env: EnvInner<'env, Ro, Rw>,
{
    match block.block.header.major_version {
        0..v1 => add_to_v1_table(block),
        v2..v3 => add_to_v2_table(block),
        v3 => add_to_v3_table(block),
        _ => ???
    }
}
```

database/src/ops/block.rs
hinto-janai: So just `add_block()` and `add_blocks()` is needed, both which branch and add to the appropriate table depending on the hard fork?

database/src/ops/block.rs
Boog900: yp but v3 doesn't mean hf 3 btw:

v1 = up to HF 4  
v2 = HF 4 to 10
v3 = HF 10 onwards

database/src/ops/block.rs
Boog900: yep, also I think `add_block` should do more than just add to the block info table, it should fill in everything for that block. Each tx/ output etc.

database/src/ops/block.rs
```diff
@@ -16,18 +25,149 @@ use crate::{
...
+        let mut table_pruned_tx_blobs = env.open_db_rw::<PrunedTxBlobs>(tx_rw)?;
+        let mut table_prunable_hashes = env.open_db_rw::<PrunableHashes>(tx_rw)?;
+        let mut table_outputs = env.open_db_rw::<Outputs>(tx_rw)?;
+        let mut table_prunable_tx_blobs = env.open_db_rw::<PrunableTxBlobs>(tx_rw)?;
+        let mut table_rct_outputs = env.open_db_rw::<RctOutputs>(tx_rw)?;
+        let mut table_tx_ids = env.open_db_rw::<TxIds>(tx_rw)?;
+        let mut table_tx_heights = env.open_db_rw::<TxHeights>(tx_rw)?;
+        let mut table_tx_unlock_time = env.open_db_rw::<TxUnlockTime>(tx_rw)?;
+
+        for tx in block.txs {
```
SyntheticBird45, line 123: That is very annoying. Wouldn't `iter()`ate over `txs` at every table smh possible ? I wonder what would be the worst performance scenario between reopening each table every transactions, or the reiterating over each transactions at every table.

SyntheticBird45: I also feel that you will need to declare sub-functions for keeping clarity. Like monerod did. There are a lot of tables to deal with...

hinto-janai: Yeah, this is distinctly a problem with `cuparate_database` as:
- `redb`'s table type itself implements [`put()` with `&self`](https://docs.rs/redb/latest/redb/struct.Table.html#method.insert)
- `heed`'s table type requires a [`&mut TxRw` as input](https://docs.rs/heed/0.20.0-alpha.9/heed/struct.Database.html#method.put)

and in order to abstract over both, our custom `HeedTableRw` has to hold onto a `&mut TxRw`.

> I also feel that you will need to declare sub-functions for keeping clarity

If these tables are only added to here, I think keeping it one body is more clear.

hinto-janai: We could make our current `trait DatabaseR{o,w}` functions take in a TX instead?

@Boog900 what do you think?

Boog900: redb & heed both allow opening a table with a &tx. The problem for us is that because redb tracks the tx in the table and heed doesn't we need to track the tx for heed and it needs to be mutable.

Knowing this we could wrap heeds rw txs in a RefCell (an UnsafeCell would also be fine but the performance difference would probably be negligible here).

So the method to open could be moved to take an &tx and heeds DB would store a `&RefCell<RwTx>` (notice not a mutable reference) and when it comes time to use the tx it gets a multiple reference.

This should be fine as long as:
- Txs are not `Sync` (they are not)
- A table operation does not hold a reference to the inner cell once the call is over, unless the return also holds a lifetime (This is also true)

@hinto-janai what do you think?

Boog900: I sent the message before seeing yours but it still stands.

> We could make our current trait DatabaseR{o,w} functions take in a TX instead?

this would remove the atomicity of adding a block right? if we make a tx per table ...

hinto-janai: RefCell causes temporary value issues since it creates a temporary `Ref<'_>` before getting to the `heed::RoTx`.
```rust
fn get_range<'a, Range>(
    &'a self,
    range: Range,
) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
where
    Range: RangeBounds<T::Key> + 'a,
{
    //                 cannot return, temporary value
    //                         created here
    //                              v
    get_range::<T, Range>(&self.db, &self.tx_rw.borrow(), range)
}
```
If underlying `get_range()` is changed, it'll error as well, it'll force our `TxRo` to be using `RefCell` as well too.

Could use [`unsafe { try_borrow_unguarded() }`](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.try_borrow) to get around this.

> this would remove the atomicity of adding a block right?

I was thinking 1 write per all table but `open_db_rw(&mut TxRw)` wouldn't allow that - and yeah we wouldn't be able to `commit()` multiple transactions atomically...

Boog900: can we inline these functions? they are all one liners.

Boog900: wait no, that wont work :(

hinto-janai: This works, is it safe?
```rust
fn get_range<'a, Range>(
    &'a self,
    range: Range,
) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
where
    Range: RangeBounds<T::Key> + 'a,
{
    // SAFETY:
    // - our lifetime is bounded to `&'a self`
    // - `&self` means there are no mutable references
    // - we're creating/dereferencing a pointer to a good reference
    Ok(self
        .db
        .range(unsafe { &*self.tx_rw.as_ptr() }, &range)?
        .map(|res| Ok(res?.1)))
}
```

Boog900: I would just use `UnsafeCell` but yes that should be safe

Boog900: I would really rather used RefCell for runtime checks though rather than unsafe if you can think of a way ....

Boog900: this seems like a value guard issue again, I can't see an easy way to use RefCell so UnsafeCell it is I guess.

I will look at integrating miri into CI if we do decide to do this (#82)

Boog900: I have realized this is not safe!

Although it would be safe if we just had open a single table if we had open 2 tables made and iterator out of one and put values into the other while the iterator is still live we would have an active mutable and immutable reference to the same thing.

I don't think we can solve this, with `RefCell` it would cause a panic so it wouldn't be too bad but we cant use `RefCell`.

We could just ban the use of iterators with RW txs, I can't see a use for RW iters in Cuprate so it wouldn't be too much of a loss, this would also allow us to use `RefCell` again.

hinto-janai: How do we feel about just leaving this as is?

The cons are:
1. Performance
2. Cumbersome for callers

Counter-arguments:
1. Based on these [benchmarks](https://gist.github.com/hinto-janai/546d6308643f515a95ec749445963464), even using `redb`'s `open_db_rw()`, opening+closing a table takes `0.000000264s`, or `1.319s` for `5,000,000` calls - it'll be a tiny fraction of total time spent in something like `add_block()`
2. Users of `cuprate_database` aren't mutating tables directly anyway, they are using `ops/` - so _we_, `cuprate_database` itself, are the only ones dealing with these uglier calls - which I'm fine with writing/maintaining

Boog900: It's not awful but I would still prefer to not have this restriction.

If it makes it nicer we could also change the definition of the DatabaseRw from having a requirement of DatabaseRo to instead defining the read functions possible for DatabaseRw in the trait itself.

SyntheticBird45: 264ns isn't a lot on its own. If we take [this block for example](https://xmrchain.net/block/3119473), It'll be 9*264*156 =  370656 ns or 370 µs waste at opening tables. So yes this isn't a lot, but I'm genuinely suggesting opening tables once and iterating over `txs` 9 times for a total of 9*156 tx instead of opening 9*156 tables.

hinto-janai: https://github.com/Cuprate/cuprate/pull/104

Boog900:
Why did this close? weird

database/src/tables.rs
```diff
@@ -12,6 +13,184 @@ use crate::{
...
+    fn tx_ids_mut(&mut self) -> &mut impl DatabaseRw<TxIds> {
+        &mut self.12
+    }
+    fn tx_heights_mut(&mut self) -> &mut impl DatabaseRw<TxHeights> {
+        &mut self.13
+    }
+    fn tx_unlock_time_mut(&mut self) -> &mut impl DatabaseRw<TxUnlockTime> {
+        &mut self.14
+    }
+}
```
hinto-janai, line 192: 1. Trait is defined that gives access to all tables, blanket impl is provided for `(a, tuple, containing, all, open, database, types, ...)`

database/src/env.rs
```diff
@@ -222,6 +227,54 @@ where
...
+            EnvInner::open_db_rw::<PrunedTxBlobs>(self, tx_rw)?,
+            EnvInner::open_db_rw::<PrunableHashes>(self, tx_rw)?,
+            EnvInner::open_db_rw::<Outputs>(self, tx_rw)?,
+            EnvInner::open_db_rw::<PrunableTxBlobs>(self, tx_rw)?,
+            EnvInner::open_db_rw::<RctOutputs>(self, tx_rw)?,
+            EnvInner::open_db_rw::<TxIds>(self, tx_rw)?,
+            EnvInner::open_db_rw::<TxHeights>(self, tx_rw)?,
+            EnvInner::open_db_rw::<TxUnlockTime>(self, tx_rw)?,
+        ))
+    }
```
hinto-janai, line 276: 2. `trait EnvInner` gives access to this, but instead of returning the messy tuple, it returns a nice clean `impl TablesMut` object - this is also more type safe as the tuple is bounded by the blanket impl from before - these functions are also pre-implemented since they aren't backend specific

database/src/ops/block.rs
```diff
@@ -1,89 +1,422 @@
...
+}
+
+/// Internal function that [`add_block()`] and [`add_blocks()`] uses.
+///
+/// It takes the already opened tables.
+#[allow(clippy::cast_possible_truncation)] // TODO: remove me
+fn add_block_inner(
+    tables: &mut impl TablesMut,
+    block: &VerifiedBlockInformation,
+) -> Result<(), RuntimeError> {
```
hinto-janai, line 76: 3. The (private) function that does the real work takes in an `impl TablesMut`, essentially acting as a anonymous struct that has access to all tables (which is exactly what it is, a tuple)

database/src/ops/block.rs
```diff
@@ -1,89 +1,422 @@
...
+    Ro: TxRo<'env>,
+    Rw: TxRw<'env>,
+    Env: EnvInner<'env, Ro, Rw>,
+{
+    let mut tables = env.open_tables_mut(tx_rw)?;
+    for block in blocks {
+        add_block_inner(&mut tables, block)?;
+    }
+    Ok(())
+}
```
hinto-janai, line 67: 4. Now, our public "frontend" functions just open up the tables and pass it to `add_block_inner()` - and it is still efficient since we only open them once, even in the bulk case

database/src/ops/block.rs
```diff
@@ -1,89 +1,422 @@
...
+                block_hash: block.block_hash,
+                cumulative_rct_outs: todo!(),                    // TODO
+                long_term_weight: block.long_term_weight as u64, // TODO
+            },
+        )
+    }?;
+
+    // BlockBlobs: BlockHeight => BlockBlob
+    // TODO: what is a block blob in Cuprate's case?
+    tables.block_blobs_mut().put(&block.height, todo!())?;
```
hinto-janai, line 123: Btw, this is what the usage looks like.

`tables` here is an `impl TablesMut`, we're getting access to `&mut impl DatabaseRw<BlockBlobs>` via `TablesMut::block_blobs_mut()` here then calling `put()`.

hinto-janai (commented):
@Boog900 @SyntheticBird45 Thoughts on this?

This is to tackle the problem of a whole bunch of unwieldy `trait DatabaseRw<Table>` types being passed around.

To encapsulate all tables, we pass around an `impl Trait` object that is essentially an anonymous struct that has access to all (already opened) tables.

This was originally a macro only defined in `ops/blocks.rs`:
```rust
macro_rules! open_db_rw_all_or_return {
    ($env_inner:ident, $tx_rw:ident) => {
        (
            EnvInner::open_db_rw::<Table>($env_inner, $tx_rw)?,
            EnvInner::open_db_rw::<Table2>($env_inner, $tx_rw)?,
            EnvInner::open_db_rw::<Table3>($env_inner, $tx_rw)?,
            // [...]
        )
    };
}
```
but since this will be used elsewhere I thought more type safe functions would be more appropriate.

SyntheticBird45:
This is a good idea. The semantic benefits are undeniable. The use of `impl Trait` is necessary as it is (performance wise) better to let the compiler duplicate than using VTables like `dyn Trait` do. 

I'll take a look back but despite being unusual, I don't see any significant caveats to this approach. I would just ask for an effort on the documentation side, so new comers can better grasp the use `trait TablesMut`. A paragraph summarizing the callflow you just explained would be great.

Boog900:
I like the idea but that trait impl is not pretty, is a macro possible here?

I would also make the macro produce a free function that does what `open_tables_mut` and just have `open_tables_mut` call this function

database/src/env.rs
SyntheticBird45, line 276: @Boog900 is it this part that you wish to replace with a macro ?

Boog900: I want to replace the Ok parts here with a function call generated by a macro and the `Tables(Mut)` definition with a macro as well

SyntheticBird45: I don't understand what you mean by *function call generated by a macro*. Why would you want to add another function inside this one ?

Boog900: Because currently this is confusing to change if we were to add/ delete tables.

If instead this was all generated by a macro then it would be as simple as adding another argument/ removing an argument from the macro

hinto-janai: @Boog900 I think I understand - all `$table:ty`'s should be defined in some `output_table_types!()` which just outputs a list of all table types - and that should be used as input everywhere?

I'm not sure how to do with this `macro_rules!()` as the output must be valid syntax - this isn't allowed:
```rust
macro_rules! output_table_types {
    () => {
        crate::tables::BlockInfoV1s,
        crate::tables::BlockInfoV2s,
        crate::tables::BlockInfoV3s,
        [...]
    };
}
```
I'll try finding something that works - but the idea of passing around an `impl TablesMut` that basically is a struct that holds our tables is OK?

database/src/tables.rs
```diff
@@ -12,6 +13,76 @@ use crate::{
...
+    NumOutputs => 6,
+    PrunedTxBlobs => 7,
+    PrunableHashes => 8,
+    Outputs => 9,
+    PrunableTxBlobs => 10,
+    RctOutputs => 11,
+    TxIds => 12,
+    TxHeights => 13,
+    TxUnlockTime => 14,
+}
```
hinto-janai, line 84: This macro defines and blanket impls both `trait {Tables,TablesMut}`.

IMO I prefer manual impl for everything. For something that won't be changed often, the more verbose version is more clear in my opinion - type errors will still pop up when editing as well so it's still safe.

Boog900: It's not really a huge issue but I still think the macro is nicer, I don't think there is much difference in the clarity of the macro vs the verbose option, especially with docs.

SyntheticBird45: Fwiw, I also prefer the manual impl over the macro one.

database/src/ops/output.rs
```diff
@@ -1,34 +1,213 @@
...
+}
+
+#[doc = doc_inner!(get_pre_rct_num_outputs)]
+#[inline]
+pub fn get_pre_rct_num_outputs_inner(
+    table_outputs: &(impl DatabaseRo<Outputs> + DatabaseIter<Outputs>),
+) -> Result<u64, RuntimeError> {
+    // TODO: is this correct?
+    table_outputs.len()
 }
```
hinto-janai, line 213: @Boog900 @SyntheticBird45 

Here's the current pattern for functions in `ops/` - each function comes in 3 flavors:
- `fn()` -> inputs/outputs 1 thing (opens tables for caller)
- `fn_bulk()` -> inputs/outputs N things (opens tables for caller)
- `fn_inner()` -> is the actual implementation (takes in already opened tables)

For example: `get_block()`, `get_block_bulk()`, `get_block_inner()`.

Some open questions:
1. Should `fn()` just be replaced with `_inner()` - as in, should we make the _caller_ open the tables and pass them instead? It would rid the need of an `_inner()`
2. Should these `_inner()` functions be public? Here's an example of what it would look like:

image

hinto-janai: IMO I think the answer to 1 should be yes - we should take the already opened tables instead of `EnvInner + Tx`.

- The function only takes what it needs
- More flexible for callers
- No lifetime + generics on all of our functions

Only tradeoff I think is there is there's more input set-up for the caller where many tables are passed - thankfully our `impl Tables` saves us from naming all tables:
```rust
//--------------------------------------- Before
pub fn get_block<'env, Ro, Rw, Env>(
    env: &Env,
    tx_ro: &Ro,
    height: BlockHeight,
) -> Result<VerifiedBlockInformation, RuntimeError>
where
    Ro: TxRo<'env>,
    Rw: TxRw<'env>,
    Env: EnvInner<'env, Ro, Rw>;

//--------------------------------------- After
pub fn get_block(
    tables: &impl Tables,
    height: BlockHeight,
) -> Result<VerifiedBlockInformation, RuntimeError>;
```

SyntheticBird45: 2. No. It makes more sense that `fn_inner()` stays private. Even if its cheaper, then I don't see why we would use`fn()`. We could simply enforce opening everything required before  `fn()`. It would simply be moving the problem... So I public `fn()` needs to be defined, that internally use the `fn_inner()`

SyntheticBird45: Otherwise the functions flavors looks good and intuitive for me. Overall design is good :+1:

hinto-janai: To be clear I don't like the flavors - I want the caller to handle opening stuff so that there's only 1 minimal/flexible function.

We could remove the `_bulk()` functions as well since the caller is now free to `.collect()` into whatever container they want themselves since the function is cheap to call:
```rust
// Env.
let env = todo!();
let env_inner = env.env_inner()?;
let tx_ro = env_inner.tx_ro()?;

// Set-up the tables.
let tables: impl Tables = env_inner.open_tables(&tx_ro)?;

// Pass table directly and collect into whatever.
// This iterator returns early on error.
let blocks: Result<Vec<Block>, Error> = (0..100_000)
    .map(|height| get_block(height, &tables))
    .collect();
```

SyntheticBird45: I don't understand. What do you call the caller ?

Boog900: I agree we should just make the caller open the tables.

> We could remove the _bulk() functions as well since the caller is now free to .collect() into whatever container they want themselves since the function is cheap to call

Seems like a good idea. 

BTW If we are going to use `rayon`s parallel iterators we will probably need to use a [ThreadLocal]( https://docs.rs/thread_local/latest/thread_local/struct.ThreadLocal.html) which holds the tables as txs are not sync 

> What do you call the caller

The caller would be the service functions

database/src/ops/block.rs
```diff
@@ -1,89 +1,321 @@
...
+                timestamp: block.block.header.timestamp,
+                total_generated_coins: block.generated_coins,
+                weight: block.weight as u64, // TODO
+                cumulative_difficulty: block.cumulative_difficulty,
+                block_hash: block.block_hash,
+                cumulative_rct_outs: todo!(),                    // TODO
+                long_term_weight: block.long_term_weight as u64, // TODO
+            },
+        )
+    }?;
```
hinto-janai, line 95: What to do with the size difference between `VerifiedBlockInformation` and our `BlockInfoV*` types?

Casting here is lossy?

database/src/ops/block.rs
```diff
@@ -1,89 +1,321 @@
...
+    } else {
+        tables.block_info_v3s_mut().put(
+            &block.height,
+            &BlockInfoV3 {
+                timestamp: block.block.header.timestamp,
+                total_generated_coins: block.generated_coins,
+                weight: block.weight as u64, // TODO
+                cumulative_difficulty: block.cumulative_difficulty,
+                block_hash: block.block_hash,
+                cumulative_rct_outs: todo!(),                    // TODO
```
hinto-janai, line 91: Where is `cumulative_rct_outs` found?

database/src/ops/block.rs
```diff
@@ -1,89 +1,321 @@
...
+        )
+    }?;
 
-/// TODO
-pub fn block_exists() {
-    todo!()
-}
+    // Block blobs.
+    // TODO: what is a block blob in Cuprate's case?
+    tables.block_blobs_mut().put(&block.height, todo!())?;
```
hinto-janai, line 99: Is this the blob of tx's in the block?
```rust
monero_serai::Block {
    /* ... */
    txs: Vec<[u8; 32]> // <- this?
}
```

database/src/ops/block.rs
```diff
@@ -1,89 +1,321 @@
...
+    let (block_height, block_hash) = if todo!() {
+        let (block_height, block_info) = tables.block_info_v1s_mut().pop_last()?;
+        (block_height, block_info.block_hash)
+    } else if todo!() {
+        let (block_height, block_info) = tables.block_info_v2s_mut().pop_last()?;
+        (block_height, block_info.block_hash)
+    } else {
+        let (block_height, block_info) = tables.block_info_v3s_mut().pop_last()?;
+        (block_height, block_info.block_hash)
+    };
```
hinto-janai, line 183: What behavior to do here?

Also - should this function construct and return a `VerifiedBlockInformation`?

database/src/ops/block.rs
hinto-janai, line 95: What to do with the `u64/u128` size differences between `VerifiedBlockInformation` and our `BlockInfoV*` types?

Casting here is lossy?

database/src/ops/block.rs
```diff
@@ -1,89 +1,321 @@
...
+        let tx_hash: &TxHash = todo!();
+
+        remove_tx(
+            tx_hash,
+            tables.tx_ids_mut(),
+            tables.tx_heights_mut(),
+            tables.tx_unlock_time_mut(),
+            tables.prunable_hashes_mut(),
+            tables.prunable_tx_blobs_mut(),
+        )?;
```
hinto-janai, line 203: When popping blocks, our `BlockInfoV*` types don't have lists of containing tx's, how do we remove them?

Loop over our `TxId => BlockHeight` table and remove until the block height is lower than the current?

database/src/ops/blockchain.rs
```diff
@@ -1,9 +1,41 @@
...
+/// ```
+#[doc = doc_error!()]
+#[inline]
+pub fn height(
+    table_block_heights: &impl DatabaseRo<BlockHeights>,
+) -> Result<BlockHeight, RuntimeError> {
+    // TODO: is this correct?
+    // TODO: is there a faster way to do this? `.len()` is already quite cheap.
+    table_block_heights.len()
 }
```
hinto-janai, line 41: Correct?

database/src/ops/output.rs
```diff
@@ -1,34 +1,192 @@
...
+#[inline]
+pub fn add_output(
+    amount: Amount,
+    output: &Output,
+    table_outputs: &mut impl DatabaseRw<Outputs>,
+    table_num_outputs: &mut impl DatabaseRw<NumOutputs>,
+) -> Result<AmountIndex, RuntimeError> {
+    // Amount index is just the (current_amount_of_outputs + 1).
+    let amount_index = get_num_outputs(table_outputs)?;
+    table_num_outputs.put(&amount, &amount_index)?;
```
hinto-janai, line 48: Correct?

database/src/ops/output.rs
```diff
@@ -1,34 +1,192 @@
...
+    pre_rct_output_id: &PreRctOutputId,
+    table_outputs: &mut impl DatabaseRw<Outputs>,
+    table_num_outputs: &mut impl DatabaseRw<NumOutputs>,
+) -> Result<(), RuntimeError> {
+    // Decrement the amount index by 1, or delete the entry out-right.
+    match table_num_outputs.get(&pre_rct_output_id.amount)? {
+        0 => (),
+        1 => table_num_outputs.delete(&pre_rct_output_id.amount)?,
+        amount_index => table_num_outputs.put(&pre_rct_output_id.amount, &(amount_index - 1))?,
+    }
```
hinto-janai, line 81: When removing outputs, we decrement the `NumOutputs` mapping and remove it completely if 0, correct?

database/src/ops/output.rs
```diff
@@ -1,34 +1,192 @@
...
+/// // TODO
+/// ```
+#[inline]
+pub fn add_rct_output(
+    rct_output: &RctOutput,
+    table_rct_outputs: &mut impl DatabaseRw<RctOutputs>,
+) -> Result<AmountIndex, RuntimeError> {
+    // Amount index is just the (current_amount_of_outputs + 1).
+    let amount_index = get_rct_num_outputs(table_rct_outputs)? + 1;
+    table_rct_outputs.put(&amount_index, rct_output)?;
```
hinto-janai, line 108: Correct?

database/src/ops/property.rs
```diff
@@ -1,9 +1,57 @@
...
+/// ```rust
+/// # use cuprate_database::{*, tables::*, ops::block::*, ops::tx::*};
+/// // TODO
+/// ```
+#[inline]
+pub const fn get_blockchain_pruning_seed() -> Result<PruningSeed, RuntimeError> {
+    // TODO: impl pruning.
+    // We need a DB properties table.
+    Ok(PruningSeed::NotPruned)
+}
```
hinto-janai, line 41: Should pruning be a `TODO` for now? We don't have a properties table yet.

database/src/ops/property.rs
```diff
@@ -1,9 +1,57 @@
...
+/// # Example
+/// ```rust
+/// # use cuprate_database::{*, tables::*, ops::block::*, ops::tx::*};
+/// // TODO
+/// ```
+#[inline]
+pub const fn db_version() -> Result<usize, RuntimeError> {
+    // TODO: We need a DB properties table.
+    Ok(0)
 }
```
hinto-janai, line 57: Same here re: properties table.

database/src/ops/tx.rs
```diff
@@ -1,64 +1,177 @@
...
-pub fn remove_tx_outputs() {
-    todo!()
-}
+    // TODO: What exactly is a `UnlockTime (u64)` in Cuprate's case?
+    // What should we be storing? How?
+    match tx.prefix.timelock {
+        Timelock::None => (),
+        Timelock::Block(height) => todo!(), // Calculate from height?
+        Timelock::Time(time) => todo!(),
+    };
```
hinto-janai, line 63: How are we storing the timelock? This is inside `add_tx()`.

database/src/ops/tx.rs
```diff
@@ -1,64 +1,177 @@
...
+    // Is it `tx.blob: Vec<u8>`?
+    table_prunable_tx_blobs.put(&tx_id, todo!())?;
 
-/// TODO
-pub fn tx_exists() {
-    todo!()
+    // TODO: impl pruning
+    if let PruningSeed::Pruned(decompressed_pruning_seed) = get_blockchain_pruning_seed()? {
+        // TODO: what to store here? which table?
+    }
```
hinto-janai, line 79: Also inside `add_tx()`.

database/src/tables.rs
```diff
@@ -127,6 +326,13 @@ tables! {
     PrunableHashes,
     TxId => PrunableHash,
 
+    // TODO: impl a properties table:
+    // - db version
+    // - pruning seed
+    //
+    // Properties,
+    // KEY => VALUE,
```
hinto-janai, line 334: What would the `(key, value)` be?

hinto-janai (commented):
@Boog900 Many questions.

database/src/ops/block.rs
Boog900: It is the output from this: https://github.com/serai-dex/serai/blob/93be7a30674ecedfb325b6d09dc22d550d7c13f8/coins/monero/src/block.rs#L110 but it can be added to `VerifiedBlockInformation` as we already serialize the block in the consensus code.

database/src/ops/blockchain.rs
Boog900: yeah but this should be called `chain_height` to make it clear that this returns the height of the chain and not the height of the top block

database/src/ops/output.rs
Boog900: yep but the comment is wrong, no plus 1

database/src/ops/output.rs
Boog900: yep

database/src/ops/output.rs
Boog900: nope, no plus 1

database/src/ops/property.rs
Boog900: yeah

database/src/ops/tx.rs
Boog900: just store it as a raw u64, but only store it if it is not `None`

database/src/ops/block.rs
```diff
@@ -1,89 +1,321 @@
...
+                    amount,
+                    &Output {
+                        key: *output.key.as_bytes(),
+                        height: todo!(),
+                        output_flags: todo!(),
+                        tx_idx: todo!(),
+                    },
+                    tables.outputs_mut(),
+                    tables.num_outputs_mut(),
+                )?;
```
Boog900, line 137: All V2 transactions produce RCT outputs not just txs which use RCT.

What this means is that V2 miner transactions are stored in the RCT table, so to decide what table to store the output in we need to look at the tx version

Boog900 (commented):
I think we should change some tables:

- Remove the different versions of Block Info only having v3 which would just be Block Info, we would have to store extra data but the amount of data would be minimal.

- Remove `Prunable Hash` & `Pruned Tx Blobs` & `Prunable Tx Blobs` (for now), monero-serai does not yet support pruning and it isn't needed for now, we can look to add it later these tables shouild be replaced by just a `Tx Blobs` table.

database/src/tables.rs
hinto-janai: We'll have to make a `StorableString` eventually I guess?

database/src/ops/tx.rs
hinto-janai: Are these not 2 different units of time? We don't need to treat them differently?

This is okay?
```rust
match tx.prefix.timelock {
    Timelock::None => (),
    Timelock::Block(height) => tables.tx_unlock_time_mut().put(&tx_id, &(height as u64))?,
    Timelock::Time(time) => tables.tx_unlock_time_mut().put(&tx_id, &time)?,
}
```

database/src/ops/block.rs
hinto-janai: > Remove the different versions of Block Info only having v3 which would just be Block Info

Will we just insert dummy values in Pre-RCT `BlockInfo`'s where needed?

Boog900: > Will we just insert dummy values in Pre-RCT BlockInfo's where needed?

no, `cumulative_rct_outs` will still be valid it will just be 0 and `long_term_weight` should be the value in `VerifiedBlockInformation`

> TxId (u64) => TxBlob (StorableVec) as the (key, value)?

yeah

database/src/ops/tx.rs
Boog900: > Are these not 2 different units of time? We don't need to treat them differently?

height is any value less than 500_000_000 and timestamp is any value above so we will be able to tell them apart by that

database/src/ops/block.rs
```diff
@@ -1,89 +1,327 @@
...
+                Input::Gen(_) => miner_tx = true,
+            }
+        }
+
+        // Output bit flags.
+        // Set to a non-zero bit value if the unlock time is non-zero.
+        let output_flags = match tx.prefix.timelock {
+            Timelock::None => 0x0000_0000,
+            Timelock::Block(_) | Timelock::Time(_) => 0x0000_0001,
+        };
```
hinto-janai, line 123: Is this the correct bit value?

database/src/ops/block.rs
```diff
@@ -1,89 +1,327 @@
...
+                add_rct_output(
+                    &RctOutput {
+                        key,
+                        height,
+                        output_flags,
+                        tx_idx,
+                        commitment: todo!(), // Zero dummy commit?
+                    },
+                    tables.rct_outputs_mut(),
+                )?;
```
hinto-janai, line 141: What is `commitment` in the v2 miner transaction case?

hinto-janai (commented):
@Boog900 More questions.

database/src/ops/block.rs
SyntheticBird45:
```cpp
/* (blockchain_db/blockchain_db.cpp, l.232) */
    // miner v2 txes have their coinbase output in one single out to save space,
    // and we store them as rct outputs with an identity mask
    if (miner_tx && tx.version == 2)
    {
      cryptonote::tx_out vout = tx.vout[i];
      rct::key commitment = rct::zeroCommit(vout.amount);
      vout.amount = 0;
      amount_output_indices[i] = add_output(tx_hash, vout, i, tx.unlock_time,
        &commitment);
    }
```
Zero commitment indeed.

database/src/ops/block.rs
Boog900: Yeah, might be a good idea to use `bitflags`

database/src/ops/block.rs
Boog900: it's:

```math
Commitment = G +  H*ammount
```

Where G is the base point and H is from monero-serai, amount is the output amount.

Monero implements a look up table for common these values: https://github.com/monero-project/monero/blob/c8214782fb2a769c57382a999eaf099691c836e7/src/ringct/rctOps.cpp#L322

For now we can just compute them all on the fly

database/src/ops/block.rs
Boog900: > Inside our cuprate_database::types::Output { .., output_flags: u32 }? Or just as a type for external use? It'll mess with bytemuck if we include the type in the struct I think.

we can do:

```rust
#[derive(Pod, etc)]
#[repr(transparent)]
pub struct OutputFlags(u32;

bitflags! {
    impl OutputFlags: u32 {
        const TIME_LOCKED = 0b0000_0001;
     }
}
```

Which should work right?

> Height's are u64 so add_block() doesn't work unless you wanted to cast since 500_000_000 < u32::MAX?

Yeah just do a cast. Do `try_into` though so we panic if we somehow get a higher height when adding a block (should never happen but better than a silent corrupted DB).

database/src/ops/block.rs
```diff
@@ -1,89 +1,537 @@
...
+/// ```rust
+/// # use cuprate_database::{*, tables::*, ops::block::*};
+/// // TODO
+/// ```
+#[doc = doc_error!()]
+#[inline]
+pub fn get_block(
+    tables: &impl Tables,
+    block_hash: BlockHash,
+) -> Result<VerifiedBlockInformation, RuntimeError> {
```
Boog900, line 287: I don't think this function will be needed FWIW

hinto-janai: I was just about to ping you saying that returning `VerifiedBlockInformation` is impossible within `cuprate_database` unless we pulled in `consensus/` for stuff like `pow_hash` but then realized `Response` only needs `ExtendedBlockHeader` and `monero_serai::Block`...

Should these be the only functions we have for retrieval?
```rust
fn get_block(...) -> Result<Block, RuntimeError>; // for other users
fn get_block_and_tx(...) -> Result<(Block, Vec<Transaction>), RuntimeError>;
fn get_block_extended_header(...) -> Result<ExtendedBlockHeader, RuntimeError>;
```

Boog900: we can probably get away with just `get_block_extended_header` for now. `BlockBatchInRange` was a hack for getting the RPC scanner to work so can be removed.

We will need a request for block and tx blobs but we don't need to do that now.

database/Cargo.toml
```diff
@@ -9,25 +9,30 @@ repository  = "https://github.com/Cuprate/cuprate/tree/main/database"
...
 # default   = ["redb", "service"]
-default   = ["redb-memory", "service"]
+# default   = ["redb-memory", "service"]
 heed       = ["dep:heed"]
 redb       = ["dep:redb"]
 redb-memory = ["redb"]
 service     = ["dep:crossbeam", "dep:futures", "dep:tokio", "dep:tokio-util", "dep:tower", "dep:rayon"]
 
 [dependencies]
+bitflags = { workspace = true, features = ["serde", "bytemuck"] }
```
hinto-janai, line 21: ## `database/Cargo.toml`

Added:
- `bitflags`
- `cuprate_types`
- `curve25519_dalek`
- `monero_pruning`
- `monero_serai`

Added (dev-only):
- `cuprate-test-utils`
- `pretty_assertions`
- `hex`
- `hex-literal`

database/src/backend/heed/env.rs
```diff
@@ -206,24 +206,22 @@ impl Env for ConcreteEnv {
...
         create_table::<NumOutputs>(&env, &mut tx_rw)?;
         create_table::<Outputs>(&env, &mut tx_rw)?;
         create_table::<PrunableHashes>(&env, &mut tx_rw)?;
         create_table::<PrunableTxBlobs>(&env, &mut tx_rw)?;
         create_table::<PrunedTxBlobs>(&env, &mut tx_rw)?;
         create_table::<RctOutputs>(&env, &mut tx_rw)?;
+        create_table::<TxBlobs>(&env, &mut tx_rw)?;
         create_table::<TxHeights>(&env, &mut tx_rw)?;
         create_table::<TxIds>(&env, &mut tx_rw)?;
         create_table::<TxUnlockTime>(&env, &mut tx_rw)?;
```
hinto-janai, line 227: ## Tables

Changed:
- `BlockInfo{V1,V2,V3}` => `BlockInfo`

Added:
- `TxBlobs`

The `Pruned` tables still exist but don't get touched at all in `ops/`.

database/src/database.rs
```diff
@@ -149,6 +149,41 @@ pub trait DatabaseRw<T: Table>: DatabaseRo<T> {
...
+        match f(value) {
+            Some(value) => DatabaseRw::put(self, key, &value)?,
+            None => match DatabaseRw::delete(self, key) {
+                Ok(()) | Err(RuntimeError::KeyNotFound) => (),
+                Err(e) => return Err(e),
+            },
+        }
+
+        Ok(())
+    }
```
hinto-janai, line 185: ## `trait DatabaseRw`

Added functions:
- `take()`
- `update()`

`take()` is used a lot when deleting/popping data and we need access to the value after, it's pretty much just `get()` + `delete()`.

`update()` is used in `remove_output()` where we sometimes:
- update the value without removal or...
- ...remove the `(key, value)` entry completely

Both seem useful enough to have in the API.

database/src/backend/heed/database.rs
```diff
@@ -200,59 +200,63 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
...
-        // We are deleting the value and never accessing
-        // the iterator again so this should be safe.
-        unsafe {
-            iter.del_current()?;
+        match self.db.delete(tx_rw, &key) {
+            Ok(true) => Ok((key, value)),
+            Err(e) => Err(e.into()),
+            // We just `get()`'ed the value - it is
+            // incorrect for it to suddenly not exist.
+            Ok(false) => unreachable!(),
```
hinto-janai, line 239: ## `backend/heed`

Updated function bodies:
- `pop_first()`
- `pop_last()`

These now just use `heed::Database::first()/last()` + `delete()` instead of `unsafe` cursors.

The old comment doesn't matter since we don't care about duplicates anymore.

database/src/tables.rs
```diff
@@ -4,26 +4,271 @@
...
 }
 
+//---------------------------------------------------------------------------------------------------- `trait Tables[Mut]`
+/// Creates:
+/// - `pub trait Tables`
+/// - `pub trait TablesMut`
+/// - Blanket implementation for `(tuples, containing, all, open, database, tables, ...)`
+///
+/// For why this exists, see: <https://github.com/Cuprate/cuprate/pull/102#pullrequestreview-1978348871>.
+macro_rules! define_trait_tables {
```
hinto-janai, line 32: ## `EnvInner` / `trait Tables{Mut,Iter}`

- Added `trait Tables{Mut,Iter}` as discussed before
- Added `fn open_tables()`, `fn open_tables_mut()` to return those `impl Tables{Mut,Iter}`

`trait TablesIter` was added so we can do `trait TablesMut: Tables` so
our `impl TablesMut` objects have access to multiple table reads too.

database/src/tests.rs
```diff
@@ -0,0 +1,94 @@
+//! Utilities for `cuprate_database` testing.
+//!
+//! These types/fn's are only:
+//! - enabled on #[cfg(test)]
+//! - only used internally
```
hinto-janai, line 5: ## Tests

- Updated a bunch of tests
- Added top-level `tests.rs` file for testing util functions

database/src/types.rs
```diff
@@ -266,6 +175,38 @@ pub struct BlockInfoV3 {
...
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    #[derive(Copy, Clone, Debug, PartialEq, PartialOrd, Eq, Ord, Hash, Pod, Zeroable)]
+    #[repr(transparent)]
+    pub struct OutputFlags: u32 {
+        /// No flags set.
+        const NONE = 0b0000_0000;
+        /// This output has a non-zero unlock time.
+        const NON_ZERO_UNLOCK_TIME = 0b0000_0001;
+    }
+}
```
hinto-janai, line 208: ## Types

- Added `OutputFlags` with `bitflags!()` to replace our `u32` in `struct Output`
- Added `block_blob` in `struct VerifiedBlockInformation`

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
+pub fn add_block(
+    block: &VerifiedBlockInformation,
+    tables: &mut impl TablesMut,
+) -> Result<(), RuntimeError> {
+    // let new_block_height = chain_height(tables.block_heights())?;
+    // if block.height > new_block_height {
+    //     // ...provided a block that isn't (chain tip + 1)
+    // } else if block.height < new_block_height {
+    //     // ... provided a block that already exists
+    // }
```
hinto-janai, line 61: Should `ops/` functions react to user error?

Some conditions that aren't handled:
- Input block height > chain tip
- Input `{block,tx,output,etc}` already exists

What should happen here? Should we add/return a `RuntimeError::AlreadyExists` error?

Should `add_block()` and others just trust the caller and assume they satisfied documented preconditions?

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
+                    // TODO: all test tx's have the same key image so the 1st
+                    // removal removes everything - fix me after we have real data.
+                    if cfg!(test) {
+                        match result {
+                            Ok(()) | Err(RuntimeError::KeyNotFound) => (),
+                            Err(e) => return Err(e),
+                        }
+                    } else {
+                        result?;
+                    }
```
hinto-janai, line 236: Will be fixed after we have real test data - the current ones all have the same key image.

hinto-janai (commented):
@Boog900 Ready for review.

This PR ended up touching a lot more than just `ops/` so I marked the changes below.

As for `ops/`, it ended up being many fine-grained functions + big `block` functions that use them all. All fine-grained functions (`add_tx()`, `add_output()`, etc) have docs on them explaining this.

Re: https://github.com/Cuprate/cuprate/pull/108 I still don't have access to real `VerifiedBlockInformation` data so all the current tests for `ops/` uses semi-fake data. I think this is okay enough for now since real `cuprate_database` usage will show obvious mistakes + by that time we'll have real data to test on?

In terms of logic and I/O, `ops/` seems to work fine, data gets written to the correct places, it can be read back, etc.

database/src/types.rs
Boog900: `NONE` doesn't need to exist, we can just use `empty()`. The docs also say to avoid zero bit flags: https://docs.rs/bitflags/latest/bitflags/#zero-bit-flags

database/src/ops/block.rs
Boog900: We could add asserts but I don't think these should be runtime errors

Boog900 (changes requested):

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
+        for (i, output) in tx.prefix.outputs.into_iter().enumerate() {
+            // Outputs with clear amounts.
+            if let Some(amount) = output.amount {
+                // RingCT miner outputs.
+                if miner_tx && tx.prefix.version == 2 {
+                    let amount_index = get_rct_num_outputs(tables.rct_outputs())?.saturating_sub(1);
+                    remove_rct_output(&amount_index, tables.rct_outputs_mut())?;
+                // Pre-RingCT outputs.
+                } else {
+                    let amount_index = tables.num_outputs_mut().take(&amount)?;
```
Boog900, line 253: We need a minus 1 here right?

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
+            }
+        }
+
+        // Remove each output in the transaction.
+        for (i, output) in tx.prefix.outputs.into_iter().enumerate() {
+            // Outputs with clear amounts.
+            if let Some(amount) = output.amount {
+                // RingCT miner outputs.
+                if miner_tx && tx.prefix.version == 2 {
+                    let amount_index = get_rct_num_outputs(tables.rct_outputs())?.saturating_sub(1);
```
Boog900, line 249: We don't need `saturating_sub` here right? this should never be 0

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
+                        &PreRctOutputId {
+                            amount,
+                            amount_index,
+                        },
+                        tables,
+                    )?;
+                }
+            // RingCT outputs.
+            } else {
+                let amount_index = get_rct_num_outputs(tables.rct_outputs())?.saturating_sub(1);
```
Boog900, line 264: and here

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
+        let output_flags = match tx.prefix.timelock {
+            Timelock::None => OutputFlags::NONE,
+            Timelock::Block(_) | Timelock::Time(_) => OutputFlags::NON_ZERO_UNLOCK_TIME,
+        };
 
-/// TODO
-pub fn get_block_timestamp() {
-    todo!()
+        // Output data.
+        for (i, output) in tx.prefix.outputs.iter().enumerate() {
```
Boog900, line 131: it doesn't look like the [`tx outputs`](https://github.com/Cuprate/cuprate/tree/db-schema/database#tx-outputs) table is being filled?

Boog900 (changes requested):

database/src/backend/heed/database.rs
```diff
@@ -200,59 +200,63 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
 
     #[inline]
     fn delete(&mut self, key: &T::Key) -> Result<(), RuntimeError> {
-        self.db.delete(&mut self.tx_rw.borrow_mut(), key)?;
-        Ok(())
+        match self.db.delete(&mut self.tx_rw.borrow_mut(), key) {
+            Ok(true) => Ok(()),
+            Ok(false) => Err(RuntimeError::KeyNotFound),
+            Err(e) => Err(e.into()),
+        }
```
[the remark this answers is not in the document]
SyntheticBird45: I don't see how we could fall in that case. I agree the outcome is the same, but its worth logging it, as it will be an indicator that somehow we tried to.
```rust
if !self.db.delete(&mut self.tx_rw.borrow_mut(), key)? {
    warn!("Tried to delete non existent key") // replace with println
}
```

database/src/database.rs
```diff
@@ -149,6 +149,41 @@ pub trait DatabaseRw<T: Table>: DatabaseRo<T> {
...
+    where
+        F: FnMut(T::Value) -> Option<T::Value>,
+    {
+        let value = DatabaseRo::get(self, key)?;
+
+        match f(value) {
+            Some(value) => DatabaseRw::put(self, key, &value)?,
+            None => match DatabaseRw::delete(self, key) {
+                Ok(()) | Err(RuntimeError::KeyNotFound) => (),
+                Err(e) => return Err(e),
```
SyntheticBird45, line 180: Took 30 sec to wonder what would be the use case for `None` resulting items to be deleted. Can you add to the doc a little mention that it can be used for conditional deletion ?

SyntheticBird45: And I also don't see how we can fall into `RuntimeError::KeyNotFound` since we just `get`d it earlier.

database/src/ops/block.rs
SyntheticBird45: We should trust the caller. We're most likely to have handled these input parameters higher in the code. therefore, checks should be done higher.

database/src/ops/block.rs
SyntheticBird45: Just took a look at tables.rs. There are no defined tables that link TxID to amount idxs...
Documentation error ? or table mistake ?

database/src/ops/blockchain.rs
```diff
@@ -1,9 +1,124 @@
...
+/// - The blockchain has 0 blocks => this returns `Err(RuntimeError::KeyNotFound)`
+/// - The blockchain has 1 block (height 0) => this returns `Ok(0)`
+/// - The blockchain has 2 blocks (height 1) => this returns `Ok(1)`
+#[doc = doc_error!()]
+#[inline]
+pub fn top_block_height(
+    table_block_heights: &impl DatabaseRo<BlockHeights>,
+) -> Result<BlockHeight, RuntimeError> {
+    match table_block_heights.len()? {
+        0 => Err(RuntimeError::KeyNotFound),
```
SyntheticBird45, line 55: At which point could the blockchain has 0 blocks ?

SyntheticBird45:
Oops. @Boog900 some of my review comments were supposed to answer some of your comments.

database/src/ops/block.rs
hinto-janai: Okay, adding asserts and panic docs for:
- `block.height > u32::MAX`
- `block.height` is not != `chain_height`

As for data that already exists, the function will just overwrite previous data.

database/src/backend/heed/database.rs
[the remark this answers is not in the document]
hinto-janai: My bad actually, this would have caused behavior differences between our current `redb` impl.

Changed back to returning `Ok(())` if key doesn't exist.

database/src/ops/block.rs
```diff
@@ -1,89 +1,492 @@
...
-}
+    // Block Info.
+    tables.block_infos_mut().put(
+        &block.height,
+        &BlockInfo {
+            timestamp: block.block.header.timestamp,
+            total_generated_coins: block.generated_coins,
+            cumulative_difficulty: block.cumulative_difficulty,
+            block_hash: block.block_hash,
+            cumulative_rct_outs,
```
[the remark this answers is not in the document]
hinto-janai: Oooo yeah.

Block operations are now _after_ transaction/outputs.

database/src/ops/output.rs
```diff
@@ -1,34 +1,259 @@
...
+#[doc = doc_add_block_inner_invariant!()]
+#[doc = doc_error!()]
+#[inline]
+pub fn add_output(
+    amount: Amount,
+    output: &Output,
+    tables: &mut impl TablesMut,
+) -> Result<PreRctOutputId, RuntimeError> {
+    let amount_index = get_num_outputs(tables.outputs_mut())?;
+    tables.num_outputs_mut().put(&amount, &amount_index)?;
```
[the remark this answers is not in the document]
hinto-janai: Is it not `.len() == new_index`?

```
1st output == [0]
2nd output == [1]
[...]
```

so `outputs.len() == 1` means there's 1 output with index `[0]`, so the next is 1?

database/src/ops/block.rs
hinto-janai: Same question as https://github.com/Cuprate/cuprate/pull/102#discussion_r1568076162

database/src/ops/output.rs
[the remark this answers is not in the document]
Boog900: previous len == new index.

This line is setting the new len of outputs with that amount to `new index`.

I have also juts realized the line before which gets the amount index also should be taking into account the output amount as the index is per amount

database/src/ops/block.rs
Boog900: This is removing outputs so the index of the top output is one less than the amount of outputs

database/src/ops/block.rs
hinto-janai: Oops, that table wasn't even implemented... added table and updated `add_block()`

database/src/database.rs
hinto-janai: > Can you add to the doc a little mention that it can be used for conditional deletion ?

Added. I think this function could be replaced by some `std::collections::btree_map::Entry`-like API but that can be for later.

>  And I also don't see how we can fall into RuntimeError::KeyNotFound since we just getd it earlier.

True, removed.

database/src/ops/block.rs
hinto-janai: Sent message before saw yours @SyntheticBird45 :)

I'm ok with either but I think asserts further up the stack will just be optimized away if they happened in a previous function so more asserts is never bad?

(also I already wrote the tests...)

Boog900: the asserts may mess with testing btw

hinto-janai: Which testing? Should I remove them for now?

SyntheticBird45: > I think asserts further up the stack will just be optimized away if they happened in a previous function so more asserts is never bad?

You're right. asserts help rustc understand the bounds. I would say to got for it. Tho I don't see what @Boog900 is referring to

database/src/ops/output.rs
```diff
@@ -1,34 +1,259 @@
...
+#[inline]
+pub fn remove_output(
+    pre_rct_output_id: &PreRctOutputId,
+    tables: &mut impl TablesMut,
+) -> Result<(), RuntimeError> {
+    // Decrement the amount index by 1, or delete the entry out-right.
+    tables
+        .num_outputs_mut()
+        .update(&pre_rct_output_id.amount, |amount| {
+            if amount == 0 {
```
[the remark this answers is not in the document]
hinto-janai: Oops, this is the `amount_index` not `amount`, we should be removing if `0` since it starts at `0` right?

```rust
if amount_index == 0 {
    None
} else {
    Some(amount_index - 1)
}
```

database/src/ops/block.rs
hinto-janai: Does the 1st output not have index `0`?

In that case `get_rct_num_outputs` will return `0`, so we want to `saturating_sub`?

database/src/ops/blockchain.rs
hinto-janai: Before initial sync? Unless we are supposed to special case the first block manually?

database/src/ops/block.rs
Boog900: If you have a test adding blocks to the DB with the assert that block height == chain height all test blocks will need to have their height as 0/ all blocks before it would need to be added. Not a big deal but still

Boog900:
@hinto-janai 

The table `num_outputs` should be storing the number of outputs with a certain amount not the top amount index for that amount.

And when getting the new index of an output you would need to do a lookup on that table with the outputs amount, currently you are just getting the amount of entries in the table.

database/src/ops/block.rs
Boog900: Yes the first index will be 0, no the amount of RCT outputs will not be zero if we have one RCT output.

database/src/ops/output.rs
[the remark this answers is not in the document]
hinto-janai: > previous len == new index.

If `outputs.len() == 0` then the index for a newly added output is `0`, no?

> I have also juts realized the line before which gets the amount index also should be taking into account the output amount as the index is per amount

Oops, yeah fixed.

database/src/ops/blockchain.rs
SyntheticBird45: > Unless we are supposed to special case the first block manually?

That's what I'm wondering. Genesis block is a constant. We're still supposed to receive it from network iirc, it should be accepted without questioning, so impossible we hit 0. Maybe I'm wrong on my thinking. @Boog900 ?

database/src/ops/block.rs
hinto-janai: Ah, right. `block.height` is currently mutated in tests to avoid this.

I'll see if they have to be removed when we have real data.

database/src/ops/block.rs
hinto-janai: Nvm I was confused here.

hinto-janai:
@Boog900 fixes made. Resolve previous conversations if OK.

database/src/ops/block.rs
```diff
@@ -1,89 +1,555 @@
...
+///
+/// Note: This is more expensive than the above.
+#[doc = doc_error!()]
+#[inline]
+pub fn get_block_extended_header_from_height(
+    block_height: &BlockHeight,
+    tables: &impl Tables,
+) -> Result<ExtendedBlockHeader, RuntimeError> {
+    get_block_extended_header(&tables.block_infos().get(block_height)?.block_hash, tables)
 }
```
Boog900, line 335: These functions should be the other way around IMO `get_block_extended_header` should call `get_block_extended_header_from_height` after getting the blocks height from hash

database/src/ops/block.rs
Boog900: @hinto-janai This should be get not take right? and we still need the minus 1

database/src/ops/tx.rs
```diff
@@ -1,64 +1,296 @@
...
+/// Note that the caller's input is trusted implicitly and no checks
+/// are done (in this function) whether the `block_height` is correct or not.
+///
+/// # Notes
+/// This uses the [`chain_height`] when addings to the [`TxHeights`] table.
+///
+#[doc = doc_add_block_inner_invariant!()]
+#[doc = doc_error!()]
+#[inline]
+pub fn add_tx(
```
Boog900, line 47: I think I would prefer if output data was added in this function, it feels more at home here than `add_block`

database/src/ops/tx.rs
```diff
@@ -1,64 +1,296 @@
...
-}
+    // Transaction data.
+    tables.tx_ids_mut().put(&tx.tx_hash, &tx_id)?;
+    tables.tx_heights_mut().put(&tx_id, block_height)?;
+    tables
+        .tx_blobs_mut()
+        .put(&tx_id, StorableVec::wrap_ref(&tx.tx_blob))?;
+    tables.tx_outputs_mut().put(&tx_id, amount_indices)?;
+
+    // Key images.
```
Boog900, line 63: Misplaced? although it can stay if adding this data gets moved to this function :)

Boog900 (changes requested):

database/src/ops/tx.rs
Boog900: (the key image comment)

database/src/ops/blockchain.rs
Boog900: Another part of Cuprate will handle the genesis block, it is hard coded but different for each network

SyntheticBird45: Alright so impossible to get 0. You can make it `unreachable!()`

hinto-janai: Although, is there a reason why we should panic? Returning `Result` where possible over panic is preferred IMO - the function already returns `Result`.

I don't think it'll do anything for optimizations since `0` is always branched for:
```rust
match table_block_heights.len()? {
    0 => panic!(), // branch
    height => Ok(height - 1),
}

// or...
//          this operator branches and panics too
//                         v
table_block_heights.len()? - 1
```
Thinking about non-`cuprated` use-cases too.

database/src/ops/tx.rs
hinto-janai: Sure, but I'll note that this breaks the logical `data type <-> fn` groupings, i.e:
- `Transaction` <-> `{get,remove}_tx()`
- `Output` <-> `{get,remove}_output()`

As of now each sub-function only mutates its related tables. This makes `add_tx()` have a side effect that it also touches outputs.

database/src/ops/blockchain.rs
SyntheticBird45: > Although, is there a reason why we should panic?

Just to avoid the overhead of handling that case higher in block downloader

>  Returning Result where possible over panic is preferred IMO. Thinking about non-cuprated use-cases too.

Fair enough. The error is a little unclear so maybe add a mention in the doc "If no blocks registered yet, then this function return KeyNotFound"

database/src/ops/block.rs
```diff
@@ -1,89 +1,413 @@
...
+/// - `block.height > u32::MAX` (not normally possible)
+/// - `block.height` is not != [`chain_height`]
+///
+/// # Already exists
+/// This function will operate normally even if `block` already
+/// exists, i.e., this function will not return `Err` even if you
+/// call this function infinitely with the same block.
+// no inline, too big.
+#[allow(clippy::too_many_lines)]
+#[allow(clippy::manual_assert)] // assert doesn't let you `{}`
```
Boog900, line 60: I am not sure what this comment means, if it means you can do formatting with assert, I am pretty sure you can: https://doc.rust-lang.org/std/macro.assert.html#custom-messages

Boog900 (commented):
Just one small thing.

database/src/ops/block.rs
```diff
@@ -1,89 +1,411 @@
...
+    // INVARIANT: must be below the above transaction loop since this
+    // RCT output count needs account for _this_ block's outputs.
+    let cumulative_rct_outs = get_rct_num_outputs(tables.rct_outputs())?;
+
+    // Block Info.
+    tables.block_infos_mut().put(
+        &block.height,
+        &BlockInfo {
+            timestamp: block.block.header.timestamp,
+            total_generated_coins: block.generated_coins,
```
Boog900, line 95: `total_generated_coins` is a cumulative value btw, so this should be adding the previous blocks total_generated_coins

Boog900: could maybe rename to make it more clear? up to you

database/src/ops/block.rs
```diff
@@ -1,89 +1,454 @@
...
+        assert_eq!(block.block.hash(), block.block_hash);
+        assert_eq!(block.block.txs.len(), block.txs.len());
+        assert_ne!(block.generated_coins, 0);
+        for (i, tx) in block.txs.iter().enumerate() {
+            assert_eq!(tx.tx_blob, tx.tx.serialize());
+            assert_eq!(tx.tx_weight, tx.tx.weight());
+            assert_eq!(tx.tx_hash, tx.tx.hash());
+            assert_eq!(tx.tx_hash, block.block.txs[i]);
+        }
+    }
```
hinto-janai, line 92: Could do `#[cfg(test)]` instead if you don't want to slow down debug builds too much.

database/src/ops/block.rs
```diff
@@ -1,89 +1,454 @@
...
-}
+    //------------------------------------------------------ Block Info
+
+    // INVARIANT: must be below the above transaction loop since this
+    // RCT output count needs account for _this_ block's outputs.
+    let cumulative_rct_outs = get_rct_num_outputs(tables.rct_outputs())?;
+
+    let cumulative_generated_coins =
+        cumulative_generated_coins(&block.height.saturating_sub(1), tables.block_infos())?
+            + block.generated_coins;
```
hinto-janai, line 107: New function in `cuprate_database::ops::block` to return cumulative generated coins, defined...

database/src/ops/blockchain.rs
```diff
@@ -1,9 +1,156 @@
...
+pub fn cumulative_generated_coins(
+    block_height: &BlockHeight,
+    table_block_infos: &impl DatabaseRo<BlockInfos>,
+) -> Result<u64, RuntimeError> {
+    match table_block_infos.get(block_height) {
+        Ok(block_info) => Ok(block_info.cumulative_generated_coins),
+        Err(RuntimeError::KeyNotFound) if block_height == &0 => Ok(0),
+        Err(e) => Err(e),
+    }
+}
```
hinto-janai, line 85: ...here.

database/src/ops/block.rs
```diff
@@ -1,89 +1,454 @@
...
+            block_v1_tx2().clone(),
+            block_v9_tx3().clone(),
+            block_v16_tx0().clone(),
+        ];
+        // HACK: `add_block()` asserts blocks with non-sequential heights
+        // cannot be added, to get around this, manually edit the block height.
+        for (height, block) in blocks.iter_mut().enumerate() {
+            block.height = height as u64;
+            assert_eq!(block.block.serialize(), block.block_blob);
+        }
```
hinto-janai, line 286: `ops/` tests use real data from `test-utils/` now.

But, we modify the height so `add_block()` doesn't panic - kinda hacky. Could remove the asserts if you'd prefer.

database/src/ops/block.rs
Boog900: I would remove the hashes and weight as they both require serialization as well and if they are wrong we are doing something horribly wrong elsewhere also are you sure the reward can never be 0? even with penalties? 

```suggestion
    // Expensive checks - debug only.
    #[cfg(debug_assertions)]
    {
        assert_eq!(block.block.serialize(), block.block_blob);
        assert_eq!(block.block.txs.len(), block.txs.len());
        assert_ne!(block.generated_coins, 0);
        for (i, tx) in block.txs.iter().enumerate() {
            assert_eq!(tx.tx_blob, tx.tx.serialize());
            assert_eq!(tx.tx_hash, block.block.txs[i]);
        }
    }
```

database/src/ops/blockchain.rs
Boog900: I don't think we need the if block_height here

database/src/ops/block.rs
Boog900: I mean the data added isn't 100% accurate anyway - the indexs for outputs will be incorrect so should be fine

Boog900 (changes requested):
see comments above

database/src/ops/blockchain.rs
hinto-janai: We need to special case `before block 0` but make sure this returns `Err` if the `block_height` doesn't exist, right?

If this returned `Ok(0)` on a height that didn't exist (other than before 0), the above call in `add_block()` would be proceeding with a bad value.

Boog900 (approved):

PR #104, opened 2024-04-03
database: make `open_db_rw()` take `&TxRw`

hinto-janai:
## Part 21
https://github.com/Cuprate/cuprate/pull/102 <- Previous

Fixes https://github.com/Cuprate/cuprate/pull/102#discussion_r1548695610.

- Makes `EnvInner::open_db_rw()` take a `&` instead of `&mut` write transaction
- Makes `heed` compatible by using `RefCell`
- Moves iteration functions from `trait DatabaseRo` -> `trait DatabaseIter` that are no longer accessible from `trait DatabaseRw` objects

### DIFF
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -3,7 +3,7 @@
 //---------------------------------------------------------------------------------------------------- Import
 use std::{
     borrow::{Borrow, Cow},
-    cell::{Ref, RefCell},
+    cell::RefCell,
     fmt::Debug,
     ops::RangeBounds,
     sync::RwLockReadGuard,
@@ -219,7 +219,6 @@ impl<T: Table> DatabaseRw<T> for HeedTableRw<'_, '_, T> {
         // remove the _first_ and only the first `(key, value)`.
         // `delete()` removes all keys including duplicates which
         // is slightly different behavior.
-        // SAFETY: we have `&mut self`.
         let mut iter = self.db.iter_mut(tx_rw)?;
 
         // SAFETY:
--- a/database/src/backend/heed/database.rs
+++ b/database/src/backend/heed/database.rs
@@ -3,7 +3,7 @@
 //---------------------------------------------------------------------------------------------------- Import
 use std::{
     borrow::{Borrow, Cow},
-    cell::{Ref, UnsafeCell},
+    cell::{Ref, RefCell},
     fmt::Debug,
     ops::RangeBounds,
     sync::RwLockReadGuard,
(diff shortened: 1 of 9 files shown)


hinto-janai:
@Boog900 Re: https://github.com/Cuprate/cuprate/pull/104

If we were to re-impl all functions on `trait DatabaseRw` itself, we could make it take a `&mut self` instead, this would mean we have mutual exclusive access to the TX, so `iter()` and such would be safe?

If not - are these the functions that should be removed?
- `get_range()`
- `iter()`
- `keys()`
- `values()`

Boog900:
> Are these the functions that should be removed?

Yep every function that holds has a lifetime on the return 

> Makes heed compatible by using UnsafeCell

FWIW I would use a RefCell instead of a UnsafeCell as we no longer need to return a reference to the guard

hinto-janai:
How do you feel about moving the iteration functions to `trait DatabaseIter` and only implementing that on our read tables?

It would allow keeping `trait DatabaseRw: DatabaseRo` too.

Boog900:
Yeah that would work if you would prefer to do that

database/src/backend/heed/database.rs
```diff
@@ -204,79 +168,48 @@ impl<T: Table> DatabaseRo<T> for HeedTableRo<'_, T> {
...
 
     #[inline]
     fn delete(&mut self, key: &T::Key) -> Result<(), RuntimeError> {
-        self.db.delete(self.tx_rw, key)?;
+        self.db.delete(&mut self.tx_rw.borrow_mut(), key)?;
         Ok(())
     }
 
     #[inline]
     fn pop_first(&mut self) -> Result<(T::Key, T::Value), RuntimeError> {
```
hinto-janai, line 208: Just for confirmation, `DatabaseRw::pop_{first,last}()` are still safe, correct?

We have `&mut heed::RwTxn` and aren't returning any value with a lifetime.

Boog900: yeah they are ok

database/src/backend/heed/env.rs
```diff
@@ -96,7 +97,23 @@ impl Env for ConcreteEnv {
...
+
+    /// HACK:
+    /// `heed::RwTxn` is wrapped in `RefCell` to allow:
+    /// - opening a database with only a `&` to it
+    /// - allowing 1 write tx to open multiple tables
+    ///
+    /// Our mutable accesses are safe and will not panic as:
+    /// - Write transactions are `!Sync`
+    /// - A table operation does not hold a reference to the inner cell
+    ///   once the call is over, unless the return also holds a lifetime
```
Boog900, line 109: The unless needs to be removed here as it is not safe and is no longer what happens 

```suggestion
    /// - A table operation does not hold a reference to the inner cell
    ///   once the call is over
```

Boog900 (changes requested):

Boog900 (approved):

PR #108, opened 2024-04-15
test-utils: return `&'static` verified superset types

hinto-janai:
Changes the free functions in `cuprate_test_utils::data` to return `&'static` and the superset types `VerifiedBlockInformation, TransactionVerificationData` instead of `Block, Transaction`.

Needed in https://github.com/Cuprate/cuprate/pull/102 tests since `add_block()` needs a `VerifiedBlockInformation` (using a dummy `VerifiedBlockInformation` isn't ideal since I'd also like to test if all the inner data were added to the right tables).

The new `VerifiedBlockMap` and `VerifiedTxMap` just map `Block, Transaction` into the verified forms with some help from static input.

This is sorta hacky since:
1. Static input isn't verified
2. It's only semi-automatic (ideally we just provide a `block.bin` file and all fields get computed/verified)

But doing this involves pulling in `consensus/` and doing PoW hash calculation so... doing it manually (for now).


hinto-janai:
@Boog900 This doesn't actually yet work since we can't easily get a `Transaction` from the `Block`'s TX hashes - solution is manually inputting each TX blob as hex but one of the blocks has 513...

What do you think we should do? We need real `VerifiedBlockInformation` for database testing, ideally a variety of them.

Boog900:
Hmm, I think no matter what we are going to need to input these values manually, it would be too hard to calculate them.

For example to calculate the long term weight of a block you need the last 100,000 blocks long term weights.

Probably best to write a little script to generate the data, using a synced node.

test-utils/src/data/constants.rs
```diff
@@ -3,147 +3,223 @@
...
+        #[doc = concat!("assert_eq!(block.header.minor_version, ", $minor_version, ");")]
+        #[doc = concat!("assert_eq!(block.header.timestamp, ", $timestamp, ");")]
+        #[doc = concat!("assert_eq!(block.header.nonce, ", $nonce, ");")]
+        #[doc = concat!("assert!(matches!(block.miner_tx.prefix.inputs[0], Input::Gen(", $miner_tx_generated, ")));")]
+        #[doc = concat!("assert_eq!(block.txs.len(), ", $tx_len, ");")]
+        #[doc = concat!("assert_eq!(hex::encode(block.hash()), \"", $hash, "\")")]
+        /// ```
+        pub const $name: &[u8] = include_bytes!($data_path);
+    };
+}
```
hinto-janai, line 48: Generated output:

image

test-utils/src/data/constants.rs
```diff
@@ -3,147 +3,223 @@
...
+        #[doc = concat!("assert_eq!(tx.prefix.version, ", $version, ");")]
+        #[doc = concat!("assert_eq!(tx.prefix.timelock, ", stringify!($timelock), ");")]
+        #[doc = concat!("assert_eq!(tx.prefix.inputs.len(), ", $input_len, ");")]
+        #[doc = concat!("assert_eq!(tx.prefix.outputs.len(), ", $output_len, ");")]
+        #[doc = concat!("assert_eq!(tx.signatures.len(), ", $signatures_len, ");")]
+        #[doc = concat!("assert_eq!(hex::encode(tx.hash()), \"", $hash, "\")")]
+        /// ```
+        pub const $name: &[u8] = include_bytes!($data_path);
+    };
+}
```
hinto-janai, line 134: Generated output:

image

test-utils/src/data/free.rs
```diff
@@ -6,105 +6,260 @@
...
+                    weight: $weight,
+                    long_term_weight: $long_term_weight,
+                    cumulative_difficulty: $cumulative_difficulty,
+                    txs: vec![$($tx_blob),*],
+                }
+                .into_verified()
+            })
+        }
+    };
+}
```
hinto-janai, line 174: Generated output:

image

test-utils/src/data/free.rs
```diff
@@ -6,105 +6,260 @@
...
+        #[doc = concat!("assert_eq!(tx.tx_weight, ", $weight, ");")]
+        #[doc = concat!("assert_eq!(tx.tx_hash, hex!(\"", $hash, "\"));")]
+        #[doc = "assert_eq!(tx.fee, tx.tx.rct_signatures.base.fee);"]
+        /// ```
+        pub fn $fn_name() -> &'static TransactionVerificationData {
+            static TX: OnceLock<TransactionVerificationData> = OnceLock::new();
+            TX.get_or_init(|| to_tx_verification_data($tx_blob))
+        }
+    };
 }
```
hinto-janai, line 244: Generated output:

image

test-utils/src/data/free.rs
```diff
@@ -6,105 +6,260 @@
...
+    block_blob: BLOCK_5ECB7E,
+    tx_blobs: [TX_2180A8, TX_D7FEBD],
+    pow_hash: "84f64766475d51837ac9efbef1926486e58563c95a19fef4aec3254f03000000",
+    height: 202_612,
+    generated_coins: 13_138_270_468_431,
+    weight: 55_503,
+    long_term_weight: 55_503,
+    cumulative_difficulty: 126_654_460_829_362,
+    tx_len: 2,
+}
```
hinto-janai, line 187: Less-than-ideal part for defining blocks is that we have to provide each TX blob, so I swapped out the previous one that had 513.

hinto-janai (commented):
@Boog900 @SyntheticBird45 Ready.

Turning raw block/tx blobs into `cuprate_types` types still isn't automatic but https://github.com/Cuprate/cuprate/pull/110 + the macros in this PR make it semi-automatic, we just need to provide some fields.

Some things are verified as well (hashes, tx length, etc).

hinto-janai:
After this and https://github.com/Cuprate/cuprate/pull/110 are merged, I'll create tests in this module that asserts the data constructed here and data from RPC calls match.

Boog900 (approved):

PR #109, opened 2024-04-16
P2P: async buffer

Boog900:
This adds an async buffer that works as a queue which supports arbitrary weights for values.

This is used between the block downloader and Cuprate.


Boog900:
ready @hinto-janai @SyntheticBird45

SyntheticBird45 (approved):
I may be a bad reviewer on this one since I'm not comfortable with async implementations, but I didn't found anything strange.

hinto-janai:
Will review tomorrow but some questions:
- Could this be in `cuprate_helper::asynch`?
- What does `weight` mean in this context? What does it do?
- Is there no other data structure like this? Do we have to make/maintain one?

Boog900:
> Could this be in cuprate_helper::asynch

I don't think so, this is more than just a helper function/ type 

> What does weight mean in this context? What does it do?

It's used to bound the channel, you specify a max weight and a weight for each item. For the block downloader the weight will be the size, in bytes, of the downloaded batch of blocks.

> Is there no other data structure like this? Do we have to make/maintain one?

I couldn't find one, I don't think this is too much to maintain.

p2p/async-buffer/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Attempts to add an item to the buffer returning an error if there is not enough capacity or
+    /// the [`BufferStream`] was dropped.
+    pub fn try_send(&mut self, item: T, size_needed: usize) -> Result<(), BufferError> {
+        let size_needed = min(self.max_item_weight, size_needed);
+
+        if self.capacity.load(Ordering::Acquire) < size_needed {
+            return Err(BufferError::NotEnoughCapacity);
+        }
+
+        let prev_size = self.capacity.fetch_sub(size_needed, Ordering::AcqRel);
```
hinto-janai, line 122: This is the [ABA problem](https://en.wikipedia.org/wiki/ABA_problem) - 2 threads can enter after 118 and wrap the `fetch_sub`.

You need a `compare_exchange` loop for this to be correct.

p2p/async-buffer/tests/basic.rs
```diff
@@ -0,0 +1,37 @@
+use futures::{FutureExt, StreamExt};
+
+use async_buffer::new_buffer;
+
+#[tokio::test]
+async fn async_buffer_send_rec() {
```
hinto-janai, line 6: IMO `loom` tests should always be present when introducing hand-written data structures involving async and/or atomics.

Could be left as a TODO for a big "introduce loom" PR.

p2p/async-buffer/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.project();
+
+        let size_needed = this.ready.size_needed;
+
+        this.ready.as_mut().poll(cx).map(|_| {
+            this.ready
+                .sink
+                .try_send(this.item.take().unwrap(), size_needed)
+        })
```
hinto-janai, line 165: Can you explain in `.expect()` or a comment why `item` is `Option<T>` + `.take()`n?

p2p/async-buffer/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        // set the waker
+        self.sink.sink_waker.register(cx.waker());
+
+        // check the capacity again to avoid a race condition that would result in lost notifications.
+        if self.sink.capacity.load(Ordering::Acquire) >= self.size_needed {
+            Poll::Ready(())
+        } else {
+            Poll::Pending
+        }
+    }
```
hinto-janai, line 197: I'm not sure if this is correct in the context of the previous atomics (this function could return `Ready` after 118 but before 122).

hinto-janai (changes requested):

p2p/async-buffer/src/lib.rs
Boog900: No, this is a single producer, single consumer, so this can't happen

hinto-janai: Ahhh I missed that, my bad.

p2p/async-buffer/tests/basic.rs
hinto-janai: Less of a concern on SPSC.

p2p/async-buffer/src/lib.rs
hinto-janai: Ignore since SPSC.

hinto-janai (approved):

PR #110, opened 2024-04-18
test-utils: add `rpc` module

hinto-janai:
Adds a `cuprate_test_utils::rpc` module for generating `cuprate_type` types from `monerod`'s RPC.


test-utils/src/rpc/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+//!     // Create RPC client.
+//!     let rpc = HttpRpcClient::new(None).await;
+//!
+//!     // Collect 20 blocks.
+//!     let mut vec: Vec<VerifiedBlockInformation> = vec![];
+//!     for height in (3130269 - 20)..3130269 {
+//!         vec.push(rpc.get_verified_block_information(height).await);
+//!     }
+//! }
+//! ```
```
hinto-janai, line 19: Usage.

Pretty simple, just create a client and fetch blocks (but typed).

test-utils/src/rpc/client.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// Assert the default address is localhost.
+    #[tokio::test]
+    async fn localhost() {
+        assert_eq!(HttpRpcClient::new(None).await.address(), LOCALHOST_RPC_URL);
+    }
+
+    /// Assert blocks are correctly received/calculated.
+    #[ignore] // FIXME: doesn't work in CI, we need a real unrestricted node
+    #[tokio::test]
+    async fn get() {
```
hinto-janai, line 180: TODO: It would be nice if our CI had access to `monerod` RPC calls.

We could make our current CI `monerod` forward RPC calls from other nodes but unfortunately [I believe getting `pow_hash` in the response requires an unrestricted node](https://github.com/monero-project/monero/blob/c8214782fb2a769c57382a999eaf099691c836e7/src/rpc/core_rpc_server.cpp#L2615) so maybe we could get our own 24/7 node for testing purposes? `username:password` would be passed via CI env variables I guess.

test-utils/src/rpc/client.rs
```diff
@@ -0,0 +1,275 @@
...
+            .into_iter()
+            .enumerate()
+            .map(|(i, tx)| {
+                let tx_hash = tx.hash();
+                assert_eq!(tx_hash, tx_hashes[i]);
+                TransactionVerificationData {
+                    tx_blob: tx.serialize(),
+                    tx_weight: tx.weight(),
+                    tx_hash,
+                    fee: tx.rct_signatures.base.fee,
```
hinto-janai, line 158: Is this the correct transaction `fee` mapping?

test-utils/src/rpc/client.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        #[derive(Debug, Deserialize)]
+        struct BlockHeader {
+            block_weight: usize,
+            long_term_weight: usize,
+            cumulative_difficulty: u128,
+            hash: String,
+            height: u64,
+            pow_hash: String,
+            reward: u64, // generated_coins
```
hinto-janai, line 80: Is this correct? `reward == generated_coins`? `reward` is the field name in the RPC `get_block` response.

hinto-janai (commented):
@Boog900 @SyntheticBird45 ready.

test-utils/src/rpc/client.rs
Boog900: no, reward include tx fees, generated coins is without fees

test-utils/src/rpc/client.rs
Boog900: yep monero-serai fills this in for v1 txs

test-utils/src/rpc/client.rs
Boog900: We could but this is test code, for the test code so not high importance IMHO, probably can leave this test to be run every so often manually.

Boog900 (changes requested):

test-utils/src/rpc/client.rs
```diff
@@ -0,0 +1,305 @@
...
+            .outputs
+            .iter()
+            .map(|output| output.amount.expect("miner_tx amount was None"))
+            .sum::<u64>()
+            - total_tx_fees;
+        assert_eq!(
+            reward,
+            generated_coins + total_tx_fees,
+            "generated_coins ({generated_coins}) + total_tx_fees ({total_tx_fees}) != reward ({reward})"
+        );
```
hinto-janai, line 138: Correct calculation of `generated_coins` and `reward` and `total_tx_fees`? (RPC tests all pass)

Boog900: yep, correct

Boog900 (approved):

PR #111, opened 2024-04-19
dandelion++ lib

Boog900:
This crate/lib will implement D++ using the `Service` trait to be generic over clients.

To fully implement D++ you need knowledge of and control over the tx pool which would be pretty annoying to abstract over so this crate will contain a minimal D++ router that will just handle tx routing and nothing to do with keeping track of tx states etc.

This will still have a D++ txpool abstraction for completeness which Cuprate will use.


Boog900:
@hinto-janai @SyntheticBird45 this is ready for review, but isn't urgent, I can work on P2P code without this for now.

Although I am not expecting you to check this matches spec, you may still find the dandelion++ paper useful: https://arxiv.org/pdf/1805.11060.pdf

p2p/dandelion/src/router.rs
```diff
@@ -0,0 +1,384 @@
...
+        for (peer_id, peer_svc) in &mut self.stem_peers {
+            if ready!(peer_svc.poll_ready(cx)).is_err() {
+                failed_peers_ids.push(peer_id.clone());
+            }
+        }
+
+        for failed_peer in failed_peers_ids {
+            tracing::debug!(parent: &self.span, "Peer returned an error on `poll_ready`, removing from router.");
+            self.stem_peers.remove(&failed_peer);
+        }
```
hinto-janai, line 242: Is there a iterator version so the vec doesn't have to grow/exist?

p2p/dandelion/src/router.rs
```diff
@@ -0,0 +1,384 @@
...
+            epoch_start: Instant::now(),
+            local_route: None,
+            stem_routes: HashMap::new(),
+            stem_peers: HashMap::new(),
+            state_dist,
+            config,
+            span: tracing::debug_span!("dandelion_router", state = ?current_state),
+            _tx: PhantomData,
+        }
+    }
```
hinto-janai, line 383: If every field must be set I don't think it needs to be/is a builder.

This could just take the values as input in a constructor function directly instead of panics? If builder style is still wanted, a panic doc would be nice.

p2p/dandelion/src/tests/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        Key = usize,
+        Service = impl Service<
+            Req,
+            Future = impl Future<Output = Result<(), tower::BoxError>> + Send + 'static,
+            Error = tower::BoxError,
+        > + Send
+                      + 'static,
+        Error = tower::BoxError,
+    >,
+    UnboundedReceiver<(u64, Req)>,
```
hinto-janai, line 30: Type aliases?

p2p/dandelion/src/tests/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+                            .get_mut(&tx_id)
+                            .map(|tx| tx.1 = State::Fluff);
+
+                        Ok(TxStoreResponse::Ok)
+                    }
+                }
+            }
+        }),
+        txs_2,
+    )
```
hinto-janai, line 129: Bind to variable instead of indenting this much code IMO

p2p/dandelion/src/tests/router.rs
```diff
@@ -0,0 +1,204 @@
...
+    broadcast_rx.close();
+
+    for _ in 0..30 {
+        router
+            .ready()
+            .await
+            .unwrap()
+            .call(DandelionRouteReq {
+                tx: 0_usize,
+                state: TxState::Stem { from: FROM_PEER },
```
hinto-janai, line 41: `FROM_PEER` only used once?

p2p/dandelion/src/tests/router.rs
```diff
@@ -0,0 +1,204 @@
...
+use tower::{Service, ServiceExt};
+
+use crate::{DandelionConfig, DandelionRouteReq, DandelionRouterBuilder, Graph, TxState};
+
+use super::*;
+
+#[tokio::test]
+async fn routes_consistent() {
+    // make sure a tx from the same peer goes to the same peer.
+
```
hinto-janai, line 12: Shouldn't this be a doc comment? Same with other tests below.

p2p/dandelion/src/tests/router.rs
```diff
@@ -0,0 +1,204 @@
...
+    const FROM_PEER: usize = 20;
+
+    for _ in 0..30 {
+        router
+            .ready()
+            .await
+            .unwrap()
+            .call(DandelionRouteReq {
+                tx: 0_usize,
+                state: TxState::Stem { from: FROM_PEER },
```
hinto-janai, line 143: `FROM_PEER` again.

p2p/dandelion/src/traits.rs
```diff
@@ -0,0 +1,49 @@
+/// A request to diffuse a transaction to all connected peers.
+///
+/// This crate does not handle diffusion it is left to implementers.
+pub struct DiffuseRequest<Tx>(pub Tx);
+
+/// A request sent to a single peer to stem this transaction.
+pub struct StemRequest<Tx>(pub Tx);
```
hinto-janai, line 7:
```suggestion
/// A request to diffuse a transaction to all connected peers.
///
/// This crate does not handle diffusion it is left to implementers.
#[repr(transparent)]
pub struct DiffuseRequest<Tx>(pub Tx);

/// A request sent to a single peer to stem this transaction.
#[repr(transparent)]
pub struct StemRequest<Tx>(pub Tx);
```

p2p/dandelion/src/traits.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[cfg(feature = "txpool")]
+/// A request sent to the backing transaction pool storage.
+pub enum TxStoreRequest<Tx, TxID> {
```
hinto-janai, line 11: Please explain in `lib.rs` docs on what features there are and what they do.

p2p/dandelion/src/txpool.rs
```diff
@@ -0,0 +1,545 @@
...
+#[error("The dandelion pool was shutdown")]
+pub struct DandelionPoolShutDown;
+
+/// A request to the dandelion tx pool.
+///
+/// This enum does not contain requests to get tx data from the pool, this is to reduce the amount
+/// of requests that get sent to the [`DandelionPool`]. Users should hold a handle to the backing tx storage
+/// to get transaction data from the pool.
+///
+/// It MUST be remembered to not expose the private stem pool to peers.
```
hinto-janai, line 41: Seems like the type system should be used here...

p2p/dandelion/src/txpool.rs
```diff
@@ -0,0 +1,545 @@
...
+    /// An incoming tx.
+    IncomingTx {
+        /// The transaction.
+        tx: Tx,
+        /// The transaction ID.
+        tx_id: TxID,
+        /// The routing state of this transaction.
+        tx_state: TxState<PID>,
+    },
+}
```
hinto-janai, line 56: Will there be more variants in the future?

p2p/dandelion/src/txpool.rs
```diff
@@ -0,0 +1,545 @@
...
+    ) -> Result<(), tower::BoxError> {
+        let TxStoreResponse::Contains(have_tx) = self
+            .backing_pool
+            .ready()
+            .await?
+            .call(TxStoreRequest::Contains(tx_id.clone()))
+            .await?
+        else {
+            panic!("Backing tx pool responded with wrong response for request.");
+        };
```
hinto-janai, line 255: Maybe a TODO for another time but `expected: Response::X(_), found: Response::Y(_)` would be helpful.

p2p/dandelion/src/txpool.rs
```diff
@@ -0,0 +1,545 @@
...
+
+        Ok(tx.map(|tx| tx.0))
+    }
+
+    async fn run(
+        mut self,
+        mut rx: mpsc::Receiver<(DandelionPoolRequest<Tx, TxID, PID>, oneshot::Sender<()>)>,
+    ) {
```
hinto-janai, line 359: This starts the txpool?

p2p/dandelion/src/txpool.rs
```diff
@@ -0,0 +1,545 @@
...
+
+        let span = tracing::debug_span!("dandelion_pool");
+
+        tokio::spawn(pool.run(rx).instrument(span));
+
+        DandelionPoolService {
+            tx: PollSender::new(tx),
+        }
+    }
+}
```
hinto-janai, line 545: Either no builder or panic docs please, there's no mention that fields must be filled on this one.

p2p/dandelion/src/config.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// The duration of an epoch.
+    pub epoch_duration: Duration,
+    /// q in the dandelion paper, this is the probability that a node will be in the fluff state for
+    /// a certain epoch.
+    ///
+    /// The dandelion paper recommends to make this value small, but the smaller this value, the higher
+    /// the broadcast latency.
+    ///
+    /// It is recommended for this value to be <= `0.2`, this value *MUST* be between 0 and 1 (not equal to either).
+    pub fluff_probability: f64,
```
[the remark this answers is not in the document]
Boog900: > Is there a type safe version that could be used

I don't _think_ so

> maybe change this to a builder and assert upon construction

 I think having a builder would be a bit overkill here as an invalid value will cause a panic anyway when making the distribution.

p2p/dandelion/src/config.rs
```diff
@@ -0,0 +1,149 @@
...
+impl DandelionConfig {
+    /// Returns the number of outbound peers to use to stem transactions.
+    ///
+    /// This value depends on the [`Graph`] chosen.
+    pub fn number_of_stems(&self) -> usize {
+        match self.graph {
+            Graph::Line => 1,
+            Graph::FourRegular => 2,
+        }
+    }
```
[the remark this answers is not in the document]
Boog900: I don't think this is necessary, these values will never change

p2p/dandelion/src/router.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// A [`HashMap`] linking peer's IDs to IDs in `stem_peers`.
+    pub(crate) stem_routes: HashMap<ID, ID>,
+    /// Peers we are using for stemming.
+    ///
+    /// This will contain peers, even in [`State::Fluff`] to allow us to stem [`TxState::Local`]
+    /// transactions.
+    pub(crate) stem_peers: HashMap<ID, S>,
+
+    /// The distribution to sample to get the [`State`], true is [`State::Fluff`].
+    pub(crate) state_dist: Bernoulli,
```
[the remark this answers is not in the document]
Boog900: none of them, I think I changed the test, will remove this

p2p/dandelion/src/router.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    _tx: PhantomData<Tx>,
+}
+
+impl<Tx, ID, P, B, S> DandelionRouter<P, B, ID, S, Tx>
+where
+    ID: Hash + Eq + Clone,
+    P: Discover<Key = ID, Service = S, Error = tower::BoxError>,
+    B: Service<DiffuseRequest<Tx>, Error = tower::BoxError>,
+    S: Service<StemRequest<Tx>, Error = tower::BoxError>,
```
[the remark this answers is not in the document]
Boog900: slightly more verbose but I like specifying `tower::BoxError`

p2p/dandelion/src/router.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn stem_tx(&mut self, tx: Tx, from: ID) -> S::Future {
+        loop {
+            let stem_route = self.stem_routes.entry(from.clone()).or_insert_with(|| {
+                self.stem_peers
+                    .iter()
+                    .choose(&mut thread_rng())
+                    .expect("No peers in `stem_peers` was poll_ready called?")
+                    .0
+                    .clone()
+            });
```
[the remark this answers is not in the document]
Boog900: Maybe, I know this is very inefficient but the maximum amount of peers is 2 and this will only be ran once every 10 mins per inbound connection. So if you know a way I can improve it I will, but this shouldn't be an issue.

p2p/dandelion/src/router.rs
Boog900: Yeah I thought there were more fields than there turned out to be which is why I created a builder, seen as there are only 3 I'll remove the builder

p2p/dandelion/src/tests/mod.rs
Boog900: This is test code so I think this is fine.

p2p/dandelion/src/tests/mod.rs
Boog900: I don't think that would clear this up much, plus I don't think this is too bad considering it is test code.

p2p/dandelion/src/tests/mod.rs
Boog900: can't do impl trait in type aliases

p2p/dandelion/src/tests/router.rs
Boog900: do doc comments make sense for tests?

I'll change these though

p2p/dandelion/src/traits.rs
Boog900: Why? no one should be transmuting these, surly we should let Rust decide the layout?

p2p/dandelion/src/txpool.rs
Boog900: We can't enforce this on outside crates without requiring all requests to the txpool to go through the dandelion pool, which would add unnecessary indirection.

Even then crates could still misuses this this.

Cuprate will enforce it in it's own tx-pool but it's not something we can do here

p2p/dandelion/src/txpool.rs
Boog900: no I'll make this a struct

p2p/dandelion/src/txpool.rs
Boog900: It's the pools task, so this is function you put in a `tokio::spawn`

p2p/dandelion/src/config.rs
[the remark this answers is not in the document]
hinto-janai: Up to you here but asserting constants don't change over time as code changes is a big reason for tests.

p2p/dandelion/src/tests/router.rs
hinto-janai: Btw, not saying this is bad - it could actually be more clear, just wondering if it was intentional.

p2p/dandelion/src/tests/router.rs
hinto-janai: Shows intention that the comment is describing the whole test and what it does, rather than some code below it.

p2p/dandelion/src/traits.rs
hinto-janai: Should have noted it was a suggestion if applicable.

I'd usually follow [this](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits) (and extend it to `repr`) if you wanted to guarantee layout for users, up to you here.

p2p/dandelion/src/router.rs
```diff
@@ -109,6 +119,33 @@ where
     B: Service<DiffuseRequest<Tx>, Error = tower::BoxError>,
     S: Service<StemRequest<Tx>, Error = tower::BoxError>,
 {
+    /// Creates a new [`DandelionRouter`], with the provided services and config.
+    pub fn new(broadcast_svc: B, outbound_peer_discover: P, config: DandelionConfig) -> Self {
```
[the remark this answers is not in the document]
hinto-janai: Not sure if range is correct but there should be a panic doc.

p2p/dandelion/src/config.rs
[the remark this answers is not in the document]
Boog900: added a test for this

p2p/dandelion/src/tests/router.rs
Boog900: I'll add it back, yeah it was

p2p/dandelion/src/router.rs
```diff
@@ -228,17 +265,19 @@ where
...
+                    true
+                }
+            });
 
-        for failed_peer in failed_peers_ids {
-            tracing::debug!(parent: &self.span, "Peer returned an error on `poll_ready`, removing from router.");
-            self.stem_peers.remove(&failed_peer);
+        if peers_pending {
+            return Poll::Pending;
         }
```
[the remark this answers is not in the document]
Boog900: added back tracing but not on the pending branch, as pending does not mean an error, it just means it's not ready yet

p2p/dandelion/src/router.rs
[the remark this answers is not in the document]
hinto-janai: Up to you, I would be for more docs than less when it comes to [errors/panics](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure).

The field gets linked in `cargo doc`, are you talking about editor resolving?

p2p/dandelion/src/tests/router.rs
```diff
@@ -6,6 +6,59 @@ use crate::{DandelionConfig, DandelionRouteReq, DandelionRouter, Graph, TxState}
...
+        time_between_hop: Duration::from_millis(175),
+        epoch_duration: Duration::from_secs(60_000),
+        fluff_probability: 0.0, // we want to be in stem state
+        graph: Graph::FourRegular,
+    };
+
+    let (broadcast_svc, _broadcast_rx) = mock_broadcast_svc();
+    let (outbound_peer_svc, _outbound_rx) = mock_discover_svc();
+
+    let mut router = DandelionRouter::new(broadcast_svc, outbound_peer_svc, config.clone());
```
hinto-janai, line 22: https://github.com/Cuprate/cuprate/actions/runs/8944812790/job/24572545093?pr=111
```suggestion
    let mut router = DandelionRouter::new(broadcast_svc, outbound_peer_svc, config);
```

hinto-janai (changes requested):

p2p/dandelion/src/router.rs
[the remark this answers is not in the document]
Boog900: > The field gets linked in cargo doc, are you talking about editor resolving?

Yeah I was my bad.

hinto-janai (approved):

PR #113, opened 2024-04-23
database: impl `service` fn bodies 

hinto-janai:
## Part 22
https://github.com/Cuprate/cuprate/pull/104 <- Previous Next -> https://github.com/Cuprate/cuprate/pull/117

- Implements the `service` function signatures introduced in https://github.com/Cuprate/cuprate/pull/101 using the database functions introduced in https://github.com/Cuprate/cuprate/pull/102.

Requires https://github.com/Cuprate/cuprate/pull/102 to be merged first.


database/src/service/read.rs
```diff
@@ -230,55 +241,81 @@ fn map_request(
...
+            .into_par_iter()
+            .map(|block_height| {
+                let tx_ro = env_inner.tx_ro()?;
+                let tables = env_inner.open_tables(&tx_ro)?;
+                get_block_extended_header_from_height(&block_height, &tables)
+            })
+            .collect::<Result<Vec<ExtendedBlockHeader>, RuntimeError>>()?
+    };
+
+    Ok(Response::BlockExtendedHeaderInRange(vec))
```
hinto-janai, line 286: @Boog900 Re: https://github.com/Cuprate/cuprate/pull/102#discussion_r1554580888

LMDB docs:
> A thread can only use one transaction at a time, plus any child transactions. Each transaction belongs to one thread.

And `heed` (issue 20):

> A thread is a single unit of execution. So is a transaction. It is nonsensical to associate them other than 1:1.

So we'll be opening a read transaction + table(s) per reader thread.

And although `redb`'s is `Send + Sync`, it seems like the read transaction + table per thread model is correct there too:

https://github.com/cberner/redb/blob/0d0e11ea901a3fe8a7ec4ac0b2b4aee121d11461/benches/lmdb_benchmark.rs#L220-L225

Boog900: If we open RO LMDB txs with `MDB_NOTLS` we can impl Send for our read only txs.

We can then use a [thread local](https://docs.rs/thread_local/latest/src/thread_local/lib.rs.html#103-111) which will hold the tables, so we don't need to keep opening txs and tables.

hinto-janai: Are you sure 1 TX per thread is not how reads scale? Issue 20 seems to answer the `MDB_NOTLS` question:
> yes but a txn can only be used by one thread at a time, regardless. concurrent use of a single txn from multiple threads will cause memory corruption.

I believe it's mostly for moving the transaction, not sharing a reference across many threads.

This does break read atomicity though, AFAIK this is prevented in `monerod` with a spinlock, we could use a `RwLock` (only inside `service`) that syncs the reader-pool + 1 writer.

Boog900: > Are you sure 1 TX per thread is not how reads scale

We would still have 1 tx per thread but currently this is doing 1 tx per _block height_, using a thread local allows us to keep 1 tx per thread.

> This does break read atomicity though, AFAIK this is prevented in monerod with a spinlock, we could use a RwLock (only inside service) that syncs the reader-pool + 1 writer.

I am not really worried about this too much, the syncer will have a consistent view as it waits on writes so the worse event that we could get is an RPC (or other interface) request during a reorg when one tx sees one chain and the other sees another, the chance of this happening though are very low.

What we could do to combat this:

- Keep an atomic counter that gets incremented on a reorg (no request to do this yet)
- When a tx is added to the thread local record the atomics value with the tx. i.e (Tx, u64)
- after retrieving all values loop over the thread local and check all the txs have the same value.
- If they differ, try again.

It doesn't matter if they differ with the current atomic value just that they all agree on the state.

This does still allow for 2 txs to disagree on the top block but I can't see an issue with this.

hinto-janai: > We would still have 1 tx per thread but currently this is doing 1 tx per block height, using a thread local allows us to keep 1 tx per thread.

Ahhh okay I see the mistake. `heed` has a `read-txn-no-tls` feature for `MDB_NOTLS` I'll enable.

> the syncer will have a consistent view as it waits on writes

IMO the correctness shouldn't depend on external code using it correctly.

> the worse event that we could get is an RPC (or other interface) request during a reorg when one tx sees one chain and the other sees another

Wouldn't a normal `add_block()` call be bad?
```
Reader 1 (same request as reader 2)
   |
   v      Writer
 tx_ro      |
   |        v        Reader 2
   |  tx_rw + commit   |
   |                   v
    \                tx_ro <- different state than reader 1
      \________________|
            |
collecting from different database states
```

> What we could do to combat this:

Do you think a counter + retries would allow more reader throughput than a `RwLock`? Is it worth taking away simplicity? Each function will now need `loop { do(); maybe_break }` style code instead of `lock(); do()`.

Could we use `RwLock` for now and implement a counter later if there's bottlenecks?

Boog900:
> Wouldn't a normal add_block() call be bad?

An add block only adds new data and does not change old data, the worst that can happen is one tx does not see the new data and the request fails, which would have happened anyway, even with locking, if that same request came a couple seconds earlier.

> Do you think a counter + retries would allow more reader throughput than a RwLock

Yes I do, I don't know how significant the difference would be but I don't really want to add a lock for this.

> Each function will now need loop { do(); maybe_break }

Or you could make it recurse, which would be fine as this event is extremely unlikely.

> Could we use RwLock for now and implement a counter later if there's bottlenecks?

I would rather just not handle this now than add a lock.

database/src/service/read.rs
Boog900: > Wouldn't a normal add_block() call be bad?

An add block only adds new data and does not change old data, the worst that can happen is one tx does not see the new data and the request fails, which would have happened anyway, even with locking, if that same request came a couple seconds earlier.

> Do you think a counter + retries would allow more reader throughput than a RwLock

Yes I do, I don't know how significant the difference would be but I don't really want to add a lock for this.

> Each function will now need loop { do(); maybe_break }

Or you could make it recurse, which would be fine as this event is extremely unlikely.

> Could we use RwLock for now and implement a counter later if there's bottlenecks?

I would rather just not handle this now than add a lock.

Boog900:
I am getting compile errors on 1.77.1 because of: https://blog.rust-lang.org/2024/03/30/i128-layout-update.html

I think we are going to have to change `cumulative_difficulty` to `cumulative_difficulty_low` and `cumulative_difficulty_high` on `BlockInfo`

database/src/service/read.rs
```diff
@@ -106,23 +121,15 @@ impl DatabaseReadHandle {
...
-    /// TODO
+    /// Access to the actual database environment.
+    ///
+    /// TODO: we need this for testing but should we allow it publicly?
+    /// Or within `crate`? It allows anyone to start tampering with the
+    /// database directly instead of going through `service`.
     #[inline]
     pub const fn env(&self) -> &Arc<ConcreteEnv> {
         &self.env
     }
```
hinto-janai, line 132: Should the `DatabaseReadHandle` (the read service) give access to the actual database?

It lets anybody outside of `cuprate_database` directly use the database instead of going through `service`.

database/src/service/read.rs
```diff
@@ -228,57 +317,191 @@ fn map_request(
...
+
+        let time_lock = if output
+            .output_flags
+            .contains(OutputFlags::NON_ZERO_UNLOCK_TIME)
+        {
+            // TODO: how to recover the timelock height/time?
+            todo!()
+        } else {
+            Timelock::None
+        };
```
hinto-janai, line 427: Context: `ReadRequest::Outputs`, we're mapping `cuprate_database::types::Output` -> `cuprate_types::OutputOnChain`.

How do we recover this `Timelock`? I don't think we store the value directly anywhere.

We could deserialize the parent transaction from `TxBlob` then get the timelock, but that seems expensive to do for every output?

database/src/service/read.rs
```diff
@@ -228,57 +317,191 @@ fn map_request(
...
+        let key = CompressedEdwardsY::from_slice(&output.key)
+            .map(|y| y.decompress())
+            .unwrap_or(None);
+
+        let output_on_chain = OutputOnChain {
+            height: u64::from(output.height),
+            time_lock,
+            key,
+            commitment,
+        };
```
hinto-janai, line 438: Is `key` correct here?

database/src/service/read.rs
```diff
@@ -228,57 +317,191 @@ fn map_request(
...
+                // INVARIANT: #[cfg] @ lib.rs asserts `usize == u64`
+                #[allow(clippy::cast_possible_truncation)]
+                Ok(count) => Ok((amount, count as usize)),
+                Err(e) => Err(e),
+            }
+        })
+        .collect::<Result<HashMap<Amount, usize>, RuntimeError>>()?;
+
+    Ok(Response::NumberOutputsWithAmount(map))
 }
```
hinto-janai, line 480: What is the valid response to `ReadResponse::NumberOutputsWithAmounts` where the `amount` doesn't exist, i.e. there are 0 outputs with `amount`.

Should it be `Err(KeyNotFound)` or `Ok(0)`?

database/src/service/read.rs
```diff
@@ -228,57 +317,191 @@ fn map_request(
...
+        // `Ok(true)` => a key image was found, return early
+        // `Err` => an error was found, return early
+        //
+        // Else, `Ok(false)` will continue the iterator.
+        .find_any(|result| !matches!(result, Ok(false)))
+    {
+        None | Some(Ok(false)) => Ok(Response::CheckKIsNotSpent(true)), // Key image was NOT found.
+        Some(Ok(true)) => Ok(Response::CheckKIsNotSpent(false)),        // Key image was found.
+        Some(Err(e)) => Err(e), // A database error occurred.
+    }
```
hinto-janai, line 506: `Response::CheckKIsNotSpent(true)` means the key image doesn't exist, right?

hinto-janai (commented):
@Boog900 Some questions.

database/src/service/read.rs
hinto-janai: Think adding an `enum cuprate_types::Exist { Does, DoesNot }` or similar would be nice btw.

database/src/service/read.rs
Boog900: we could but I would rather not, probably should put a warning on it though

database/src/service/read.rs
Boog900: From the `Tx Unlock Time` table: https://github.com/Cuprate/cuprate/tree/db-schema/database#tx-unlock-time

database/src/service/read.rs
Boog900: yep

database/src/service/read.rs
Boog900: It should be `Ok(0)`

database/src/service/read.rs
Boog900: IIRC I switched it to this in the document consensus PR so for now just put it as a big TODO to make sure, if this is incorrect all txs will fail verification anyway.

database/src/service/read.rs
hinto-janai: Ahhh...

database/src/service/tests.rs
```diff
@@ -7,76 +7,364 @@
...
+/// - Assert read requests lead to expected responses
+#[allow(clippy::future_not_send)] // INVARIANT: tests are using a single threaded runtime
+async fn test_template(
+    // Which block(s) to add?
+    block_fns: &[fn() -> &'static VerifiedBlockInformation],
+    // Total amount of generated coins after the block(s) have been added.
+    cumulative_generated_coins: u64,
+    // What are the table lengths be after the block(s) have been added?
+    assert_table_len: AssertTableLen,
+) {
```
hinto-janai, line 78: Tests all `Request`'s lead to expected `Response`'s, which is currently tested on...

database/src/service/tests.rs
```diff
@@ -7,76 +7,364 @@
...
+            prunable_tx_blobs: 0,
+            rct_outputs: 0,
+            tx_blobs: 0,
+            tx_ids: 0,
+            tx_heights: 0,
+            tx_unlock_time: 0,
+        },
+    )
+    .await;
+}
```
hinto-janai, line 370: ...all the blocks we have currently in `test-utils/`.

database/src/service/write.rs
```diff
@@ -226,5 +229,26 @@ impl DatabaseWriter {
...
+        Err(e) => {
+            // INVARIANT: ensure database atomicity by aborting
+            // the transaction on `add_block()` failures.
+            tx_rw
+                .abort()
+                .expect("could not maintain database atomicity by aborting write transaction");
+            Err(e)
+        }
+    }
 }
```
hinto-janai, line 254: If adding a block fails we should panic if we cannot maintain atomicity, correct?

Otherwise the database may proceed with half-inserted data.

database/src/free.rs
```diff
@@ -1,10 +1,48 @@
...
+        .map(|y| y.decompress())
+        .unwrap_or(None);
+
+    Ok(OutputOnChain {
+        height: u64::from(output.height),
+        time_lock,
+        key,
+        commitment,
+    })
+}
```
hinto-janai, line 45: This is kinda awkward but I'm not sure where else to put this.

This is needed since we need to map `Output` -> `OutputOnChain` in:
- Reader service's `output()` response function
- Reader service's test to make sure the `Response` is returning correct `OutputOnChain`'s.

database/src/backend/heed/database.rs
```diff
@@ -49,6 +49,17 @@ pub(super) struct HeedTableRw<'env, 'tx, T: Table> {
...
+/// SAFETY: `cuprate_database`'s Cargo.toml enables a feature
+/// for `heed` that turns on the `MDB_NOTLS` flag for LMDB.
+///
+/// This makes read transactions `Send`, but only if that flag is enabled.
+///
+/// This is required as in `crate::service` we must put our transactions and
+/// tables inside `ThreadLocal`'s to use across multiple threads.
+///
+/// `ThreadLocal<T>` requires that `T: Send`.
+unsafe impl<T: Table> Send for HeedTableRo<'_, T> {}
```
hinto-janai, line 61: This is safe?

The feature makes `heed::RoTxn: Send`, and I believe `heed::Database` is `Send` for all of our cases.

hinto-janai (commented):
@Boog900 ready.

database/src/backend/heed/database.rs
Boog900: This is making `heed::RoTxn` `Sync`, which I don't think is safe.

For our current use cases this is ok as we aren't using them between threads however this would be a soundness hole. 

I only really see 2 ways of making this sound: add a mutex on the inner tx or mark every method on `HeedTableRo` unsafe. Both are not great.

database/src/free.rs
```diff
@@ -1,10 +1,48 @@
...
+) -> Result<OutputOnChain, RuntimeError> {
+    // FIXME: implement lookup table for common values:
+    // <https://github.com/monero-project/monero/blob/c8214782fb2a769c57382a999eaf099691c836e7/src/ringct/rctOps.cpp#L322>
+    let commitment = ED25519_BASEPOINT_POINT + H() * Scalar::from(amount);
+
+    let time_lock = if output
+        .output_flags
+        .contains(OutputFlags::NON_ZERO_UNLOCK_TIME)
+    {
+        Timelock::Time(table_tx_unlock_time.get(&output.tx_idx)?)
```
Boog900, line 30: this needs to branch on the value of the timelock - right now it is setting them all to a time based lock.

https://github.com/serai-dex/serai/blob/bc1dec79917d37d326ac3d9bc571a64131b0424a/coins/monero/src/transaction.rs#L139

database/src/service/read.rs
```diff
@@ -228,57 +342,177 @@ fn map_request(
...
+    let inner_map = |amount, amount_index| {
+        let (tx_ro, tables) = get_tx_ro_and_tables!(env_inner, tx_ro, tables);
+
+        let pre_rct_output_id = PreRctOutputId {
+            amount,
+            amount_index,
+        };
+
+        let output = get_output(&pre_rct_output_id, tables.outputs())?;
+        let output_on_chain = output_to_output_on_chain(&output, amount, tables.tx_unlock_time())?;
```
Boog900, line 440: This needs to handle getting RCT outputs as as well, currently it is only getting pre-rct outputs

database/src/service/read.rs
```diff
@@ -228,57 +342,177 @@ fn map_request(
...
+
+            match tables.num_outputs().get(&amount) {
+                // INVARIANT: #[cfg] @ lib.rs asserts `usize == u64`
+                #[allow(clippy::cast_possible_truncation)]
+                Ok(count) => Ok((amount, count as usize)),
+                // If we get a request for an `amount` that doesn't exist,
+                // we return `0` instead of an error.
+                Err(RuntimeError::KeyNotFound) => Ok((amount, 0)),
+                Err(e) => Err(e),
+            }
```
Boog900, line 481: I think we need to handle RCT outputs here as well

database/src/service/write.rs
Boog900: yes, however does failing to abort cause data to half write?

Boog900 (changes requested):

database/src/backend/heed/database.rs
hinto-janai: We're using `read-txn-no-tls` which only allows `Send`. `sync-read-txn` is removed in the version of `heed` we're using. 
 
image

image

Sidenote but this would be a good excuse to add `static_assertions` so we assert traits (among other things).

Boog900: I should have been more clear, but this is what `HeedTableRo` looks like:

```rust
pub(super) struct HeedTableRo<'tx, T: Table> {
    /// An already opened database table.
    pub(super) db: HeedDb<T::Key, T::Value>,
    /// The associated read-only transaction that opened this table.
    pub(super) tx_ro: &'tx heed::RoTxn<'tx>,
}
```

So impl'ing `Send` for `HeedTableRo` means you are making `heed::RoTxn` `Sync` (in this struct). You could create 2 `HeedTableRo` from the same tx and send one to another thread, then you would have the same tx shared between 2 threads.

database/src/service/write.rs
hinto-janai: - `heed`'s abort is infallible and panics if the tx was already commited (API doesn't let us do this so it's unreachable)
- `redb`'s abort is fallible: https://docs.rs/redb/2.0.0/src/redb/transactions.rs.html#939

The above eventually calls https://docs.rs/redb/2.0.0/src/redb/tree_store/page_store/page_manager.rs.html#600-642 which always returns `Ok(())` or panics. So, regardless we will panic if data cannot be rolled back. I think we should keep panics here too as to rely on the signature and not the impl.

database/src/backend/heed/database.rs
hinto-janai: Ahhh I see, `&T: Send` only if `T: Sync`. I think Mutex is how `redb` does it?

I know this is rough but in our case (non-contended mutex) it's 3-5% slower. I think it's okay unless you preferred something else, although this would allow us to `impl Sync` for our tables too? No more `ThreadLocal`?

```rust
use heed::{types::*, *};
use std::{hint::black_box, sync::Mutex, time::Instant};

fn main() {
    let dir = tempfile::tempdir().unwrap();
    let env = unsafe { EnvOpenOptions::new().open(dir.path()).unwrap() };

    let mut tx_rw = env.write_txn().unwrap();
    let db: Database<Str, Str> = env.create_database(&mut tx_rw, None).unwrap();

    db.put(&mut tx_rw, "a", "").unwrap();
    tx_rw.commit().unwrap();

    const RANGE: std::ops::Range<usize> = 0..10_000_000;

    loop {
        let now = Instant::now();
        for _ in RANGE {
            let tx_ro = black_box(Mutex::new(env.read_txn().unwrap()));
            let _get = black_box(db.get(&tx_ro.lock().unwrap(), "zero")).unwrap();
        }
        println!("mutex: {}", now.elapsed().as_secs_f32());

        let now = Instant::now();
        for _ in RANGE {
            let tx_ro = black_box(env.read_txn()).unwrap();
            let _get = black_box(db.get(&tx_ro, "zero")).unwrap();
        }
        println!("tx_ro: {}", now.elapsed().as_secs_f32());
    }
}

```
```
mutex: 0.8341315
tx_ro: 0.76794577
mutex: 0.7981665
tx_ro: 0.7648049
mutex: 0.7917462
tx_ro: 0.76621574
```

database/src/service/read.rs
hinto-janai: If RCT, what number are we supposed to return? The total amount of RCT outputs? `tables.rct_outputs().len()`?

database/src/backend/heed/database.rs
Boog900: > although this would allow us to impl Sync for our tables too? No more ThreadLocal?

We would then have contention between threads retrieving data, only one will be able to at a time.

>  I think Mutex is how redb does it

Does this mean only one thread can get data an once? if so I would recommend putting that in a thread local as well 

>  3-5% slower

Not awful, still annoying though, probably our best option

database/src/service/read.rs
Boog900: yes total amount of RCT outputs

Boog900:
@hinto-janai When adding new blocks, I don't think the current code is adding the miner tx to the tx tables, this has to be done _before_ adding the normal txs in the block so the output indexes are correct

database/src/ops/block.rs
```diff
@@ -89,6 +89,19 @@ pub fn add_block(
...
+            fee: 0,
+            tx_hash: block.block.miner_tx.hash(),
+        },
+        &chain_height,
+        tables,
+    )?;
+
     for tx_verification_data in &block.txs {
         add_tx(tx_verification_data, &chain_height, tables)?;
     }
```
hinto-janai, line 107: `add_tx()` takes in a `TransactionVerificationData`, is there a cheaper way to map our `block.block.miner_tx: Transaction` to it?

I'm not sure what `fee` should be here either.

database/src/backend/heed/database.rs
```diff
@@ -49,6 +49,21 @@ pub(super) struct HeedTableRw<'env, 'tx, T: Table> {
...
+/// for `heed` that turns on the `MDB_NOTLS` flag for LMDB.
+/// This makes read transactions `Send`, but only if that flag is enabled.
+///
+/// 2. Our `tx_ro` is wrapped in Mutex, as `&T: Send` only if `T: Sync`.
+/// This is what is happening as we have `&TxRw`, not `TxRw`.
+/// <https://github.com/Cuprate/cuprate/pull/113#discussion_r1582189108>
+///
+/// This is required as in `crate::service` we must put our transactions and
+/// tables inside `ThreadLocal`'s to use across multiple threads.
+unsafe impl<T: Table> Send for HeedTableRo<'_, T> {}
```
hinto-janai, line 65: Clippy gives scary warnings without this allow, pretty sure it's a false positive?

hinto-janai: Only happens after wrapping it in `Mutex` btw, weird.

Boog900: It's because the struct is:
```rust
pub(super) struct HeedTableRo<'tx, T: Table> {
    /// An already opened database table.
    pub(super) db: HeedDb<T::Key, T::Value>,
    /// The associated read-only transaction that opened this table.
    pub(super) tx_ro: Mutex<&'tx heed::RoTxn<'tx>>,
}
```
It needs to be:

 ```rust
 pub(super) struct HeedTableRo<'tx, T: Table> {
    /// An already opened database table.
    pub(super) db: HeedDb<T::Key, T::Value>,
    /// The associated read-only transaction that opened this table.
    pub(super) tx_ro: &'tx Mutex<heed::RoTxn<'tx>>,
}
 ```
 
 Otherwise the mutex isn't actually doing anything. This means you will have to do the same changes that you did to wrap write txs in a refcell :(

hinto-janai: For future reference:

Discussion in matrix lead to the solution of creating and using a private, separate [`struct UnsafeSendable<T>(T)`](https://github.com/Cuprate/cuprate/pull/113/commits/06a2a869a7d0cc61e115e938578b437173886028) that implements `Send`.

We use it for `HeedTableRo` in [`service`](https://github.com/Cuprate/cuprate/pull/113/commits/08d5715ba96ce663e5c825dee9dcabfe54830d10) where our use-case (inside `ThreadLocal<T: Send>`) is safe.

database/src/service/read.rs
```diff
@@ -228,57 +308,210 @@ fn map_request(
...
+
+        let id = PreRctOutputId {
+            amount,
+            amount_index,
+        };
+
+        let output_on_chain = id_to_output_on_chain(&id, tables)?;
+
+        Ok((amount_index, output_on_chain))
+    };
```
hinto-janai, line 418: When returning outputs, how are we supposed to differentiate post-RCT miner outputs? `id_to_output_on_chain()` here checks `if amount == 0 { /* rct */ } else { /*...*/ }` but that skips miner outputs, no?

database/src/service/read.rs
```diff
@@ -228,57 +308,210 @@ fn map_request(
...
+                    #[allow(clippy::cast_possible_truncation)]
+                    Ok(count) => Ok((amount, count as usize)),
+                    // If we get a request for an `amount` that doesn't exist,
+                    // we return `0` instead of an error.
+                    Err(RuntimeError::KeyNotFound) => Ok((amount, 0)),
+                    Err(e) => Err(e),
+                }
+            }
+        })
+        .collect::<Result<HashMap<Amount, usize>, RuntimeError>>()?;
```
hinto-janai, line 477: Same here.

database/src/ops/tx.rs
```diff
@@ -128,7 +128,7 @@ pub fn add_tx(
     };
 
     let mut amount_indices = Vec::with_capacity(tx.prefix.outputs.len());
-    let tx_idx = get_num_tx(tables.tx_ids_mut())?;
+    let tx_idx = get_num_tx(tables.tx_ids())?.saturating_sub(1);
```
hinto-janai, line 131: I think this change is correct? It was causing problems in tests (1st transaction added had index of 1).

hinto-janai (commented):
@Boog900 questions :)

database/src/service/read.rs
Boog900: miner txs are stored under 0 as well even though they have a none zero amount so this is ok

database/src/ops/tx.rs
Boog900: I don't think so, this is shadowing the correct index at the top of this function, was that intentional?

database/src/service/read.rs
```diff
@@ -211,6 +226,71 @@ fn map_request(
...
+                        //
+                        // <https://github.com/Cuprate/cuprate/pull/113#discussion_r1581684698>
+                        Ok(tables) => Ok(unsafe { crate::unsafe_sendable::UnsafeSendable::new(tables) }),
+                        Err(e) => Err(e),
+                    }
+                })?.as_ref()
+            }
+        }
+    }};
+}
```
hinto-janai, line 292: How `tx_ro` + `tables` will work in the `service` reader thread-pool.

- Both get wrapped in `ThreadLocal`
- `tx_ro` is `Send`
- For `redb`, `tables` is `Send`, normal `ThreadLocal::get_or_try` is used
- For `heed`, `tables` is `!Send`, but gets wrapped in `UnsafeSendable`

usage is...

database/src/service/read.rs
```diff
@@ -228,57 +308,210 @@ fn map_request(
...
+        .into_par_iter()
+        .map(|block_height| {
+            let tx_ro = tx_ro.get_or_try(|| env_inner.tx_ro())?;
+            let tables = get_tables!(env_inner, tx_ro, tables);
+            get_block_extended_header_from_height(&block_height, tables)
+        })
+        .collect::<Result<Vec<ExtendedBlockHeader>, RuntimeError>>()?;
+
+    Ok(Response::BlockExtendedHeaderInRange(vec))
 }
```
hinto-janai, line 362: ...here.

- Prep the `tx_ro` + `tables` before entering `rayon`
- Set & use those `ThreadLocal`'s within

`tables` here is either `&impl Tables` or `UnsafeSendable::<impl Tables>::as_ref()`.

database/src/unsafe_sendable.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// of the resulting `Self` is follows all the [`Send`] rules.
+    pub(crate) const unsafe fn new(t: T) -> Self {
+        Self(t)
+    }
+
+    /// Extract the inner `T`.
+    pub(crate) fn into_inner(self) -> T {
+        self.0
+    }
+}
```
hinto-janai, line 54: `UnsafeSendable` impl.

The inner `T` is private so the caller must use `unsafe fn new()` and document their usage.

hinto-janai (commented):
@Boog900 last review I hope, listed recent changes.

database/src/backend/heed/database.rs
```diff
@@ -165,7 +166,8 @@ impl<T: Table> DatabaseRo<T> for HeedTableRo<'_, T> {
 }
 
 //---------------------------------------------------------------------------------------------------- DatabaseRw Impl
-impl<T: Table> DatabaseRo<T> for HeedTableRw<'_, '_, T> {
+// SAFETY: Same as `HeedTableRo`.
+unsafe impl<T: Table> DatabaseRo<T> for HeedTableRw<'_, '_, T> {
```
Boog900, line 170: This is not the same as `HeedTableRo` as write txs are not Send, right?

This is still fine, the comment should probably say that the send bound only applies to read tables

database/src/database.rs
```diff
@@ -82,7 +82,26 @@ pub trait DatabaseIter<T: Table> {
...
+/// However if the table holds a reference to a transaction:
+/// - only the transaction only has to be `Send`
+/// - the table cannot implement `Send`
+///
+/// For example:
+///
+/// `heed`'s transactions are `Send` but `HeedTableRo` contains a `&`
+/// to the transaction, as such, if `Send` were implemented on `HeedTableRo`
+/// then 1 transaction could be used to open multiple tables, then sent to
+/// other threads - this would be a soundness hole against `Sync`.
```
Boog900, line 98: I wouldn't say this is a soundness hole against `Sync`:
```suggestion
/// other threads - this would be a soundness hole against `HeedTableRo`.
```

database/src/free.rs
```diff
@@ -1,10 +1,111 @@
...
+    types::{Amount, Output, OutputFlags, PreRctOutputId, RctOutput},
+    DatabaseRo, RuntimeError,
+};
+
 //---------------------------------------------------------------------------------------------------- Free functions
+/// Map an [`Output`] to a [`cuprate_types::OutputOnChain`].
+#[inline]
+pub(crate) fn output_to_output_on_chain(
+    output: &Output,
+    amount: Amount,
```
Boog900, line 22: would these functions be better in `ops/output`?

database/src/ops/block.rs
```diff
@@ -89,6 +89,19 @@ pub fn add_block(
...
+            fee: block.block.miner_tx.rct_signatures.base.fee,
+            tx_hash: block.block.miner_tx.hash(),
+        },
+        &chain_height,
+        tables,
+    )?;
+
     for tx_verification_data in &block.txs {
         add_tx(tx_verification_data, &chain_height, tables)?;
     }
```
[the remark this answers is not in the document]
Boog900: fee should just be 0, it might be good to split the inputs to that function up as I don't think we need to clone the tx or calculated it's weight

database/src/service/read.rs
Boog900: Couldn't we just store both redb & heed tables in `UnsafeSendable` the overhead should be nothing and it means we don't need to cfg if this high up in the abstraction. Also means this can be a function I think

database/src/service/read.rs
```diff
@@ -211,6 +226,71 @@ fn map_request(
...
+                $tables.get_or_try(|| $env_inner.open_tables($tx_ro))?
+            } else {
+                $tables.get_or_try(|| {
+                    #[allow(clippy::significant_drop_in_scrutinee)]
+                    match $env_inner.open_tables($tx_ro) {
+                        // SAFETY:
+                        // `heed`'s transactions are `Send` but `HeedTableRo` contains a `&`
+                        // to the transaction, as such, if `Send` were implemented on `HeedTableRo`
+                        // then 1 transaction could be used to open multiple tables, then sent to
+                        // other threads - this would be a soundness hole against `Sync`.
```
Boog900, line 265:
```suggestion
                        // then 1 transaction could be used to open multiple tables, then sent to
                        // other threads - this would be a soundness hole against `HeedTableRo`.
```

database/src/service/read.rs
```diff
@@ -211,6 +226,71 @@ fn map_request(
...
+                        //
+                        // `heed::RoTxn: !Sync`, therefore our table
+                        // holding `&heed::RoTxn` must NOT be `Send`.
+                        //
+                        // <https://doc.rust-lang.org/std/marker/trait.Sync.html>
+                        // <https://doc.rust-lang.org/nomicon/send-and-sync.html>
+                        //
+                        // Instead, we are unsafely using `UnsafeSendable` in `service`'s reader
+                        // thread-pool as we are pairing our usage with `ThreadLocal` - only 1 thread
+                        // will ever access a transaction at a time. This is an INVARIANT.
```
Boog900, line 277: Did you mean safely here?

```suggestion
                        // Instead, we are safely using `UnsafeSendable` in `service`'s reader
                        // thread-pool as we are pairing our usage with `ThreadLocal` - only 1 thread
                        // will ever access a transaction at a time. This is an INVARIANT.
```

database/src/service/read.rs
```diff
@@ -228,57 +308,210 @@ fn map_request(
...
+    // Single-threaded, no `ThreadLocal` required.
+    let env_inner = env.env_inner();
+    let tx_ro = env_inner.tx_ro()?;
+    let table_block_heights = env_inner.open_db_ro::<BlockHeights>(&tx_ro)?;
+    let table_block_infos = env_inner.open_db_ro::<BlockInfos>(&tx_ro)?;
+
+    let top_height = top_block_height(&table_block_heights)?;
+    let block_hash = get_block_info(&top_height, &table_block_infos)?.block_hash;
+
+    Ok(Response::ChainHeight(top_height, block_hash))
```
Boog900, line 376: I think this is meant to return the chain height right?

database/src/unsafe_sendable.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use crate::storable::StorableVec;
+
+//---------------------------------------------------------------------------------------------------- Aliases
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd, Eq, Ord, Hash, TransparentWrapper)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[allow(clippy::unsafe_derive_deserialize)] // see safety message on the Send impl
```
Boog900, line 19: I don't think this type needs to be serde

database/src/service/read.rs
hinto-janai: Sure, although making it a function means we have to name everything:
```rust
fn get_tables<'a, EnvInner, TxRo, TxRw, Tables>(
    env_inner: &'a EnvInner,
    tx_ro: &'a TxRo,
    tables: ThreadLocal<UnsafeSendable<Tables>>,
) -> Result<&'a Tables, RuntimeError>
where
    EnvInner: crate::EnvInner<'a, TxRo, TxRw>,
    TxRo: crate::TxRo<'a>,
    TxRw: crate::TxRw<'a>,
    Tables: crate::tables::TablesIter,
{/*...*/}
```
not sure how to tell the compiler the returned value's type == `Tables` here.
```
error[E0308]: mismatched types
   --> database/src/service/read.rs:285:82
    |
246 | fn get_tables<'a, EnvInner, TxRo, TxRw, Tables>(
    |                                         ------ expected this type parameter
...
285 |                 Ok(t) => Ok(unsafe { crate::unsafe_sendable::UnsafeSendable::new(t) }),
    |                                      ------------------------------------------- ^ expected type parameter `Tables`, found associated type
    |                                      |
    |                                      arguments to this function are incorrect
    |
    = note: expected type parameter `Tables`
              found associated type `impl tables::TablesIter`
```
Can it just be a macro?

database/src/database.rs
Boog900: @hinto-janai I don't think this was changed?

hinto-janai: Ah, only the `get_tables!()` doc was updated.

I fixed the above and de-duplicated the docs.

database/src/service/read.rs
Boog900: > Can it just be a macro?

yeah, alright

database/src/unsafe_sendable.rs
```diff
@@ -0,0 +1,87 @@
...
+    ops::{Deref, DerefMut},
+};
+
+use bytemuck::TransparentWrapper;
+
+use crate::storable::StorableVec;
+
+//---------------------------------------------------------------------------------------------------- Aliases
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd, Eq, Ord, Hash, TransparentWrapper)]
+#[allow(clippy::unsafe_derive_deserialize)] // see safety message on the Send impl
```
Boog900, line 15: I don't think this line is needed anymore 

```suggestion
```

Boog900 (changes requested):

Boog900 (approved):

PR #114, opened 2024-04-24
database: split `cumulative_difficulty` into low/high bits

hinto-janai:
- Adds a `map` module in `cuprate_helper` for mapping data types
- Adds functions for splitting/combining `(u64, u64) <-> u128` which can be used for https://github.com/Cuprate/cuprate/pull/113#issuecomment-2073827676.
- Splits `cuprate_database`'s `cumulative_difficulty` into 2 `u64` bit values

There's a bunch of other mapping functions I'd like to add but will do that later.

A note: although `ExtendedBlockHeader` has `cumulative_difficulty: u128`, it seems to compile fine. What should we do here? Is the overall layout for that type okay and it was only `BlockInfo` that was invalid?


helper/src/map.rs
```diff
@@ -0,0 +1,72 @@
...
+#[inline]
+pub const fn combine_low_high_bits_to_u128(low_bits: u64, high_bits: u64) -> u128 {
+    let low = low_bits.to_le_bytes();
+    let high = high_bits.to_le_bytes();
+
+    u128::from_le_bytes([
+        low[0], low[1], low[2], low[3], low[4], low[5], low[6], low[7], high[0], high[1], high[2],
+        high[3], high[4], high[5], high[6], high[7],
+    ])
+}
```
Boog900, line 68: Could we do:

```suggestion
pub const fn combine_low_high_bits_to_u128(low_bits: u64, high_bits: u64) -> u128 {
    let res = (high_bits as u128) << 64;
    res | (low_bits as u128)
}
```

helper/src/map.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    let low = u64::from_le_bytes([
+        bits[0], bits[1], bits[2], bits[3], bits[4], bits[5], bits[6], bits[7],
+    ]);
+    let high = u64::from_le_bytes([
+        bits[8], bits[9], bits[10], bits[11], bits[12], bits[13], bits[14], bits[15],
+    ]);
+
+    (low, high)
+}
```
Boog900, line 39: Also this:

```suggestion
pub const fn split_u128_into_low_high_bits(value: u128) -> (u64, u64) {
    (value as u64, (value >> 64) as u64)
}
```

Boog900 (commented):
`ExtendedBlockHeader` doesn't impl `Pod`/ doesn't need to be stored in the DB right?

helper/src/map.rs
hinto-janai: 44 vs 14 lines of assembly with `black_box` inputs btw.

hinto-janai: And yeah `ExtendedBlockHeader` isn't stored directly, it takes fields from `Block` + `BlockInfo`.

Boog900 (approved):

PR #117, opened 2024-04-26
database: final docs + cleanup

hinto-janai:
## Part 23
Previous <- https://github.com/Cuprate/cuprate/pull/113

- Fixes all clippy issues
- Fixes errors not caught in previous PRs
- Removes dead code
- Documents everything that hasn't been yet

There's still some TODOs and unfinished stuff like `Config` which will be left to be implemented/documented when other things are implemented.


hinto-janai:
@Boog900 Do you prefer `tracing` be added in a different PR? I think it would be easier for you to spot them and suggest changes to what gets logged.

Boog900:
Yeah, would probably be easier to review if they were split

database/src/backend/tests.rs
```diff
@@ -331,6 +322,60 @@ fn db_read_write() {
     }
 }
 
+/// Assert that `key`'s in database tables are sorted in
+/// an ordered B-Tree fashion, i.e. `min_value -> max_value`.
+#[test]
+fn tables_are_sorted() {
```
hinto-janai, line 328: After our dup-key changes, is this trait still required?
```rust
trait Key: Storable {
    fn compare(left: &[u8], right: &[u8]) -> Ordering;
}
```
This test asserts that keys are ordered by value (at least, `u64`).

database/src/config/config.rs
```diff
@@ -26,13 +17,143 @@ use crate::{
...
+    /// [`ConfigBuilder::build`] can be called immediately
+    /// after this function to use default values.
+    pub const fn new() -> Self {
+        Self {
+            db_directory: None,
+            sync_mode: None,
+            reader_threads: None,
+            resize_algorithm: None,
+        }
+    }
```
hinto-janai, line 52: Added `ConfigBuilder`, usage...

database/src/config/mod.rs
```diff
@@ -1,7 +1,44 @@
...
+//!     .sync_mode(SyncMode::Fast)
+//!     // Build into `Config`
+//!     .build();
+//!
+//! // Start a database `service` using this configuration.
+//! let (reader_handle, _) = cuprate_database::service::init(config.clone())?;
+//! // It's using the config we provided.
+//! assert_eq!(reader_handle.env().config(), &config);
+//! # Ok(()) }
+//! ```
```
hinto-janai, line 38: ...here.

database/src/lib.rs
```diff
@@ -82,45 +78,66 @@
 //!
 //! The default is `heed`.
 //!
+//! `tracing` is always enabled and cannot be disable via feature-flag.
+//! <!-- FIXME: tracing should be behind a feature flag -->
```
hinto-janai, line 82: Am willing to do feature gated `tracing`, although I'm not how sure that will work with stuff like `#[instrument]`.

hinto-janai (commented):
@Boog900 ready for review.

Most of the diffs can be skimmed as they are:
- `use` changes
- clippy changes
- typo fixes

I also think `database/README.md` and `cuprate_database` cargo docs are worth looking at again. Each main abstraction layer (traits, `ops`, `service`) has its own doc-test example.

As far as documentation accuracy goes, these files may need closer review:
- `database/src/lib.rs`
- `database/src/ops/mod.rs`
- `database/src/service/mod.rs`
- `database/src/types.rs`
- `database/src/tables.rs`

I've listed code changes and some questions below.

hinto-janai:
Actually, there's some missing docs I'd like to add.

database/src/backend/tests.rs
Boog900: Not needed but if we are going to be getting values using ranges then yes.

Probably something for the future but right now when getting a range of values we do a look up for each one whereas using an iterator we can get them all with one search. It will be hard to make this multi-threaded though

database/src/lib.rs
Boog900: I am ok with just leaving this as a TODO for now

database/src/backend/tests.rs
hinto-janai: Okay, will keep for now.

types/src/types.rs
```diff
@@ -15,88 +15,115 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
-    /// TODO
-    /// This is a `cuprate_consensus::HardFork`.
+    /// The block's hard-fork vote.
+    ///
+    /// This can also be represented with `cuprate_consensus::HardFork`.
+    ///
+    /// This is the same value as [`monero_serai::block::BlockHeader::minor_version`].
     pub vote: u8,
-    /// TODO
+    /// The UNIX time at which the block was recorded into the blockchain.
```
Boog900, line 39:
```suggestion
    /// The UNIX time at which the block was mined.
```

types/src/types.rs
```diff
@@ -15,88 +15,115 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
+    /// The block's height.
     pub height: u64,
-    /// TODO
+    /// The amount of generated coins (atomic units) in this block.
     pub generated_coins: u64,
-    /// TODO
+    /// The adjusted block size, in bytes.
     pub weight: usize,
-    /// TODO
+    /// The long term block weight, based on the median weight of the preceding `100_000` blocks.
```
Boog900, line 106:
```suggestion
    /// The long term block weight, which is the weight factored in with previous block weights.
```

types/src/types.rs
```diff
@@ -15,88 +15,115 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
-    /// <https://github.com/Cuprate/cuprate/pull/102#discussion_r1556694072>
-    /// <https://github.com/serai-dex/serai/blob/93be7a30674ecedfb325b6d09dc22d550d7c13f8/coins/monero/src/block.rs#L110>
-    pub block_blob: Vec<u8>,
 }
 
 //---------------------------------------------------------------------------------------------------- OutputOnChain
-/// An already approved previous transaction output.
+/// An already existing transaction output.
+///
+/// For more definitions, see also: <https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_outs>.
```
Boog900, line 115: I would rather not link to RPC docs here

database/README.md
```diff
@@ -59,86 +60,100 @@ The code within `src/` is also littered with some `grep`-able comments containin
...
+| `free.rs`              | General free functions (related to the database)
+| `key.rs`               | Abstracted database keys; `trait Key`
+| `resize.rs`            | Database resizing algorithms
+| `storable.rs`          | Data (de)serialization; `trait Storable`
+| `table.rs`             | Database table abstraction; `trait Table`
+| `tables.rs`            | All the table definitions used by `cuprate-database`
+| `tests.rs`             | Utilities for `cuprate_database` testing
+| `transaction.rs`       | Database transaction abstraction; `trait TxR{o,w}`
+| `types.rs`             | Database-specific types
+| `unsafe_unsendable.rs` | Marker type for `Send` objects not proveable to the compiler
```
Boog900, line 86: They aren't actually `Send`.

 ```suggestion
| `unsafe_unsendable.rs` | Marker type to impl `Send` for objects not `Send` 
```

Boog900 (changes requested):

Boog900 (approved):

PR #120, opened 2024-04-29
[Database] small edits to resize.rs

SyntheticBird45:
- Changed "0.0000082s" to "8.2µs" because its clearer imo
- Removed `pub` marker to free functions

### DIFF
--- a/database/src/resize.rs
+++ b/database/src/resize.rs
@@ -30,7 +30,7 @@ use std::{num::NonZeroUsize, sync::OnceLock};
 /// # TODO
 /// We could test around with different algorithms.
 /// Calling `heed::Env::resize` is surprisingly fast,
-/// around `0.0000082s` on my machine. We could probably
+/// around 8.2µs on my machine. We could probably
 /// get away with smaller and more frequent resizes.
 /// **With the caveat being we are taking a `WriteGuard` to a `RwLock`.**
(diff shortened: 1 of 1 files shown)


Boog900:
This fails to compile as the functions are private, they could be made `pub(crate)`

hinto-janai:
Is there a reason why these can't be `pub`? If they aren't necessary then the free functions should be inlined into `ResizeAlgorithm` + doc-tests should be moved to the appropriate variant:
```rust
pub enum ResizeAlgorithm {
    /// <insert [`monero`] doctest>
    Monero,

    /// <insert [`fixed_bytes`] doctest>
    FixedBytes(NonZeroUsize),

    /// <insert [`percent`] doctest>
    Percent(f32),
}

impl ResizeAlgorithm {
    // remove #[inline], too big
    pub fn resize(&self, current_size_bytes: usize) -> NonZeroUsize {
        match self {
            Self::Monero => /* inlined monero(current_size_bytes) */,
            Self::FixedBytes(add_bytes) => // etc,
            Self::Percent(f) => // etc,
        }
    }
}
```

> My bad. I didn't see heed/env.rs using fixed_bytes(). Why not replace:

Yeah that could work too.

Boog900:
@SyntheticBird45 is this this still something you want IMO these are ok being `pub`

SyntheticBird45:
lmao nah no need to resolve conflict for that

PR #126, opened 2024-05-11
database: more README docs

hinto-janai:
Expands `database/README.md` a bit more.

Closes https://github.com/Cuprate/cuprate/pull/59.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -374,25 +374,33 @@ The `CamelCase` names of the table headers documented here (e.g. `TxIds`) are th
 
 Note that words written within `code blocks` mean that it is a real type defined and usable within `cuprate_database`. Other standard types like u64 and type aliases (TxId) are written normally.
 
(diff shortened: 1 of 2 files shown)


Boog900:
@hinto-janai I think something with the windows runner changed, I got errors wit LMDB in #65 so I re-ran this one after clearing the cache and it failed as well, just looking at the error I am unsure how to fix, do you have any ideas?

database/README.md
```diff
@@ -361,4 +422,171 @@ Compatibility structs also exist for any `Storable` containers:
...
+```
+where `PreRctOutputId` looks like this:
+```rust
+struct PreRctOutputId {
+    amount: u64,
+    amount_index: u64,
+}
+```
+
+<!-- TODO(Boog900):  We also need to get the amount of outputs with a certain amount so the database will need to allow getting keys less than a key i.e. to get the number of outputs with amount `10` we would get the first key below `(10 | MAX)` and add one to `KEY_PART_2`. We also have to make sure the DB is storing these values in the correct order for this to work. -->
```
Boog900, line 511: This can probably be removed.

PR #127, opened 2024-05-12
ci: downgrade to `gcc 13.2` for windows

hinto-janai:
Will fix issue in https://github.com/Cuprate/cuprate/pull/65, https://github.com/Cuprate/cuprate/pull/126.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
   workflow_dispatch:
 
 env:
+  # Show colored output in CI.
   CARGO_TERM_COLOR: always
   # Show full panics.
   RUST_BACKTRACE: "full"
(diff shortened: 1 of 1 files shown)


hinto-janai:
@Boog900 no idea what's going on here, I'm going to setup a VM and debug there - will take a bit, sorry.

Boog900:
Alright thanks, we updated both Rust and `lmdb-master-sys` recently, might be worth rolling them back and seeing if it fixes this?

Boog900:
I think this might be because of gcc 14, it seems that recently landed in msys2 repos, we could try downgrade it

hinto-janai:
Yup, was going to say the same, not sure if msys2 repos allow downgrading though.

Is `cryptonight/` the only thing preventing us from using MSVC?

Boog900:
yeah I think so

Boog900 (approved):
Thanks for fixing this

PR #128, opened 2024-05-15
P2P: Peer sync state svc

Boog900:
This PR adds the sync state service, which keeps track of what peers claim their chain is at, so we know if we are behind. It then allows getting a list of peers IDs who claim they are ahead


p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
...
+    top_hash: [u8; 32],
+    /// The peers cumulative difficulty.
+    cumulative_difficulty: u128,
+}
+
+/// A service that keeps track of our peers blockchains.
+///
+/// This is the service that handles finding out if we need to sync and giving the peers that should
+/// be synced from to the requester.
+pub struct PeerSyncSvc<N: NetworkZone> {
```
hinto-janai, line 42: punctuation:
```suggestion
/// A service that keeps track of our peers blockchains.
///
/// This is the service that handles:
/// 1. Finding out if we need to sync
/// 1. Giving the peers that should be synced _from_, to the requester
pub struct PeerSyncSvc<N: NetworkZone> {
```

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// A map of cumulative difficulties to peers.
+    cumulative_difficulties: BTreeMap<u128, HashSet<InternalPeerID<N::Addr>>>,
+    /// A map of peers to cumulative difficulties.
+    peers: HashMap<InternalPeerID<N::Addr>, (u128, PruningSeed)>,
+    /// A watch channel for *a* top synced peer info.
+    ///
+    /// This is guaranteed to hold the sync info of a peer with the highest cumulative difficulty seen,
+    /// this makes no guarantees about which peer will be chosen in case of a tie.
+    new_height_watcher: watch::Sender<NewSyncInfo>,
```
hinto-janai, line 51: Should this doc be on the public `fn new()` instead?

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// This function checks if any peers have disconnected, removing them if they have.
+    fn poll_disconnected(&mut self, cx: &mut Context<'_>) {
+        while let Poll::Ready(Some(peer_id)) = self.closed_connections.poll_next_unpin(cx) {
+            tracing::trace!("Peer {peer_id} disconnected, removing from peers sync info service.");
+            let (peer_cum_diff, _) = self.peers.remove(&peer_id).unwrap();
+
+            let cum_dif_peers = self
+                .cumulative_difficulties
+                .get_mut(&peer_cum_diff)
+                .unwrap();
```
hinto-janai, line 91: nit: saving 1 letter with `cum_dif` is not worth it imo

https://github.com/search?q=repo%3Amonero-project%2Fmonero%20cum_dif&type=code

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
...
+            .insert(peer_id);
+
+        // If the claimed cumulative difficulty is higher than the current one in the watcher
+        // or if the peer in the watch has disconnected, update it.
+        if self.new_height_watcher.borrow().cumulative_difficulty < new_cumulative_difficulty
+            || self
+                .last_peer_in_watcher_handle
+                .as_ref()
+                .is_some_and(|handle| handle.is_closed())
+        {
```
hinto-janai, line 196: Up to you on this one, I like taking out predicates when there's multiple or are long (both here). 
```suggestion
        let claimed_cum_diff_higher =
            self.new_height_watcher.borrow().cumulative_difficulty < new_cumulative_difficulty;
        let peer_disconnected = || {
            self.last_peer_in_watcher_handle
                .as_ref()
                .is_some_and(|handle| handle.is_closed())
        };
        if claimed_cum_diff_higher || peer_disconnected() {
```

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
...
+            tracing::debug!(
+                "Updating sync watcher channel with new highest seen cumulative difficulty."
+            );
```
hinto-janai, line 199: Not sure if needed.
```suggestion
            tracing::debug!(
                "Updating sync watcher channel with new highest seen cumulative difficulty: {new_cumulative_difficulty}"
            );
```

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
...
+            let _ = self.new_height_watcher.send(NewSyncInfo {
+                top_hash: core_sync_data.top_id,
+                chain_height: core_sync_data.current_height,
+                cumulative_difficulty: new_cumulative_difficulty,
+            });
```
hinto-janai, line 204: Should this be handled? What does it mean when the `Receiver` is dropped?

pruning/src/lib.rs
```diff
@@ -126,6 +126,14 @@ impl PruningSeed {
         }
     }
 
+    /// Returns if a peer with this pruning seed should have a non-pruned version of a block.
+    pub fn has_full_block(&self, height: u64, blockchain_height: u64) -> bool {
```
hinto-janai, line 130: Is `blockchain_height` the current height? I assumed so but `CRYPTONOTE_MAX_BLOCK_HEIGHT` is passed elsewhere.
```suggestion
    /// Returns `true` if a peer with this pruning seed should have a non-pruned version of a block.
    pub fn has_full_block(&self, height: u64, blockchain_height: u64) -> bool {
```

pruning/src/lib.rs
```diff
@@ -303,6 +311,17 @@ impl DecompressedPruningSeed {
             | ((self.stripe - 1) << PRUNING_SEED_STRIPE_SHIFT)
     }
 
+    /// Returns if a peer with this pruning seed should have a non-pruned version of a block.
+    pub fn has_full_block(&self, height: u64, blockchain_height: u64) -> bool {
```
hinto-janai, line 315:
```suggestion
    /// Returns `true` if a peer with this pruning seed should have a non-pruned version of a block.
    pub fn has_full_block(&self, height: u64, blockchain_height: u64) -> bool {
```

pruning/src/lib.rs
```diff
@@ -303,6 +311,17 @@ impl DecompressedPruningSeed {
...
+        let Some(block_stripe) =
+            get_block_pruning_stripe(height, blockchain_height, self.log_stripes)
+        else {
+            return true;
+        };
+
+        self.stripe == block_stripe
```
hinto-janai, line 322:
```suggestion
        match get_block_pruning_stripe(height, blockchain_height, self.log_stripes) {
            Some(block_stripe) => self.stripe == block_stripe,
            None => true,
        }
```

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,429 @@
+//! # Sync States
+//!
+//! This module contains a [`PeerSyncSvc`] which keeps track of connected peers claimed chain states,
+//! to allow checking if we are behind and getting a list of peers who claim they are ahead.
```
hinto-janai, line 4: punctuation:
```suggestion
//! This module contains a [`PeerSyncSvc`], which keeps track of the claimed chain states of connected peers.
//! This allows checking if we are behind and getting a list of peers who claim they are ahead.
```

p2p/cuprate-p2p/src/sync_states.rs
Boog900: I am going to leave this is as, I don't think i's too bad

p2p/cuprate-p2p/src/sync_states.rs
Boog900: It means there are no handles left, which means the network has been dropped and we are shutting down.

I don't think we need to propagate this error to the caller.

pruning/src/lib.rs
Boog900: It's the height of whatever peer we are doing calculations for.

Using a value of `CRYPTONOTE_MAX_BLOCK_HEIGHT` just means we don't take into account the tip blocks which are non-pruned

p2p/cuprate-p2p/src/sync_states.rs
Boog900: ngl this was not intentional

p2p/cuprate-p2p/src/sync_states.rs
```diff
@@ -0,0 +1,427 @@
...
+                // If this was the last peer remove the whole entry for this cumulative difficulty.
+                self.cumulative_difficulties.remove(&peer_cum_diff);
+            }
+        }
+    }
+
+    /// Returns a list of peers that claim to have a higher cumulative difficulty than `current_cum_dif`.
+    fn peers_to_sync_from(
+        &self,
+        current_cum_dif: u128,
```
hinto-janai, line 101:
```suggestion
    /// Returns a list of peers that claim to have a higher cumulative difficulty than `current_cum_diff`.
    fn peers_to_sync_from(
        &self,
        current_cum_diff: u128,
```

pruning/src/lib.rs
hinto-janai: This diff was not applied btw

hinto-janai (approved):
mostly nits

Boog900:
my bad missing all those `dif`s

Issue #131, opened 2024-05-20
Workspace lints


hinto-janai:
## What
This issue is for discussing which initial [clippy](https://rust-lang.github.io/rust-clippy/master/index.html) and [rustc](https://doc.rust-lang.org/rustc/lints/index.html) lints will be added onto Cuprate's workspace.

Most discussion needed here is on the [`Lints (hot)`](#lints-hot) section.

## How
In the workspace `Cargo.toml`:
```toml
[workspace.lints.clippy]
clippy_lint = "deny"

[workspace.lints.rust]
rustc_lint = "deny"
```
Workspace crates must opt in to the lints, they can opt in to all like this:
```toml
[lints]
workspace = true
```

Crates can also inherit all workspace lints but allow some, like so:
```toml
# /crate/Cargo.toml
[lints]
workspace = true
```
```rust
// /crate/src/{lib,mod,file}.rs
#![allow(lint)]

// or

#[allow(lint)]
fn single_fn() {}
```

Anyway, this means the addition of lints can be separate PRs for each crate, adding lints to the workspace does nothing until opted into.

## Lints
Further below is an updated list of hand-picked lints and optional reasoning on why they should be used - all are set to `deny`.

There are 3 tables:
- cold: lints that are not contentious and/or fix small things like style
- hot: lints that cause more work for contributors and/or comes with tradeoffs
- warn: lints that are somewhat in the middle

## Adding
Since there are so many lints, I think the best method of adding them is:
1. Add all `cold/warm` lints to the workspace
1. Opt crates in gradually
1. See which lints cause unwanted changes
1. Remove the lint if deemed a net negative

This process would be repeated for any new lints.

`hot` lints can be discussed here before adding and/or be added on a per-crate basis.

## Removing
When a lint has too many false positives and/or has turned into a net negative, it can be discussed and removed in any PR.

## Lints (cold)

| Lint | Reasoning |
|------|-----------|
| [`borrow_as_ptr`](https://rust-lang.github.io/rust-clippy/master/index.html#borrow_as_ptr) | |
| [`case_sensitive_file_extension_comparisons`](https://rust-lang.github.io/rust-clippy/master/index.html#case_sensitive_file_extension_comparisons) | |
| [`cast_lossless`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless) | Explicitness |
[...]

PR #133, opened 2024-05-23
workspace: add lints

hinto-janai:
Adds the cold/warm/hot lints from https://github.com/Cuprate/cuprate/issues/131 into the workspace Cargo.toml.

Only the `cold/warm` lints are enabled, the `hot` ones are commented out (for now).

No crate is enabled (yet), they will be in following PRs, see: https://github.com/Cuprate/cuprate/issues/207.


Boog900 (approved):
Sorry for taking so long to review this

PR #134, opened 2024-05-23
repo: update license

hinto-janai:
- Adds a copy of the MIT license
- Updates `LICENSE.md`

We could also symlink the respective license in each crate's directory, although I think `license = "LICENSE"` may be enough.


Boog900:
For some early crates we put a license in the directory of the crate ([example](https://github.com/Cuprate/cuprate/blob/main/net/levin/LICENSE)) would you be able to remove these please

> We could also symlink the respective license in each crate's directory, although I think license = "LICENSE" may be enough.

I think AGPL requires a license with the code wherever it is, for examples if we upload it crates.io, I don't _think_ the ID is enough.

We don't have any AGPL crates at the moment but when we do we can just put an `include = [..]` and link to the license at the root of the repo, in the Cargo.toml of those crates

hinto-janai:
> if we upload it crates.io, I don't think the ID is enough.

Seems like it is, we could still link it though:

> [crates.io](https://crates.io/) requires either license or license-file to be set.
>
> https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

As per [here](https://spdx.github.io/license-list-data/) I think we would use one of these:
- `license = "AGPL-3.0-only"`
- `license = "AGPL-3.0-or-later"`
- `license = "MIT"`

Here's the list of cargo fields we need for publishing to `crates.io`: https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate.

Boog900:
> Seems like it is, we could still link it though

My concern was with the license itself not crates.io rules: 

>You may convey verbatim copies of the Program's source code as you
receive it, in any medium, provided that you conspicuously and
appropriately publish on each copy an appropriate copyright notice;
keep intact all notices stating that this License and any
non-permissive terms added in accord with section 7 apply to the code;
keep intact all notices of the absence of any warranty; **and give all
recipients a copy of this License along with the Program**.

Realistically if we don't do this we aren't going to void the license.

Boog900 (approved):

PR #137, opened 2024-05-27
repo: add issue templates

hinto-janai:
Adds template issue forms that get auto-labeled upon creation.

Reference: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository.

Examples:
- https://github.com/hinto-janai/labeler-test/issues/new/choose
- https://github.com/hinto-janai/labeler-test/issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -8,7 +8,11 @@ assignees: ''
 ---
 
 <!--
-Note: Please search to see if an issue already exists for the bug you encountered.
+
+Notes:
+- All these fields are optional, add as much or as little info as you like
+- Please search to see if an issue already exists for the bug you encountered
+
 -->
 
 ## Environment
@@ -35,5 +39,9 @@ Example:
 ## Log
 If possible, add any related logs to help explain the bug.
 
+Note: please remove any sensitive information from the logs (e.g. IP address).
+
 ## Screenshots
 If possible, add screenshots to help explain the bug.
+
+Note: please remove any sensitive information from the screenshot.
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,39 @@
+---
+name: 🐞 Bug report
+about: Create a bug report
+title: ''
+labels: ["C-bug"]
+assignees: ''
+
+---
+
+<!--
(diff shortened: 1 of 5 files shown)


SyntheticBird45:
Proposal seems to overlap feature request and discussion. People are always going to RFC when requesting new features and discussion is essentially a RFC. The only sensible difference is that Proposal could be used to downgrade/remove/edit. While discussion and feature request can't.

I think discussion is meant to be in the Github Discussion section (if we enable it). This way Proposal do overlap feature request but we can be a little bit laxist on Feature Request, as it is likely going to be use by new users, that have no idea on how to implement it

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,39 @@
...
+1. In this environment...
+2. With this config...
+3. Run '...'
+4. See error...
+
+## Log
+If possible, add any related logs to help explain the bug.
+
+## Screenshots
+If possible, add screenshots to help explain the bug.
```
Boog900, line 39: We should probably add a note here to check the logs for private info that they don't want to make public (IP addresses etc)

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,39 @@
...
+about: Create a bug report
+title: ''
+labels: ["C-bug"]
+assignees: ''
+
+---
+
+<!--
+Note: Please search to see if an issue already exists for the bug you encountered.
+-->
```
Boog900, line 12: Can we also add that people can add as much or as little info as they like, the fields are for help, not mandatory.

I don't want to scare people away from reporting bugs because they don't want to fill all these out.

.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Feature
+What is the feature you're requesting?
+
+## Why
+Why should your feature be added?
+
+## How
+If possible, describe how the proposal could be added.
+
```
Boog900, line 22: IMO this field should be removed, A person asking for a feature and then describing how to add that feature is more a proposal in my eyes

Boog900 (approved):
@SyntheticBird45 are you ok with this?

SyntheticBird45 (approved):

PR #138, opened 2024-05-28
`database/` -> `storage/`

hinto-janai:
Turns `database/` into `storage/` with the following sub-directories:
- `storage/key-value`: Pure key-value database abstraction
- `storage/cuprate-database`: Cuprate-specific database (service, types, tables, etc) built on-top of `key-value`

Also replaces the `A-database` tag with `A-storage`.


hinto-janai:
Closing in favor of https://github.com/Cuprate/cuprate/pull/139.

PR #144, opened 2024-05-30
contributing: add crate naming rules

hinto-janai:
Formalizes the naming scheme used for Cuprate's crates.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,14 +31,15 @@ After that, ensure all lints, tests, and builds are successful by running:
 - `cargo build --all-features --all-targets --workspace`
 
 ## Crate names
-Cuprate's crates (libraries) follows these naming patterns/rules:
+All of Cuprate's crates (libraries) are prefixed with `cuprate-`.
 
-| Pattern                                               | Name             | Example |
-|-------------------------------------------------------|------------------|---------|
-| Crates defining Monero related behavior               | `monero` prefix  | `monero-consensus`
-| Crates specific to Cuprate's implementation      | `cuprate` prefix | `cuprate-blockchain`
-| Monero related code re-written for Cuprate purposes   | `cuprate` suffix | `levin-cuprate`, `cryptonight-cuprate`
-| General crate, not necessarily Monero/Cuprate related | No prefix/suffix | `database`, `async-buffer`
+All directories containing crates however, are not. For example:
+
+| Crate Directory                                       | Crate Name            |
+|-------------------------------------------------------|-----------------------|
+| `storage/database`                                    | `cuprate-database`    |
+| `net/levin`                                           | `cuprate-levin`       |
+| `net/monero-wire`                                     | `cuprate-monero-wire` |
 
 ## Coding guidelines
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,11 +35,11 @@ All of Cuprate's crates (libraries) are prefixed with `cuprate-`.
 
 All directories containing crates however, are not. For example:
 
-| Crate Directory                                       | Crate Name            |
-|-------------------------------------------------------|-----------------------|
-| `storage/database`                                    | `cuprate-database`    |
-| `net/levin`                                           | `cuprate-levin`       |
-| `net/monero-wire`                                     | `cuprate-monero-wire` |
+| Crate Directory    | Crate Name         |
+|--------------------|--------------------|
+| `storage/database` | `cuprate-database` |
+| `net/levin`        | `cuprate-levin`    |
+| `net/wire`         | `cuprate-wire`     |
 
 ## Coding guidelines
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -30,6 +30,16 @@ After that, ensure all lints, tests, and builds are successful by running:
...
 - `cargo build --all-features --all-targets --workspace`
 
+## Crate names
+Cuprate's crates (libraries) follows these naming patterns/rules:
+
+| Pattern                                               | Name             | Example |
+|-------------------------------------------------------|------------------|---------|
+| Crates defining Monero related behavior               | `monero` prefix  | `monero-consensus`
+| Crates specifically for Cuprate's implementation      | `cuprate` prefix | `cuprate-blockchain`
+| Monero related code re-written for Cuprate purposes   | `cuprate` suffix | `levin-cuprate`, `cryptonight-cuprate`
```
hinto-janai, line 40: Is there a better description for the difference between the 1st and 3rd rule?

I.e. what does the `-cuprate` suffix mean?

SyntheticBird45: 1st: Crates defining Monero intrinsic behavior that is software-agnostic (can be used outside of Cuprate).
3rd: General purpose crates re-implementing monero-related code for Cuprate specifically (usually by redefining API).

SyntheticBird45: Also:
```suggestion
| Crates defining Monero related behavior               | `monero` prefix  | `monero-consensus`
| Crates specific to Cuprate's implementation      | `cuprate` prefix | `cuprate-blockchain`
| Monero related code re-written for Cuprate purposes   | `cuprate` suffix | `levin-cuprate`, `cryptonight-cuprate`
```

SyntheticBird45:
I'm a little lost on the General crates. `database` is meant to be used by Cuprate and cuprate only, I don't see how it isn't part of its implementation

CONTRIBUTING.md
Boog900: > I.e. what does the -cuprate suffix mean?

I have used it when a better alternative isn't available, i.e. `-cuprate` pretty much means the same as no prefix.

btw I changed monero-consensus to cuprate-consensus-rules in #65 could probably change it to monero-consensus-rules so it follows this spec though

hinto-janai: > 1st: Crates defining Monero intrinsic behavior that is software-agnostic (can be used outside of Cuprate).

Is this true currently? Can all `monero-` crates be used without pulling `cuprate-` dependencies?

> I have used it when a better alternative isn't available, i.e. -cuprate pretty much means the same as no prefix.

Open to suggestions, we can change the spec to anything (I just want it formalized).

hinto-janai:
> I'm a little lost on the General crates. database is meant to be used by Cuprate and cuprate only

The names changed recently: `database` is like `HashMap`, e.g. anyone that needs a database abstraction could use `database`.

`cuprate-blockchain` (previously `cuprate-database`) is related since it implements the blockchain-specific stuff for Cuprate on top of the agnostic `database`.

CONTRIBUTING.md
hinto-janai: A simpler option is to remove prefix/suffixes from directory names and always prefix `cuprate-` to all crate names, like [`reth`](https://github.com/paradigmxyz/reth/tree/main/crates).

E.g.
- `database/` -> `cuprate-database`
-  `levin/` -> `cuprate-levin`
- `epee-encoding/` -> `cuprate-epee-encoding`
- `monero-wire/` -> `cuprate-wire` or `cuprate-monero-wire`

This is clearer if/when uploading to crates.io as well. I think I'd prefer this way actually.

hinto-janai:
Pushed changes for https://github.com/Cuprate/cuprate/pull/144#discussion_r1638344193.

If that naming scheme is chosen, I'll follow up with a PR that changes all the current crates.

SyntheticBird45 (approved):

CONTRIBUTING.md
Boog900: This would be nice and simple, however we have:

`monero-p2p` a lower level p2p crate

`cuprate-p2p` a higher level crate built off `monero-p2p`

I don't really know what we would change `monero-p2p` to.

hinto-janai: `cuprate-monero-p2p` or `cuprate-p2p-core`? Examples of `-core` suffix representing lower-level crates:
- https://docs.rs/rayon_core
- https://docs.rs/rand_core
- https://docs.rs/parking_lot_core
- https://docs.rs/tracing_core
- https://docs.rs/futures_core
- https://docs.rs/axum_core

Boog900: yeah `cuprate-p2p-core` is a good idea

CONTRIBUTING.md
```diff
@@ -30,6 +30,17 @@ After that, ensure all lints, tests, and builds are successful by running:
...
+All of Cuprate's crates (libraries) are prefixed with `cuprate-`.
+
+All directories containing crates however, are not. For example:
+
+| Crate Directory                                       | Crate Name            |
+|-------------------------------------------------------|-----------------------|
+| `storage/database`                                    | `cuprate-database`    |
+| `net/levin`                                           | `cuprate-levin`       |
+| `net/monero-wire`                                     | `cuprate-monero-wire` |
+
```
Boog900, line 43:
```suggestion
| `net/monero-wire`                                     | `cuprate-wire` |

```

Boog900 (approved):

Issue #106, opened 2024-04-08
RPC design proposal


yamabiiko:
This issue is meant to initiate a discussion on the high-level design of Cuprate's daemon RPC.
The main focus is on defining the interface and  handling parsing methods/params and responses, so I abstracted away all interactions with other Cuprate's parts for now and the actual daemon.

## Overview

[Monerod's daemon RPC](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html)  has three kinds of RPC calls: 

- JSON RPC methods, called at the endpoint `/json_rpc`
- general RPC methods called at their own endpoints (eg. `/get_height`)
- binary (epee format) RPC methods called at their own endpoints ending in .bin (eg.`/get_blocks.bin`)

Given the above it seems natural to split the daemon into three different Services / Layers.

![cuprate](https://github.com/Cuprate/cuprate/assets/30393429/e6204778-a16a-4beb-96ca-b5c09f623dbf)

## (Core) RPC Service
This is the actual service responsible for calling the methods (and interfacing with the necessary Cuprate crates such as the db), it implements `Service<Rpc>` and returns a `method::Response`. 
enum/struct example:
```rust
pub enum Rpc {
    TestMethod(crate::param::TestMethod),
    ...
}
// crate::param
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
pub struct TestMethod {
    number: u8,
}

epee_object!(
    TestMethod,
    number: u8,
);
```
```rust
//method::Response
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
struct BaseResponse {
    credits: u64,
    status: String,
    top_hash: String,
    untrusted: bool,
}

epee_object!(
    BaseResponse,
    credits: u64,
    status: String,
    top_hash: String,
    untrusted: bool,
);

#[derive(Clone, Serialize, Deserialize)]
pub enum Response {
    TestMethod(TestResponse)
}

#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
pub struct TestResponse {
    base: BaseResponse,
    pub test: u8
}

epee_object!(
    TestResponse,
    base: BaseResponse,
    test: u8,
);
```
## Middleware Service
This service is mainly responsible for parsing the parameters of the RPC call so that we can construct the struct related to the specific RPC method and deserializing the response into the proper type (`Value`, `JsonResponse`, `Bytes` in epee format).
Errors regarding method name and parameters are caught and handled before calling the core service.
[...]

PR #146, opened 2024-05-31
rpc: add initial crates

hinto-janai:
## What
This proposal builds upon the previous discussion and sets the initial rough design for Cuprate's RPC libraries, for the purpose of acting as a guide for implementation, although things will change as it is implemented.

This PR only adds the crates in `# Where`.

Closes https://github.com/Cuprate/cuprate/issues/106.

## Where
```bash
# Repo root.
rpc/
│
├─ json-rpc/
│  └── [...]
│
├─ monero-rpc-types/
│  └── [...]
│
├─ cuprate-rpc-interface/
│  └── [...]
```

| Crate | Purpose | Monero (rough) equivalent |
|-------|---------|---------------------------|
| `json-rpc` | JSON-RPC 2.0 library | [`rpc/message.{cpp,h}`](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/message.cpp) |
| `monero-rpc-types` | (De)serializable Monero RPC type definitions, e.g. `GetBlock`. | [`rpc/daemon_messages.{cpp,h}`](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/daemon_messages.h)
| `cuprate-rpc-interface` | RPC interface library, uses `monero-rpc-types`, exposes a barebones `axum::Router`, configuration, types, etc. Internally implements routes that handle all pre-handling operations like `restricted`, limits, etc. Defines the `tower::Service` request/response types used by users. Expects users to bring their own handler and middleware. | [`rpc/core_rpc_server*.{cpp,h}`](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server.h) |

### `json-rpc`
Rough code:
```rust
// `method/params` -> `body` reasoning:
// <https://github.com/Cuprate/cuprate/pull/146#issuecomment-2148576495>
#[derive(Deserialize, Serialize)]
struct Request<M /* method */> {
    jsonrpc: Version,
    id: Id,
    #[serde(flatten)]
    body: M,
}

/* JSON-RPC 2.0 impl */
```

### `monero-rpc-types`
Rough code:
```rust
#[derive(EpeeAndOrSerde)]
pub struct GetBlock {/*...*/}

#[derive(EpeeAndOrSerde)]
pub struct GetInfo {/*...*/}

/* all RPC types */
```

### `cuprate-rpc-interface`
Rough code:
```rust
// State the RPC router has access to.
pub struct RpcState<S, H> {/*...*/}

// Public function that creates _the_ router.
// Sets all the routes Monero has.
pub fn create_router<S, H>() -> Router<RpcState<S, H>>
where
    H: tower::Service<crate::types::Request>,
    H: <H as Service>::Response: Into<crate::types::Response>,
    H: <H as Service>::Error: Into<crate::types::RpcError>,
{
    Router::new()
        .route("/json_rpc",   get(crate::route::json_rpc))
        .route("/*bin",       get(crate::route::binary))
        .route("/get_height", get(crate::route::get_height))
        /* all routes */
}

// Internal route functions.
mod route {
    async fn json_rpc(/*...*/)   {/*...*/}
    async fn binary(/*...*/)     {/*...*/}
    async fn get_height(/*...*/) {/*...*/}
    /* all routes */
}

// Types derive (de)serialization where needed.
mod types {
    // Could be 1 enum instead.
    pub enum JsonRequestEnum  {/*...*/}
    pub enum BinRequestEnum   {/*...*/}
    pub enum OtherRequestEnum {/*...*/}

    pub enum Request {
        Json(JsonRequestEnum),
        Bin(BinRequestEnum),
	Other(OtherRequestEnum),
    }

    pub enum Response {
        Json(JsonResponseEnum),
        Bin(BinResponseEnum),
        Other(OtherResponseEnum),
    }

    // errors that can occur during RPC.
    pub enum RpcError {/*...*/}
}

// Other stuff.
mod config    {/*...*/}
mod constants {/*...*/}
mod error     {/*...*/}
/*...*/
```


Boog900:
Looks good, just some small things: 

> cuprate-rpc-server

I don't think this crate should exist - I think it should just be in the `cuprated` binary.

> cuprate-rpc-interface

How are you planning to handle `json-rpc` types, they all have a common header, would it be something like this

```rust
async fn json_rpc(Json(req): Json(CommonHeader<JsonRequestEnum>)))   {/*...*/} 
```

If so, I think this is a pretty reason to not join the 3 enums like I suggested.

hinto-janai:
> I don't think this crate should exist - I think it should just be in the cuprated binary.

I was thinking `cuprated` would have as little logic as possible, basically being:
```rust
use /* all other cuprate crates */;

fn main() -> Result<(), Error> {
    // 1. read disk
    // 2. init everything
    // 3. hang
    // 4. exit gracefully
}
```
and all functionality would be outside for maintainability/compile times/etc, maybe aside from CLI parsing.

> How are you planning to handle json-rpc types, they all have a common header

Oops right forgot, we would implement a JSON-RPC 2.0 library in-tree (`rpc/json-rpc`?) similar to Monero's `rpc/message.{cpp,h}`.
```rust
pub enum JsonRpcMethod {/*...*/}

pub enum Request {
    // using something similar to:
    // <https://github.com/hinto-janai/json-rpc/blob/main/src/request.rs#L11>
    //
    //                  method      (optional) param
    //                     v              v
    Json(json_rpc::Request<JsonRpcMethod, SomeStrType>),
    Bin(BinRequestEnum),
    Other(OtherRequestEnum),
}
```

Boog900:
> I was thinking cuprated would have as little logic as possible, basically being:

I do want `cuprated` to be pretty slim but I think the the RPC request handler should be in there. 

The inner request handler will need access to pretty much all major Cuprate services (database/ txpool / syncer) and currently I am planning for 2 of them to be in `cuprated`. Although I could move them to separate crates, they aren't really reusable and I don't think I agree it would help maintenance.

We would still have separate modules in `cuprated`.

Plus `cuprated` is the only AGPL part of the Cuprate node, so it makes sense to have a bit more than just bare bones CLI/ config handling, although I admit this is a pretty weak point. 

> Oops right forgot, we would implement a JSON-RPC 2.0 library in-tree (rpc/json-rpc?) similar to Monero's rpc/message.{cpp,h}.

>     Json(json_rpc::Request<JsonRpcMethod, SomeStrType>),

With this we would need to ser/de twice, once for method and a second time for params, I _think_ we could make better use of serde's enum handling to only need to ser/de once: https://serde.rs/enum-representations.html#adjacently-tagged

hinto-janai:
> I do want cuprated to be pretty slim but I think the the RPC request handler should be in there.

OK, will edit crates to:
- monero-rpc-types
- cuprate-rpc-interface
- json-rpc (maybe?)

> With this we would need to ser/de twice, once for method and a second time for params

When/where does this happen?

In the following `axum` handler:
```rust
async fn json_rpc(Json(req): Json(json_rpc::Request<JsonRpcMethod, SomeStrType>))) {/*...*/}
```
doesn't `req` get  deserialized once as a `Request`?

Then when sending:
```rust
let json_rpc = json_rpc::Request {/*...*/}; // not serialized yet
let request = Request::Json(json_rpc);

let bytes = serialize(request);
```

Boog900:
> When/where does this happen?

The params are a string type so when the request is deserialized you still have the params as a string that need to be  deserialized.

 > doesn't req get deserialized once as a Request?
 
 Yes but the params are a string so if you need them you need to serde a second time.
 
 > Then when sending:
 
```rust

let serialized_params = serialize(params);  // you need to serialize the params before making the request object.
let json_rpc = json_rpc::Request { serialized_params,  /*...*/};
let request = Request::Json(json_rpc);

let bytes = serialize(request);
```

hinto-janai:
What about something like:
```rust
async fn json_rpc(Json(req): Json(json_rpc::Request<JsonRpcMethod, serde_json::Value>))) {/*...*/}
```
Sending (although practically, `serde_json::json!()` would be used?):
```rust
let method = JsonRpcMethod::GetBlock;
let map = Map::from_iter([(
    "height".to_string()
    Value::Number(Number::from(123_u64))
)]);
let params   = Value::Object(map);
let json_rpc = json_rpc::Request { method, params, /*...*/ }; // not serialized yet
let request  = Request::Json(json_rpc);
let bytes    = serialize(request);
```
If we wanted to be strict/type-safe, we could define which params are valid for each method: https://github.com/hinto-janai/festival/blob/8d7e26d55e7677e3d38d6c51c51683830e9cbd61/rpc/src/param.rs#L291-L297.

Although, this may be a behavior change. I _think_ JSON-RPC 2.0 is supposed to error if garbage params are inputted:
> [by-name: params MUST be an Object, with member names that match the Server expected parameter names](https://www.jsonrpc.org/specification#parameter_structures)

although, `monerod` ignores garbage params and responds anyway.

A non-valid param would error in deserialization for us. Although this would allow us to create 1-to-1 param types for each method that only contains what it needs instead of being a generic (and expensive) `serde_json::Value`.

Boog900:
> Although, this may be a behavior change. I think JSON-RPC 2.0 is supposed to error if garbage params are inputted:
> > [by-name: params MUST be an Object, with member names that match the Server expected parameter names](https://www.jsonrpc.org/specification#parameter_structures)

I may be wrong but I interpret that as meaning: `if params are needed, they should be provided`, not that extra param members should cause errors, it explicitly says: `The absence of expected names MAY result in an error being generated`

> What about something like:

That would still be pretty expensive, wouldn't something like this work:
```rust
#[derive(Deserialize, Serialize)]
struct GetBlock {
    /*...*/
}

#[derive(Deserialize, Serialize)]
#[serde(tag = "method", content = "params")]
#[serde(rename_all = "snake_case")]
enum Methods {
    GetBlock(GetBlock),
    /* other methods */
}

#[derive(Deserialize, Serialize)]
struct Request {
    id: u64, // These values aren't the right types, just for example 
    jsonrpc: u64,
    #[serde(flatten)]
    body: Methods, // Could make this type generic.
}

```

SyntheticBird45:
> That would still be pretty expensive, wouldn't something like this work:

That would fit nicely and I think it's the correct approach.  But I also agree this is expensive cpu and certainly memory wise. 
If we take this approach we will need to edit it later because under high-load its going to be an problematic, compared to monerod

Boog900:
> That would fit nicely and I think it's the correct approach. But I also agree this is expensive cpu and certainly memory wise.
If we take this approach we will need to edit it later because under high-load its going to be problematic, compared to monerod

Is this about my method or hintos? 

I am unsure about the internals of serde but I don't _think_ you would be able to get much better than my method without a custom json lib (not using serde).

SyntheticBird45:
> Is this about my method or hintos?

Yours

> I am unsure about the internals of serde but I don't think you would be able to get much better than my method without a custom json lib (not using serde).

I was honestly going to talk about it later because its a little bold to ask: Do we really want to rely on serde ? Can't we explore other crates in the ecosystem ? I'm sure there must be more correct and performant alternatives

Boog900:
> I was honestly going to talk about it later because its a little bold to ask: Do we really want to rely on serde ? Can't we explore other crates in the ecosystem ? I'm sure there must be more correct and performant alternatives

We could, do you know of any in particular? 

However, I don't think serde-json is incorrect, I don't know what you mean by `more correct`. 

We would have to bench this but I don't think serde-json is slow: https://github.com/serde-rs/json#performance

SyntheticBird45:
> > I was honestly going to talk about it later because its a little bold to ask: Do we really want to rely on serde ? Can't we explore other crates in the ecosystem ? I'm sure there must be more performant alternatives
> 
> We could, do you know of any in particular?
> 
> We would have to bench this but I don't think serde-json is slow: https://github.com/serde-rs/json#performance

https://github.com/cloudwego/sonic-rs to keep in mind. Benchmarking will probably need to be done to attest the performance improvements

hinto-janai:
For future readers: the `json-rpc` library is planned to use this format: https://github.com/Cuprate/cuprate/pull/146#issuecomment-2145734838.

Reasoning from [meeting logs](https://github.com/monero-project/meta/issues/1017):
```
hinto: boog900: re: https://github.com/Cuprate/cuprate/pull/146#issuecomment-2145734838 -
is there a reason it should be done like this instead of a method and params field?

hinto: e.g. like so: https://docs.rs/jsonrpsee-types/0.22.5/jsonrpsee_types/request/struct.Request.html

boog900: so serde knows what the expected params are. If we keep them separate then we either
have to serde in 2 steps or use serde(untagged) which isn't great for performance 
```

Boog900 (approved):

PR #147, opened 2024-06-03
rpc: start `cuprate-rpc-types`

hinto-janai:
Begins implementing `cuprate-rpc-types`.

- Creates macros needed for mass type generation
- Ports over some RPC types (JSON only for now)
- Adds some misc RPC types (e.g. `enum Status`)
- Adds the header/base RPC types (e.g. `struct ResponseBase`)

The main point of this PR is to implement the type generator macro, then following PR(s) will use that macro to port all `{json,bin,other}` definitions.


rpc/monero-rpc-types/src/lib.rs
```diff
@@ -1 +1,198 @@
...
+        GetBlockTemplate,
+    }
+
+    bin {
+    }
+
+    other {
+        SaveBc,
+    }
+}
```
hinto-janai, line 198: The way the types/modules are setup are as follows:

- There's 3 actual files, `{bin,json,other}.rs`, which contain all the types with either a `Response/Request` suffix, e.g. `GetBlockCount{Request,Response}`
- This macro re-exports/re-routes/re-names those types to the appropriate module  e.g. `{req,resp}::json::GetBlockCount`

I'm not sure if there's a cleaner way to do this but I think modules are nicer than a `Response/Request` suffix on every type.

rpc/monero-rpc-types/src/json.rs
```diff
@@ -0,0 +1,81 @@
...
+        // $ATTRIBUTES
+        // $FIELD_NAME: $FIELD_TYPE,
+        // ```
+        // The struct generated and all fields are `pub`.
+
+        count: u64,
+        status: crate::Status,
+        untrusted: bool,
+    }
+}
```
hinto-janai, line 38: This is the macro that defines an RPC request/response type (in 1 go), currently only supports `serde` not binary.

This creates `crate::json::GetBlockCount{Request,Response}` (which gets re-exported using the macro described before).

The macro itself currently derives common things + serde but I think I'll have to change it when some types end up not compatible...

rpc/monero-rpc-types/src/json.rs
```diff
@@ -0,0 +1,81 @@
...
+    core_rpc_server_commands_defs.h => 935..=939,
+    OnGetBlockHash,
+    #[derive(Copy)]
+    Request {
+        block_height: u64,
+    },
+    Response {
+        block_hash: String,
+    }
+}
```
hinto-janai, line 51: Another smaller example, this defines `crate::json::OnGetBlockHash{Request,Response}` which gets re-exported as `crate::{req,resp}::json::GetBlockCount`.

hinto-janai (commented):
@Boog900 requesting review on the module/type set-up.

rpc/monero-rpc-types/src/lib.rs
Boog900: People are often going to need the complimentary response and request type(s), which would have the same name so I don't know if this is a good idea, I think I would rather them just keep the suffix.

rpc/monero-rpc-types/src/macros.rs
```diff
@@ -0,0 +1,123 @@
...
+            ".",
+            stringify!($monero_code_filename_extension),
+            " @ ",
+            stringify!($monero_code_line_start),
+            "..=",
+            stringify!($monero_code_line_end),
+            "`](",
+            "https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/",
+            stringify!($monero_code_filename),
+            ".",
```
Boog900, line 62: I feel very uneasy about linking to monerod in our code docs.

At the very least I think the commit they point to should be per type, otherwise updating this will be a pain, you would have to update all of them even if only one has changed.

rpc/monero-rpc-types/README.md
```diff
@@ -0,0 +1,61 @@
...
+    - `other` (Misc JSON types from other endpoints)
+1. Each type in `req` has a corresponding type in `resp` and vice-versa with an identical name, e.g. [`req::json::GetBlockCount`] and [`resp::json::GetBlockCount`]
+
+# Documentation
+The documentation for types within [`req`] and [`resp`] are omitted,
+as they can be found in [Monero's RPC documentation](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#on_get_block_hash).
+
+However, each type will document:
+- The exact type definition location in `monerod`
+- The Monero RPC documentation link
```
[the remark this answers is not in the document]
Boog900: IMHO I would rather the types juts be laid out like the files currently are, I can't think of many situations when you would only need a request/ response

rpc/monero-rpc-types/src/lib.rs
hinto-janai: Okay yeah now that I think about it the nested modules are probably not practical (pretty though).

How is this?:
```rust
use json_rpc::json::{GetBlockCountResponse, GetBlockCountRequest};
```
`req/resp` modules are removed, the top root modules are `{json,bin,other}`.

rpc/monero-rpc-types/src/macros.rs
hinto-janai: > At the very least I think the commit they point to should be per type

Will do this.

Some of our types will probably differ in name (`GetBlocksFast` -> `crate::bin::GetBlocks{Request,Response}`) so I think having some sort of definitive link connecting the two would be helpful and worth maintaining.

rpc/monero-rpc-types/src/lib.rs
Boog900: Yeah that looks good

rpc/types/src/macros.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Really, they semantically are empty types, so `()` is used.
+///
+/// Again, other than this, the 2 branches do (should) not differ.
+///
+/// FIXME: there's probably a less painful way to branch here on input
+/// without having to duplicate 80% of the macro. Sub-macros were attempted
+/// but they ended up unreadable. So for now, make sure to fix the other
+/// branch as well when making changes. The only de-duplicated part is
+/// the doc generation with [`define_request_and_response_doc`].
+macro_rules! define_request_and_response {
```
hinto-janai, line 41: There's a lot of `Request` types that have no inputs; generating a bunch of incompatible `struct` types that are actually all the same type (`()`) would be unwieldy for callers, e.g. the interface code would have to reference specific types even though they are all `()`.

So, with this input: `Request {}`, the macro now generates this:

image

Now callers can just use `()`. A downside is that the macro has some duplicate code within the 2 branches, I don't think it's too bad though.

Cargo.toml
```diff
@@ -75,6 +75,7 @@ rayon                 = { version = "1.9.0", default-features = false }
 serde_bytes           = { version = "0.11.12", default-features = false }
 serde_json            = { version = "1.0.108", default-features = false }
 serde                 = { version = "1.0.190", default-features = false }
+strum                 = { version = "0.26.2", default-features = false }
```
hinto-janai, line 78: Used by `crate::Status` (and probably other stuff in the future).

Not strictly necessary but it provides macros for things we would otherwise implement manually (`as_ref()`, custom serialization, etc), alongside other stuff.

Boog900: I'm not completely against this if you think it's worth it, however at the moment it's only used on one type

rpc/types/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name        = "cuprate-rpc-types"
+version     = "0.0.0"
+edition     = "2021"
+description = "Monero RPC types"
+license     = "MIT"
+authors     = ["hinto-janai"]
+repository  = "https://github.com/Cuprate/cuprate/tree/main/rpc/cuprate-rpc-types"
```
Boog900, line 8:
```suggestion
repository  = "https://github.com/Cuprate/cuprate/tree/main/rpc/types"
```

rpc/types/README.md
```diff
@@ -0,0 +1,60 @@
...
+Note that some types within [`other`] mix JSON & binary together, i.e.,
+the message overall is JSON, however some fields contain binary
+values, for example:
+
+```json
+{
+  "string": "",
+  "float": 30.0,
+  "integer": 30,
+  "binary": /* serialized binary */
```
Boog900, line 56: Is it decided on how we are going to handle this yet?

Boog900: I know I mentioned using a JSON5 serde lib but I don't really want to have to pull in a lib like that in for this.

rpc/types/src/macros.rs
Boog900: Is this macro going to be generaing the binary types as well? 

If so you will need to add `epee_encoding`

Boog900 (changes requested):

rpc/types/README.md
hinto-janai: AFAICT `monerod` puts binary inside JSON strings, e.g. this is the response from `get_txpool_backlog`:
```json
{
  "id": 0,
  "jsonrpc": "2.0",
  "result": {
    "backlog": "�\b����������", // note the quotes, it's a JSON string
    "credits": 0,
    "status": "OK",
    "top_hash": "",
    "untrusted": false
  }
}
```
So I think we just model it as a `String` Or maybe some custom wrapper around `Vec<u8>` so we don't pay UTF-8 costs? The following code works:
```rust
use serde::*;
use serde_json::*;

const JSON: &str = r#"{
  "id": 0,
  "jsonrpc": "2.0",
  "result": {
    "backlog": "�\b����������",
    "credits": 0,
    "status": "OK",
    "top_hash": "",
    "untrusted": false
  }
}"#;

#[derive(Serialize, Deserialize)]
struct GetTxpoolBacklog {
    id: u64,
    jsonrpc: String,
    result: Result,
}

#[derive(Serialize, Deserialize)]
struct Result {
    backlog: String, // a JSON string holding binary data
    status: String,
    untrusted: bool,
}

#[derive(EpeeStuff)]
struct Backlog {/*...*/}

fn main() {
    let s = from_str::<GetTxpoolBacklog>(JSON).unwrap(); // this works

    let epee = epee_decode::<Backlog>(&s.result.backlog.as_bytes());
}
```

Boog900: @SyntheticBird45 mentioned that there were some methods that return the raw binary in a field, not in a string.

I just had a look and it seems that the ones they mentioned do return it in a string unless I am missing something @SyntheticBird45?

Cargo.toml
hinto-janai: Added a [TODO](https://github.com/Cuprate/cuprate/pull/147/commits/4830a261807ea149760039897ec36ac95aaeadc5), I'll remove it if it isn't used elsewhere by the end of RPC impl.

hinto-janai: Actually, removed this for now. Will add if needed.

Cargo.toml
```diff
@@ -23,8 +23,8 @@ members = [
 	"test-utils",
 	"types",
 	"rpc/json-rpc",
-	"rpc/rpc-types",
-	"rpc/rpc-interface",
+	"rpc/types",
+	"rpc/interface",
```
hinto-janai, line 27: Sort of a grey area leftover from https://github.com/Cuprate/cuprate/pull/164.

The crate is `cuprate-rpc-types`, but the directory is `rpc/types`. Should it technically be `rpc/rpc-types`?

Maybe there's leeway here if the directory already has the name, e.g. `cuprate-consensus-rules` is `consensus/rules`.

rpc/types/src/binary_string.rs
```diff
@@ -0,0 +1,29 @@
...
+/// struct Key {
+///     key: BinaryString,
+/// }
+///
+/// let binary = r"�\b����������";
+/// let json = format!("{{\"key\":\"{binary}\"}}");
+/// let key = from_str::<Key>(&json).unwrap();
+/// let binary: BinaryString = key.key;
+/// ```
+pub type BinaryString = String;
```
hinto-janai, line 23: This will be used in fields where we're expecting binary inside a `String`.

rpc/types/src/base.rs
```diff
@@ -0,0 +1,125 @@
...
+//! These are the common "headers" or "base" types that are
+//! [`flattened`](https://serde.rs/field-attrs.html#flatten)
+//! into many of Monero's RPC types.
+//!
+//! The `Access*` structs (e.g. [`AccessResponseBase`]
+//! are pseudo-deprecated structs for the RPC payment system, see:
+//!
+//! - <https://github.com/monero-project/monero/commit/2899379791b7542e4eb920b5d9d58cf232806937>
+//! - <https://github.com/monero-project/monero/issues/8722>
+//! - <https://github.com/monero-project/monero/pull/8843>
```
hinto-janai, line 12: This module defines the base types that are flattened into almost every response.

The most common being the `{ status: String, untrusted: bool }` one.

I implemented the RPC payment versions for now.

rpc/types/src/json.rs
```diff
@@ -0,0 +1,127 @@
...
+        reserved_offset: u64,
+        expected_reward: u64,
+        prev_hash: String,
+        seed_height: u64,
+        seed_hash: String,
+        next_seed_hash: String,
+        blocktemplate_blob: String,
+        blockhashing_blob: String,
+    }
+}
```
hinto-janai, line 90: ## Type generator macro 1/3

Basically the same as before but instead of `Request {}` and `Response {}`, you specify one of the bases mentioned before. The macro picks up on this, and flattens that into your generated struct.

E.g. here it is `ResponseBase {}` so `{ status: Status, untrusted: bool }` will automatically get added to the generated `GetBlockTemplateResponse`.

rpc/types/src/json.rs
```diff
@@ -0,0 +1,127 @@
...
+    GetBlockCount,
+
+    // There is no request type specified,
+    // this will cause the macro to generate a
+    // type alias to `()` instead of a `struct`.
+
+    ResponseBase {
+        count: u64,
+    }
+}
```
hinto-janai, line 105: ## Type generator macro 2/3

Now you just don't specify a `Request`, and it'll generate a `type GetBlockCountRequest = ()`.

rpc/types/src/macros.rs
```diff
@@ -0,0 +1,277 @@
...
+            )*
+        }
+
+        ::cuprate_epee_encoding::epee_object! {
+            [<$type_name Response>],
+            $(
+                $response_field: $response_field_type,
+            )*
+            !flatten: base: $response_base_type,
+        }
```
hinto-janai, line 223: ## Type generator macro 3/3

Every request/response type now implements `EpeeObject`. Does this look correct?

rpc/types/src/status.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// This exists to act as a catch-all if `monerod` adds
+    /// a string and a Cuprate node hasn't updated yet.
+    ///
+    /// The reason this isn't `Unknown(String)` is because that
+    /// disallows [`Status`] to be [`Copy`], and thus other types
+    /// that contain it.
+    #[serde(other)]
+    #[serde(rename = "UNKNOWN")]
+    Unknown,
+}
```
hinto-janai, line 97: ## Status 1/3
I was actually re-thinking doing this `String` optimization stuff, but there is no downside, right?

My main concern was that if `monerod` ever adds a new string in the `status` field, this would fail to deserialize, but, `#[serde(other)]` and...

rpc/types/src/status.rs
```diff
@@ -0,0 +1,182 @@
...
+//
+// See below for more impl info:
+// <https://github.com/Cuprate/cuprate/blob/bef2a2cbd4e1194991751d1fbc96603cba8c7a51/net/epee-encoding/src/value.rs#L366-L392>.
+impl EpeeValue for Status {
+    const MARKER: Marker = <String as EpeeValue>::MARKER;
+
+    fn read<B: Buf>(r: &mut B, marker: &Marker) -> cuprate_epee_encoding::Result<Self> {
+        let string = <String as EpeeValue>::read(r, marker)?;
+        Ok(Self::from(string))
+    }
```
hinto-janai, line 141: ## Status 2/3
...this Epee implementation and...

rpc/types/src/status.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn from(s: String) -> Self {
+        match s.as_str() {
+            CORE_RPC_STATUS_OK => Self::Ok,
+            CORE_RPC_STATUS_BUSY => Self::Busy,
+            CORE_RPC_STATUS_NOT_MINING => Self::NotMining,
+            CORE_RPC_STATUS_PAYMENT_REQUIRED => Self::PaymentRequired,
+            _ => Self::Unknown,
+        }
+    }
+}
```
hinto-janai, line 109: ## Status 3/3
...this `From<String>` impl means that `Status` can never fail to deserialize from any string, right?

rpc/types/src/status.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn should_write(&self) -> bool {
+        true
+    }
+
+    fn epee_default_value() -> Option<Self> {
+        // TODO: what is the default here?
+        Some(Self::default())
+    }
```
hinto-janai, line 150: Not sure what to put here. An empty string?

rpc/types/README.md
hinto-janai: https://github.com/Cuprate/cuprate/pull/147#discussion_r1653774069

rpc/types/src/macros.rs
Boog900: yeah it does

rpc/types/src/status.rs
Boog900: Yeah I think it should be fine, I don't think there are any downsides.

rpc/types/src/status.rs
Boog900: I would put this as `Unknown` as that's what an empty string should deserialize to

rpc/rpc-types/src/lib.rs
```diff
@@ -1 +0,0 @@
-
```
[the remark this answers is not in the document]
hinto-janai: The diff is `rpc/rpc-types/src/lib.rs` (empty file) -> `rpc/types/src/lib.rs` (non-empty file).

It doesn't show it as a move since stuff has been added I guess.

rpc/types/src/json.rs
```diff
@@ -0,0 +1,127 @@
...
+        difficulty_top64: u64,
+        height: u64,
+        reserved_offset: u64,
+        expected_reward: u64,
+        prev_hash: String,
+        seed_height: u64,
+        seed_hash: String,
+        next_seed_hash: String,
+        blocktemplate_blob: String,
+        blockhashing_blob: String,
```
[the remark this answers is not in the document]
hinto-janai: Yup was thinking about array strings and/or other optimizations too, although I'm worried about stack overflows. Will look into this / do in the later PR that generates all the types.

Btw `monerod` uses heap strings too https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L964-L975.

rpc/types/src/binary_string.rs
hinto-janai: Will do in another PR.

Boog900 (approved):

PR #148, opened 2024-06-04
rpc: implement `json-rpc` crate

hinto-janai:
Implements the `json-rpc` crate, a [JSON-RPC 2.0](https://www.jsonrpc.org) implementation, modified for usage in Cuprate.

- Originally based off: https://github.com/hinto-janai/json-rpc
- Maintains some changes made in https://github.com/Cuprate/cuprate/pull/43
- Adds `method/params` -> `body` changes from https://github.com/Cuprate/cuprate/pull/146#issuecomment-2148576495


rpc/json-rpc/README.md
```diff
@@ -0,0 +1,76 @@
...
+r#"{
+  "jsonrpc": "2.0",
+  "id": "hello",
+  "method": "get_block",
+  "params": {
+    "height": 123
+  }
+}"#;
+assert_eq!(json, expected_json);
+```
```
hinto-janai, line 72: Usage example.

rpc/json-rpc/src/error/code.rs
```diff
@@ -0,0 +1,224 @@
...
+/// found in [`ErrorObject`](crate::error::ErrorObject).
+///
+/// It associates the code integer ([`i32`]) with its defined message.
+///
+/// # Application defined errors
+/// The custom error codes past `-32099` (`-31000, -31001`, ...)
+/// defined in JSON-RPC 2.0 are not supported by this enum because:
+///
+/// 1. The `(i32, &'static str)` required makes the enum more than 3x larger
+/// 2. It is not used by Cuprate/Monero
```
hinto-janai, line 23: I don't think we'll use custom errors, right?

rpc/json-rpc/src/error/code.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        match code {
+            PARSE_ERROR_I32 => Self::ParseError,
+            INVALID_REQUEST_I32 => Self::InvalidRequest,
+            METHOD_NOT_FOUND_I32 => Self::MethodNotFound,
+            INVALID_PARAMS_I32 => Self::InvalidParams,
+            INTERNAL_ERROR_I32 => Self::InternalError,
+            code => Self::ServerError(code),
+        }
+    }
```
hinto-janai, line 150: Better way to do this?

rpc/json-rpc/src/response.rs
```diff
@@ -0,0 +1,402 @@
...
+            "result" => Key::Result,
+            "error" => Key::Error,
+            "id" => Key::Id,
+            _ => {
+                return Err(serde::de::Error::invalid_value(
+                    serde::de::Unexpected::Str(text),
+                    &self,
+                ))
+            }
+        })
```
hinto-janai, line 215: How should we treat case? I.e. should this work?:
```json
{"jSoNRPC":"2.0","iD":"0","method":"get_block"}
```
AFAICT `monerod` doesn't care about case except for both key and value of `method/params`, so the above works.

rpc/json-rpc/src/response.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+// [`Response`] has a manual deserialization implementation because
+// we need to confirm `result` and `error` don't both exist:
+//
+// > Either the result member or error member MUST be included, but both members MUST NOT be included.
+//
+// <https://www.jsonrpc.org/specification#error_object>
+impl<'de, T> Deserialize<'de> for Response<T>
```
hinto-janai, line 258: Not sure if you can do this with `serde` macros. Manually implementing is painful.

rpc/json-rpc/README.md
```diff
@@ -0,0 +1,76 @@
...
+TODO: citation needed on `monerod` not supporting batching.
+
+## Response changes
+[JSON-RPC 2.0's `Response` object](https://www.jsonrpc.org/specification#response_object) usually contains these 2 fields:
+- `method`
+- `params`
+
+This crate replaces those two with a `body` field that is `#[serde(flatten)]`ed,
+and assumes the type within that `body` field is tagged properly, for example:
+
```
Boog900, line 29: Is this supposed to be `Request changes`? (and the other usages of `response`)

rpc/json-rpc/src/error/code.rs
Boog900: If monerod doesn't then probably not

rpc/json-rpc/src/id.rs
```diff
@@ -0,0 +1,240 @@
...
+				fn from(u: Option<$u>) -> Self {
+                    match u {
+                        Some(u) => Self::Num(u as u64),
+                        None => Self::Null,
+                    }
+				}
+			}
+		)*
+	}
+}
```
Boog900, line 203: I think this was missed when changing the indentation

Boog900 (changes requested):
Gave it a quick look over and left a few comments, I'll re-review again tomorrow.

rpc/json-rpc/src/response.rs
hinto-janai: Also worth noting if we support mixed-case then I think we have to do a manual deserialization impl for `Request` as well.

rpc/json-rpc/README.md
```diff
@@ -73,4 +73,65 @@ assert_eq!(json, expected_json);
...
-For reasoning, see: <https://github.com/Cuprate/cuprate/pull/146#issuecomment-2145734838>.
\ No newline at end of file
+For reasoning, see: <https://github.com/Cuprate/cuprate/pull/146#issuecomment-2145734838>.
+
+### Serialization changes
+This crate's serialized field order slightly differs compared to `monerod`.
+
+`monerod`'s JSON objects are serialized in alphabetically order, where as this crate serializes the fields in their defined order (due to [`serde`]).
+
+With that said, parsing should be not affected at all since a key-value map is used:
```
hinto-janai, line 83: Don't think this matters, but I realized using `serde` means our fields are serialized in definition order, not alphabetical like `monerod`.

rpc/json-rpc/src/error/code.rs
Boog900: I can't think of one, this is very ugly though

rpc/json-rpc/README.md
Boog900: good spot, but should be fine

rpc/json-rpc/src/response.rs
Boog900: Is this here just because we can't have both `result` and `error`? if so, wouldn't it be possible to solve this with our own `Result` type that renames it's `Ok` to `result` and `Err` to `error`?

rpc/json-rpc/src/response.rs
```diff
@@ -0,0 +1,402 @@
...
+            Err(_) => Err(std::fmt::Error),
+        }
+    }
+}
+
+//---------------------------------------------------------------------------------------------------- Key
+/// This type represents the `"key"` values within [`Response`].
+///
+/// We need this since `Response` has a custom deserializer implementation.
+pub(crate) enum Key {
```
Boog900, line 180: Why does this need a custom serde impl?

rpc/json-rpc/src/response.rs
Boog900: I think we can get away with only accepting lower case

rpc/json-rpc/src/error/code.rs
hinto-janai: Replaced with `if` chain, seems to compile to the same assembly.

rpc/json-rpc/src/response.rs
hinto-janai: Removed with addition of `enum ResponsePayload`.

rpc/json-rpc/src/response.rs
hinto-janai: Replaced with `enum ResponsePayload`.

rpc/json-rpc/src/response.rs
hinto-janai: It's also worth noting we error instead of ignoring extra fields, unlike `monerod` and other [JSON-RPC 2.0 impls](https://docs.rs/jsonrpsee-types/0.23.1/src/jsonrpsee_types/response.rs.html#198).

I.e.:
```json
{"hello":0,"jsonrpc":"2.0","id":"0","method":"get_block"}
```
This is OK for `monerod` but will error for us. I think we need to do a manual deserialization impl to support this.

Boog900: > This means to be compliant with no result and error we must use #[serde(deny_unknown_fields)], which means we will error on extra fields

Both `result` and `error` would go to the same field which I _think_ serde disallows, might be worth adding a test to make sure?

hinto-janai: I think the tag is needed, commenting it out will allow serializing to pass: https://github.com/Cuprate/cuprate/pull/148/commits/d6646d8a12d97b50f0c79bebd9fb5e02105284ac

Boog900: IMO I don't think it's necessary that we error on both `result` and `error` being present, we can just rely on people following the spec, and if they don't that is their own problem.

It would still be impossible to create JSON that has both members using this crate.

And I don't really want to do a custom impl because of the complexity and I think it' necessary that we allow unknown fields

hinto-janai: > we can just rely on people following the spec, and if they don't that is their own problem.

I think we should enforce the spec since this is an invariant that checks future code/behavior as well. On the client side, receiving a `Response` that has both `result/error` doesn't make logical sense, assuming and serializing into one of the fields is incorrect behavior.

I.e. this should error:
```rust
// from some server
let j = json!({
    "jsonrpc": "2.0",
    "id": 0,
    "error": e,
    "result": r
});

// client
serde_json::from_value::<Response<String>>(j).unwrap();
```

> I don't really want to do a custom impl because of the complexity and I think it' necessary that we allow unknown fields

I don't think it's _too_ complex. How does a manual impl for only `Response` sound? We enforce the spec, and also allow unknown fields. `Request` can be done with macros since there's nothing to enforce. We would only allow lower-case.

rpc/json-rpc/src/error/code.rs
SyntheticBird45: I prefer seeing this syntax than an `if` chain tbh

rpc/json-rpc/src/response.rs
hinto-janai: Implementation: https://github.com/Cuprate/cuprate/blob/007c9adac735654a9f5bceb06fd13374aed8d729/rpc/json-rpc/src/response.rs#L203-L345

A bit long but I don't think it's too bad.

rpc/json-rpc/src/response.rs
```diff
@@ -0,0 +1,434 @@
...
+                            // Ignore any other keys that appear
+                            // and continue deserialization.
+                            _ => Ok(Key::Unknown),
+                        }
+                    }
+                }
+
+                deserializer.deserialize_identifier(KeyVisitor)
+            }
+        }
```
Boog900, line 268: Would it be possible to just use `&str` as the key type so we have to do this?

Boog900: I just treid this and it turns out you can't always borrow a str from the data so instead of `&std` `Cow<str>` works

Boog900: discussed in a meeting: https://github.com/monero-project/meta/issues/1021 this is slightly less optimal

rpc/json-rpc/README.md
```diff
@@ -0,0 +1,197 @@
...
+// But 2 results are not allowed.
+let json = r#"{"jsonrpc":"2.0","id":123,"result":"OK","result":"OK"}"#;
+let err = from_str::<Response<String>>(&json).unwrap_err();
+assert_eq!(format!("{err}"), "duplicate field `result/error` at line 1 column 48");
+
+// Same with errors.
+let json = r#"{"jsonrpc":"2.0","id":123,"error":{"code":-1,"message":""},"error":{"code":-1,"message":""}}"#;
+let err = from_str::<Response<String>>(&json).unwrap_err();
+assert_eq!(format!("{err}"), "duplicate field `result/error` at line 1 column 66");
+```
```
hinto-janai, line 197: Some differences between some implementations and ours.

hinto-janai (commented):
Added tests and manually implemented `serde::Deserialize` for `Request<T>` using an `enum Key` to match over key fields.

hinto-janai:
https://repo.msys2.org is down, CI should be fine: https://github.com/Cuprate/cuprate/actions/runs/9473729258/job/26103499719?pr=148.

Boog900 (approved):

PR #154, opened 2024-06-07
books: add `user-book` template

hinto-janai:
Adds the skeleton for Cuprate's user book: the end-user documentation for Cuprate, aka, "how to use `cuprated`".

Cramming everything into the architecture book doesn't make sense since 95% of it doesn't matter in a practical way for our users, so that stuff will go into this book (among other things), e.g. instrumentation, configuration, etc.

Like the other books, this is just a skeleton for now.

@Boog900 requesting:
- An OK on creating a `Cuprate/user-book` repo
- A DNS entry for `user.cuprate.org` pointing to https://github.com/Cuprate/user-book


Boog900 (commented):
Yeah sounds good, ping me when the repo is created

hinto-janai:
@Boog900 ready.

Boog900 (approved):

PR #160, opened 2024-06-12
storage: split `cuprate-blockchain` <-> `cuprate-database`

hinto-janai:
Separates the database abstraction layer & the Monero/blockchain-related code into:
- `cuprate-database`
- `cuprate-blockchain`

respectively, with the latter being built on-top of the former.


storage/database/src/env.rs
```diff
@@ -81,13 +80,13 @@ pub trait Env: Sized {
     /// Open the database environment, using the passed [`Config`].
     ///
     /// # Invariants
-    /// This function **must** create all tables listed in [`crate::tables`].
+    /// This function does not create any tables.
     ///
-    /// The rest of the functions depend on the fact
-    /// they already exist, or else they will panic.
+    /// You must create all possible tables with [`EnvInner::create_db`]
+    /// before attempting to open any.
```
hinto-janai, line 86: ## `cuprate_database::Env` 1/3

`cuprate_database` no longer has any idea what the user is doing since it is _only_ a database abstraction. This means it doesn't know what tables the user wants, so this invariant is passed on to users now.

This means after `cuprate_database::Env::open`, you have a database but with no tables and it is up to you to create them (where as before `cuprate_database` created all the blockchain tables).

storage/database/src/env.rs
```diff
@@ -246,32 +259,22 @@ where
...
     ///
-    #[doc = doc_table_error!()]
-    fn open_tables_mut(&self, tx_rw: &Rw) -> Result<impl TablesMut, RuntimeError> {
-        call_fn_on_all_tables_or_early_return! {
-            Self::open_db_rw(self, tx_rw)
-        }
-    }
+    /// # Errors
+    /// This will only return [`RuntimeError::Io`] on errors.
+    fn create_db<T: Table>(&self, tx_rw: &Rw) -> Result<(), RuntimeError>;
```
hinto-janai, line 277: ## `cuprate_database::Env` 2/3

There is a new function: `create_db`, it just creates the passed `Table`.

The existing:
- `EnvInner::open_db_ro` will error if the table doesn't exist
- `EnvInner::open_db_rw` will create the table if it doesn't exist. This is nice side effect since both `redb/heed` do this but it is documented for clarity that `create_db` should be used just in case

storage/database/src/table.rs
```diff
@@ -8,12 +8,7 @@ use crate::{key::Key, storable::Storable};
 /// Database table metadata.
 ///
 /// Purely compile time information for database tables.
-///
-/// ## Sealed
-/// This trait is [`Sealed`](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed).
-///
-/// It is only implemented on the types inside [`tables`][crate::tables].
-pub trait Table: crate::tables::private::Sealed + 'static {
+pub trait Table: 'static {
```
hinto-janai, line 11: ## `cuprate_database::Table` 1/4
Since `cuprate_database` doesn't decide the tables anymore, this is no longer sealed. Any table can be created.

storage/database/src/env.rs
```diff
@@ -246,32 +259,22 @@ where
...
-    /// and returns a structure that allows access to all tables.
+    /// # Errors
+    /// This will only return [`RuntimeError::Io`] on errors.
     ///
-    #[doc = doc_table_error!()]
-    fn open_tables(&self, tx_ro: &Ro) -> Result<impl TablesIter, RuntimeError> {
-        call_fn_on_all_tables_or_early_return! {
-            Self::open_db_ro(self, tx_ro)
-        }
-    }
```
hinto-janai, line 262: ## `cuprate_database::Table` 2/4
Since `cuprate_database` has no idea what the user is doing, this function can't exist anymore. What tables should we create? That is up to the user now.

storage/blockchain/src/open_tables.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// Create all database tables.
+    ///
+    /// This will create all the [`Table`](cuprate_database::Table)s
+    /// found in [`tables`](crate::tables).
+    ///
+    /// # Errors
+    /// This will only return [`RuntimeError::Io`] on errors.
+    fn create_tables(&'env self, tx_rw: &Rw) -> Result<(), RuntimeError>;
+}
```
hinto-janai, line 122: ## `cuprate_database::Table` 3/4
So, `cuprate_blockchain` defines the type/function that opens all tables.

This new trait `OpenTables` is basically just a copy of the previous:
- `EnvInner::open_tables`
- `EnvInner::open_tables_mut`

It is generically implemented on `EnvInner` such that `self.method()` notation works:
```rust
use cuprate_blockchain::OpenTables;

env_inner.open_tables();
```

storage/blockchain/src/free.rs
```diff
@@ -1,8 +1,73 @@
...
+            return Err(runtime_to_init_error(e));
+        };
+
+        if let Err(e) = tx_rw.commit() {
+            return Err(runtime_to_init_error(e));
+        }
+    }
+
+    Ok(env)
+}
```
hinto-janai, line 70: ## `cuprate_database::Table` 4/4
`cuprate_blockchain` now exposes the `open()` function to use, this ensures the blockchain tables (and other stuff) are ready before usage.

storage/database/src/config/config.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// [`ConfigBuilder::build`] can be called immediately
+    /// after this function to use default values.
+    pub const fn new(db_directory: Cow<'static, Path>) -> Self {
+        Self {
+            db_directory,
+            sync_mode: None,
+            reader_threads: Some(READER_THREADS_DEFAULT),
+            resize_algorithm: None,
+        }
+    }
```
hinto-janai, line 55: ## `Config` 1/2
`cuprate_database`'s config now _requires_ a PATH.

`cuprate_blockchain` just passes the PATH we used before.

storage/blockchain/src/config/config.rs
```diff
@@ -65,163 +57,137 @@ impl ConfigBuilder {
...
+/// This is a configuration built on-top of [`cuprate_database::config::Config`].
 ///
-/// For construction, either use [`ConfigBuilder`] or [`Config::default`].
+/// It contains configuration specific to this crate, plus the database config.
 ///
-// SOMEDAY: there's are many more options to add in the future.
+/// For construction, either use [`ConfigBuilder`] or [`Config::default`].
 #[derive(Debug, Clone, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Config {
```
hinto-janai, line 151: ## `Config` 2/2
`cuprate_blockchain` also has a `Config`, which is built on-top of `cuprate_database`.

For now, the only extra data this contains is the `ReaderThreads` for the `tower::Service`.

Implementation detail: LMDB allows setting max reader threads so this is actually used in `cuprate_database` for that too (although it is passed as a `NonZeroUsize`). Not sure where the appropriate place is to define this type since it has a double meaning here. Maybe this should be documented...?

consensus/fast-sync/src/create.rs
```diff
@@ -3,7 +3,9 @@ use std::{fmt::Write, fs::write};
 use clap::Parser;
 use tower::{Service, ServiceExt};
 
-use cuprate_blockchain::{config::ConfigBuilder, service::DatabaseReadHandle, RuntimeError};
+use cuprate_blockchain::{
+    config::ConfigBuilder, cuprate_database::RuntimeError, service::DatabaseReadHandle,
+};
```
hinto-janai, line 8: Not sure what re-export standards are, but `cuprate_blockchain` re-exports `cuprate_database` like this so the types/traits match up.

Maybe it should be `cuprate_blockchain::database`?

storage/blockchain/DESIGN.md
```diff
@@ -0,0 +1,600 @@
+# Database
+FIXME: This documentation must be updated and moved to the architecture book.
```
hinto-janai, line 2: The old database design document is in here for now.

Will remove after porting to architecture book.

storage/database/src/tests.rs
```diff
@@ -0,0 +1,35 @@
+//! Utilities for `cuprate_database` testing.
+//!
```
hinto-janai, line 2: None of the tests were changed, just kept database ones in `cuprate_database` and moved blockchain ones to `cuprate_blockchain`.

storage/database/src/error.rs
```diff
@@ -88,6 +88,10 @@ pub enum RuntimeError {
     #[error("database memory map must be resized")]
     ResizeNeeded,
 
+    /// The given table did not exist in the database.
+    #[error("database table did not exist")]
+    TableNotFound,
```
hinto-janai, line 93: Since `cuprate_database` doesn't know all the user's tables anymore, this runtime error now exists.

Although, `cuprate_blockchain::open` creates all the tables so this error never occurs.

storage/database/src/env.rs
```diff
@@ -81,13 +80,13 @@ pub trait Env: Sized {
     /// Open the database environment, using the passed [`Config`].
     ///
     /// # Invariants
-    /// This function **must** create all tables listed in [`crate::tables`].
+    /// This function does not create any tables.
```
hinto-janai, line 83: Another thing about `cuprate_database` not knowing about the user's tables:

There's going to have to be a "change table sort order" API created in `cuprate_database` for https://github.com/Cuprate/cuprate/issues/179.

Probably using `trait cuprate_database::Key`? Discussion for another time though.

hinto-janai (commented):
@Boog900

There's not actually too many semantic changes, just a bunch of moving around. Lots of traits/types/functions that were in `cuprate_database` are now in `cuprate_blockchain`.

I've noted the bigger changes below, the things separated are pretty much what you'd expect.

storage/database/src/error.rs
hinto-janai: Also, `cuprate_database` no longer panics if a table doesn't exist, it returns this error.

hinto-janai: Actually... after your review I'm going to add a test for this.

storage/database/src/env.rs
hinto-janai: BTW this should have been done in the initial implementation, sorry.

Boog900: I think a good way to do it would be with an enum like:

```rust
enum Sort {
    Default,
    Integer,
    Custom(compare_func)
}
```

As some DBs (for example LMDB) have specific options for integer sorting which should be faster than doing custom sorting

Boog900 (commented):
Wasn't the easiest thing to review but looks good!

consensus/fast-sync/src/create.rs
hinto-janai: It should only be `{RuntimeError, InitError}` that is needed.

Although `cuprate_blockchain::service::DatabaseReadHandle::env() -> &Arc<ConcreteEnv>` exposes the actual database, which would only be used with a bunch of the traits from `cuprate_database`.

I set this to errors only for now, can change later.

hinto-janai: Actually, `cuprate_blockchain` needs a few traits internally. Is there a downside to re-exporting all of `cuprate_database`?

Boog900: Just that there is extra indirection, needing to go through `cuprate_database` first.

storage/database/src/error.rs
Boog900: Is this planned for this PR?

storage/blockchain/src/ops/mod.rs
```diff
@@ -55,25 +55,28 @@
...
+//!         Env, EnvInner,
+//!         DatabaseRo, DatabaseRw, TxRo, TxRw,
+//!     },
+//!     OpenTables,
 //!     config::ConfigBuilder,
-//!     Env, EnvInner,
 //!     tables::{Tables, TablesMut},
-//!     DatabaseRo, DatabaseRw, TxRo, TxRw,
 //!     ops::block::{add_block, pop_block},
 //! };
```
hinto-janai, line 68: @Boog900 Lots of the `cuprate_database` traits are needed for `cuprate_blockchain::ops`.

If we scoped the required imports, it looks like:
```rust
pub use cuprate_database::{
    ConcreteEnv, Env, EnvInner,
    InitError, RuntimeError,
    DatabaseRo, DatabaseRw, DatabaseIter,
    TxRo, TxRw
};
```
...that's pretty much everything :D

consensus/fast-sync/src/create.rs
hinto-janai: https://github.com/Cuprate/cuprate/pull/160/files#r1655543354

storage/database/src/error.rs
hinto-janai: Oops, yes, but actually no. These tests already exist: https://github.com/hinto-janai/cuprate/blob/1648409727b00ee1032b866f4359fa4aea678e7a/storage/database/src/backend/tests.rs#L70-L116.

I added more tests to `cuprate_blockchain` though.

Boog900 (approved):

PR #161, opened 2024-06-13
Fix clippy & Cargo update

Boog900:


deny.toml
```diff
@@ -120,6 +120,7 @@ allow = [
 
     # Font licenses.
     "Unicode-DFS-2016",     # https://spdx.org/licenses/Unicode-DFS-2016.html
+    "Unicode-3.0",          # https://spdx.org/licenses/Unicode-3.0.html
```
hinto-janai, line 123: Do you know which crate is pulling this in?

rpc/json-rpc/src/id.rs
```diff
@@ -182,20 +182,20 @@ macro_rules! impl_u {
         $(
             impl From<$u> for Id {
                 fn from(u: $u) -> Self {
-                    Self::Num(u as u64)
+                    Self::Num(u64::from(u))
```
hinto-janai, line 185: Oops, original impl took `i8-i64` and `f32/f64` as input which is why `as` was used.

hinto-janai (commented):
Sorry, the `#![cfg(test)]` at [`rpc/json-rpc/src/tests.rs:3`](https://github.com/Cuprate/cuprate/actions/runs/9503481103/job/26193991437#step:11:62) should be removed.

deny.toml
Boog900: heed pulls in `url` which pulls in crates from here: https://github.com/unicode-org/icu4x

hinto-janai (approved):

PR #170, opened 2024-06-18
CI: add cargo hack

Boog900:
adds cargo hack to CI and fixes errors found.


.github/workflows/hack.yml
```diff
@@ -0,0 +1,35 @@
...
+          submodules: recursive
+
+      - name: Install dependencies
+        run: |
+          sudo apt install -y libboost-dev
+          cargo install cargo-hack --locked
+
+      - name: Hack Check
+        # TODO: remove the exclude when `cuprate-database` errors are fixed
+        run: cargo hack --workspace --exclude cuprate-database check --feature-powerset --no-dev-deps
```
Boog900, line 35: `cuprate-database` does not build with `--no-default-features` and I don't think there is a way to fix this without hot swapable backends, so another reason for that :)

hinto-janai: Should we set `cuprate-database` to always pull in `heed` by default? Then if another backend feature is detected, that is used instead, although `heed` will still be built (but not used).

CONTRIBUTING.md
```diff
@@ -120,12 +120,13 @@ Before pushing your code, please run the following at the root of the repository
...
+|--------------------------------------------------------------------------------------------|-----------------------------------------------------------------------|
+| `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features`                          | Checks documentation is OK                                            |
+| `cargo clippy --workspace --all-features --all-targets -- -D warnings`                     | Checks clippy lints are satisfied                                     |
+| `cargo test --all-features --workspace`                                                    | Runs all tests                                                        |
+| `cargo build --all-features --all-targets --workspace`                                     | Builds all code                                                       |
+| `cargo hack --workspace --exclude cuprate-database check --feature-powerset --no-dev-deps` | Uses cargo hack to check our crates build with different features set |
 
 **Note: in order for some tests to work, you will need to place a [`monerod`](https://www.getmonero.org/downloads/) binary at the root of the repository.**
```
hinto-janai, line 131:
```suggestion
| `cargo hack --workspace --exclude cuprate-database check --feature-powerset --no-dev-deps` | Uses `cargo hack` to check our crates build with different features set |

`cargo hack` can be installed with `cargo` from: https://github.com/taiki-e/cargo-hack.

**Note: in order for some tests to work, you will need to place a [`monerod`](https://www.getmonero.org/downloads/) binary at the root of the repository.**
```

.github/workflows/hack.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo
+            target
+          key: ${{ matrix.os }}-hack
```
hinto-janai, line 22: 1. Windows and macOS should be added too
2. `${{ matrix.os }}` is an empty variable since no matrix is setup

It may be worth putting this in `ci.yml` which already has all the Windows workarounds / dependency stuff.

hinto-janai (changes requested):

.github/workflows/hack.yml
Boog900: We could, initially I didn't want to, but thinking about it more, with hot-swapable backends both `heed` & `redb` would need to be pulled in all the time as well.

I'll add this

.github/workflows/hack.yml
Boog900: ah that's what I get for blindly copying another workflow as a template :p

This workflow takes awhile to complete (~7 mins), so I don't really want to add it as another step to `ci`, I think just testing linux should be enough, we already test full builds with Windows/ MacOS.

If we do add Windows/ MacOS support, it may make sense to separate the `ci` setup into a separate action so that both the `ci` and `hack` workflows can use it.

hinto-janai:
Btw merging this can close https://github.com/Cuprate/cuprate/issues/325 I think?

rpc/types/src/misc/distribution.rs
```diff
@@ -20,8 +20,8 @@ use cuprate_epee_encoding::{
     "rpc/core_rpc_server_commands_defs.h",
     45..=55
 )]
-#[cfg(feature = "epee")]
-fn compress_integer_array(_: &[u64]) -> error::Result<Vec<u8>> {
+#[cfg(any(feature = "epee", feature = "serde"))]
+fn compress_integer_array(_: &[u64]) -> Vec<u8> {
```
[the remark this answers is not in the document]
Boog900: No this uses a different varint format: https://github.com/Cuprate/cuprate/pull/229#discussion_r1690531904

types/Cargo.toml
```diff
@@ -12,21 +12,23 @@ keywords    = ["cuprate", "types"]
 default    = ["blockchain", "epee", "serde", "json", "hex"]
 blockchain = []
 epee       = ["dep:cuprate-epee-encoding"]
-serde      = ["dep:serde"]
+serde      = ["dep:serde", "hex"]
 proptest   = ["dep:proptest",  "dep:proptest-derive"]
 json       = ["hex", "dep:cuprate-helper"]
-hex        = ["dep:hex"]
+# We sadly have no choice but to enable serde here as otherwise we will get warnings from the `hex` dep being unused.
```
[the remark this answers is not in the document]
Boog900: It would but IMO just enabling serde here is better as the enabled type literally only makes sense with serde

storage/txpool/Cargo.toml
```diff
@@ -9,30 +9,29 @@ repository  = "https://github.com/Cuprate/cuprate/tree/main/storage/txpool"
...
 
 [features]
-default     = ["heed", "service"]
+default     = ["heed"]
 # default   = ["redb", "service"]
 # default   = ["redb-memory", "service"]
 heed        = ["cuprate-database/heed"]
 redb        = ["cuprate-database/redb"]
 redb-memory = ["cuprate-database/redb-memory"]
-service     = ["dep:tower", "dep:rayon", "dep:cuprate-database-service"]
```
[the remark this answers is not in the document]
Boog900: Because it was annoying to get this to pass CI so I just removed the feature, almost all use cases will use the tower interface anyway, for those that don't I don't think a few extra dependencies would be too bad.

I for parity removed the feature flag from `cuprate-blockchain` as well and updated the docs

hinto-janai (approved):

PR #171, opened 2024-06-18
Cargo update

Boog900:
Fixes deny, `curve25519-dalek` had a timing vulnerability that doesn't affect us.


Boog900:
Also remove a lot of deps, it seems `url` put out a new version that reduced their dependencies

hinto-janai (approved):

PR #182, opened 2024-06-20
repo: add `Tracking Issue` issue template

hinto-janai:
This adds `.github/ISSUE_TEMPLATE/tracking_issue.md` and fixes some formatting in other templates.

The purpose of this is to track wider efforts that may contain multiple PRs and/or are generally spread out. These don't usually contain the why, but if they do, they are brief. These contain no implementation detail or the how, as those are for the issues/PRs that are being tracked.

For example:
- https://github.com/rust-lang/rust/issues?q=is%3Aissue+is%3Aopen+label%3AC-tracking-issue
- https://github.com/rust-lang/rust/issues/123916

A big difference between this and a proposal is that tracking issues already have consensus, i.e. the thing in question is already decided on. The what/why/where is already answered, all that is left is implementation.

### DIFF
--- a/.github/ISSUE_TEMPLATE/tracking_issue.md
+++ b/.github/ISSUE_TEMPLATE/tracking_issue.md
@@ -1,54 +1,41 @@
 ---
 name: 🔍 Tracking Issue
-about: Create an issue that tracks an effort or PR(s)
+about: Create an issue that tracks a wider effort
 title: 'Tracking Issue for ...'
 labels: ["C-tracking-issue"]
 assignees: ''
 
 ---
 
-<!--
-Note: Please search to see if an issue already exists for this tracking issue.
-
-About tracking issues:
-
+<!-- Consider keeping the following section in the issue. -->
+### About tracking issues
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/tracking_issue.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+<!-- Consider keeping the following section in the issue. -->
+### About tracking issues
+Tracking issues are used to record the overall progress of implementation.
+They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
+A tracking issue is however not meant for large scale discussion, questions, or bug reports about a feature.
+Instead, open a dedicated issue for the specific matter.
+
+### What
```
hinto-janai, line 17: These are `###` instead of `##` like other issue templates because:
1. `rust-lang/rust` does it like this
2. It makes this template stand-out

hinto-janai:
Active example: https://github.com/Cuprate/cuprate/issues/183

Boog900 (approved):

PR #185, opened 2024-06-21
epee: make `{read,write}_varint` public, create `write_{bytes,container}`

hinto-janai:
- `epee_encoding::varint::{read,write}_varint()` is made public
- The duplicated inner implementations of some `EpeeValue` objects are extracted into `epee_encoding::write_{bytes,iterator}` and made public and generic

For example, a custom `EpeeValue` implementor can now do:
```rust
impl EpeeValue for CustomStr {
    /* ... */

    fn write<B: BufMut>(self, w: &mut B) -> Result<()> {
        epee_encoding::write_bytes(self, w)
    }
}
```

### DIFF
--- a/net/epee-encoding/src/lib.rs
+++ b/net/epee-encoding/src/lib.rs
@@ -242,7 +242,7 @@ pub fn write_bytes<T: AsRef<[u8]>, B: BufMut>(t: T, w: &mut B) -> Result<()> {
     Ok(())
 }
 
-/// Write a container of [`EpeeValue`]s to `w` with [`write_varint`].
+/// Write an [`Iterator`] of [`EpeeValue`]s to `w` with [`write_varint`].
 ///
 /// This function:
 /// - Writes the length of `i`'s bytes into `w` using [`write_varint`]
@@ -255,7 +255,7 @@ pub fn write_bytes<T: AsRef<[u8]>, B: BufMut>(t: T, w: &mut B) -> Result<()> {
 /// This will error if:
 /// - [`write_varint`] fails
 /// - [`EpeeValue::<T>::write`] fails
-pub fn write_container<T, I, B>(iterator: I, w: &mut B) -> Result<()>
+pub fn write_iterator<T, I, B>(iterator: I, w: &mut B) -> Result<()>
 where
     T: EpeeValue,
     I: Iterator<Item = T> + ExactSizeIterator,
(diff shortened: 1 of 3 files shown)


net/epee-encoding/src/value.rs
```diff
@@ -351,15 +317,7 @@ impl<const N: usize> EpeeValue for ByteArray<N> {
...
-        write_varint(N.try_into().unwrap(), w)?;
-
-        if w.remaining_mut() < N {
-            return Err(Error::IO("Not enough capacity to write bytes"));
-        }
-
-        w.put(bytes);
-        Ok(())
+        write_bytes(bytes, w)
     }
```
hinto-janai, line 321: Pretty sure `N == bytes.len()` here, so no semantic change.

net/epee-encoding/src/lib.rs
```diff
@@ -212,6 +212,87 @@ fn write_epee_value<T: EpeeValue, B: BufMut>(val: T, w: &mut B) -> Result<()> {
...
+    T: EpeeValue,
+    I: Iterator<Item = T> + ExactSizeIterator,
+    B: BufMut,
+{
+    write_varint(iterator.len().try_into()?, w)?;
+    for item in iterator.into_iter() {
+        item.write(w)?;
+    }
+    Ok(())
+}
```
hinto-janai, line 294: Is there a more generic `I` type that would accept more inputs?

net/epee-encoding/src/lib.rs
```diff
@@ -212,6 +212,87 @@ fn write_epee_value<T: EpeeValue, B: BufMut>(val: T, w: &mut B) -> Result<()> {
...
+
+    w.put_slice(bytes);
+
+    Ok(())
+}
+
+/// Write an [`Iterator`] of [`EpeeValue`]s to `w` with [`write_varint`].
+///
+/// This function:
+/// - Writes the length of `i`'s bytes into `w` using [`write_varint`]
```
Boog900, line 259: This should say the length of the iterator, not the length of the bytes, also is it supposed to be `i` or `iterator`?

net/epee-encoding/src/lib.rs
Boog900: I can't think of one

Boog900 (approved):

PR #198, opened 2024-06-27
storage: add key sorting

### LINKED ISSUE #179: Database keys are not sorted correctly.
## Bug

integer keys in the database are sorted lexicographically, which is not how they should be stored.

### Expected behavior
Integer keys should be sorted by value.

## Steps to reproduce

Increase the range here and it will fail:

https://github.com/Cuprate/cuprate/blob/b76042a4e4f364aa2c3b53235f30382ac632d2e8/storage/cuprate-blockchain/src/backend/tests.rs#L336-L340

hinto-janai:
### What
Makes some changes to the `cuprate_database::Key` trait to support custom key sorting for both backends, with types and function definitions (only at compile time).

Fixes #179.

### Where
`cuprate_database` public addition:
```rust
pub enum KeyCompare {
    Default,
    Number,
    Custom(fn(&[u8], &[u8]) -> Ordering),
}
```
```rust
pub struct StorableStr(pub Cow<'static, str>);
/* impl Storable and Key */
```

`cuprate_database` public edit:
```rust
pub trait Key: Storable + Sized {
    const KEY_COMPARE: KeyCompare = KeyCompare::Default;
}
/* impl correct compare order for types */
```

`cuprate_database` private addition:
```rust 
/* impl heed::Comparator using Key::KEY_COMPARE */
```


storage/database/src/key.rs
```diff
@@ -1,39 +1,71 @@
...
+/// This default is overridden for numbers, which use a number comparison.
+/// For example, [`u64`] keys are sorted as `{0, 1, 2 ... 999_998, 999_999, 1_000_000}`.
+///
+/// If you would like to re-define this for number types, consider creating a
+/// wrapper type around primitives like a `struct SortU8(pub u8)` and implement
+/// [`Storable`], [`Key`], and define a custom [`Key::compare`] function.
+// FIXME:
+// implement getting values using ranges.
 // <https://github.com/Cuprate/cuprate/pull/117#discussion_r1589378104>
 pub trait Key: Storable + Sized {
```
hinto-janai, line 32: ## Key 1/4

Like redb, all of `cuprate_database`'s key types must implement `Key`, which must define a sort order.

There is no diff in the `redb` backend folder because we were already forwarding `Key::compare` to it, however...

storage/database/src/backend/heed/env.rs
```diff
@@ -282,10 +280,16 @@ where
...
+        let db = {
+            let mut tx_rw = tx_rw.borrow_mut();
+            self.database_options()
+                .types()
+                .key_comparator()
+                .name(T::NAME)
+                .create(&mut tx_rw)?
+        };
+
+        Ok(HeedTableRw { db, tx_rw })
```
hinto-janai, line 292: ## Key 2/4
... `heed` requires this info at runtime (kinda). We can actually do this at compile time.
Opening/creating a table now:
- Creates a database builder
- Sets the correct key/value types
- Sets the correct sort order

The user does not have to call a "set sort order" function, they define their sort order at compile time with `trait Key`, which forwards to...

storage/database/src/backend/heed/storable.rs
```diff
@@ -16,7 +16,19 @@ pub(super) struct StorableHeed<T>(PhantomData<T>)
...
+// If `Key` is also implemented, this can act as the comparison function.
+impl<T> heed::Comparator for StorableHeed<T>
+where
+    T: Key,
+{
+    #[inline]
+    fn compare(a: &[u8], b: &[u8]) -> Ordering {
+        <T as Key>::compare(a, b)
+    }
+}
```
hinto-janai, line 29: ## Key 3/4
This implementation on our compatibility wrapper type. It just forwards our `Key::compare` to `heed`'s compare function.

Now the database table will use `Key::compare` to compare keys.

storage/database/src/key.rs
```diff
@@ -44,11 +76,41 @@ pub trait Key: Storable + Sized {
...
+                }
+            }
+        )*
+    };
+}
+
+impl_key_ne_bytes! {
+    u8,u16,u32,u64,u128,usize,
+    i8,i16,i32,i64,i128,isize,
 }
```
hinto-janai, line 114: ## Key 4/4
`Key`'s blanket impl is removed and replaced with some implementation for common types.

These impls now implement the behavior that `cuprate_blockchain` wants by default.

| Common type | Comparison behavior |
|-------------|---------------------|
| Vectors/arrays/string | Straight byte comparison; `a.cmp(&b)`
| Numbers | Convert to number, then do a _value_ comparison

The reason why strings are treated the same as bytes is because [Rust does it like that](https://doc.rust-lang.org/1.79.0/src/core/str/traits.rs.html#12-25). We can change it but I don't think we have any tables that care about this.

The reason why the blanket impl is removed is it allows outer crates to define their own
comparison functions for their `T: Storable` types.

The reverse way of doing this is:
- Leaving the blanket impl
- Doing byte comparisons for everything by default

Then in `cuprate_blockchain` we would have to create wrapper types for all the numbers, e.g. `struct SortU8(pub u8)` and implement the custom sorting. This is probably more clean and separated but kind of a pain.

It's worth noting in `redb`:
- [Number comparison is the default](https://docs.rs/redb/2.1.1/src/redb/types.rs.html#601-624)
- [Key types are selectively implemented, not blanket impl on all values](https://docs.rs/redb/2.1.1/redb/trait.Key.html)

So I think doing it like this is OK.

storage/database/src/key.rs
```diff
@@ -1,39 +1,71 @@
...
     /// assert_eq!(
-    ///     <u64 as Key>::compare([1].as_slice(), [1].as_slice()),
-    ///     std::cmp::Ordering::Equal,
-    /// );
-    /// assert_eq!(
-    ///     <u64 as Key>::compare([2].as_slice(), [1].as_slice()),
+    ///     //                    256 > 255
+    ///     <u16 as Key>::compare(&byte1, &byte2),
     ///     std::cmp::Ordering::Greater,
     /// );
```
hinto-janai, line 70: Example 1/2 of integer keys being compared by their _value_ and not bytes.

storage/database/src/backend/tests.rs
```diff
@@ -323,19 +323,27 @@ fn db_read_write() {
...
+    // number inserted is the minimum `last()` value.
+    for key in RANGE {
         table.put(&key, &123).unwrap();
+        table.contains(&key).unwrap();
         let (first, _) = table.first().unwrap();
-        assert_eq!(first, key);
+        let (last, _) = table.last().unwrap();
+        println!("{key}, {first}, {last}");
+        assert_eq!(last, key);
     }
```
hinto-janai, line 347: Example 2/2.

storage/database/src/key.rs
```diff
@@ -1,39 +1,71 @@
...
+///
+/// If you would like to re-define this for number types, consider creating a
+/// wrapper type around primitives like a `struct SortU8(pub u8)` and implement
+/// [`Storable`], [`Key`], and define a custom [`Key::compare`] function.
+// FIXME:
+// implement getting values using ranges.
 // <https://github.com/Cuprate/cuprate/pull/117#discussion_r1589378104>
 pub trait Key: Storable + Sized {
-    /// The primary key type.
-    type Primary: Storable;
```
hinto-janai, line 18: This `Primary` type was removed. This is leftover from before, but no longer makes sense since the type implementing `Key` _is_ the primary type.

Will need changing if/when multi-map support is added.

storage/database/src/storable.rs
```diff
@@ -194,6 +197,68 @@ impl<T> Borrow<[T]> for StorableVec<T> {
...
+/// let into = Storable::as_bytes(&string);
+/// assert_eq!(into, &[97]);
+///
+/// // From bytes.
+/// let from: StorableStr = Storable::from_bytes(&into);
+/// assert_eq!(from, string);
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, bytemuck::TransparentWrapper)]
+#[repr(transparent)]
+pub struct StorableStr(pub Cow<'static, str>);
```
hinto-janai, line 228: This struct was also added.

Basically the same as `StorableVec` but for strings. `Cow<'static, str>` instead of `String` to allow `Cow::Borrowed("static string")`.

Reason why this was added was because I wondering if we should do custom string comparisons, but chose not to. Can remove this.

storage/blockchain/src/types.rs
```diff
@@ -143,6 +143,8 @@ pub struct PreRctOutputId {
     pub amount_index: AmountIndex,
 }
 
+impl Key for PreRctOutputId {}
+
```
hinto-janai, line 147: I think we'll need a multi-map abstraction for FCMPs btw (and it can replace this type).

This was the only type that needed to be implemented after removing the `Key` blanket impl.

hinto-janai (commented):
Some background:

- Problem: number keys are not sorted by value
- Solution proposed before: runtime function to change sort order
- Solution in this PR: define comparison behavior _at compile time_ by implementing `trait Key`, and forward that to the backends

[`heed` has a runtime function that changes database sort order](https://docs.rs/heed/0.20.2/heed/struct.DatabaseOpenOptions.html#method.key_comparator)

[`redb`'s keys _must_ have a defined comparison order at compile time](https://docs.rs/redb/2.1.1/redb/trait.Key.html)

This PR implements this at compile time for the 2 backends in the following way:

hinto-janai:
Re-reading this comment: https://github.com/Cuprate/cuprate/pull/160#discussion_r1655141543:

> As some DBs (for example LMDB) have specific options for integer sorting which should be faster than doing custom sorting

I'm going to work on this PR more to support this.

storage/database/src/key.rs
```diff
@@ -1,54 +1,173 @@
...
+    ///     //                                               256 > 255
+    ///     <u16 as Key>::KEY_COMPARE.as_compare_fn::<u16>()(&byte1, &byte2),
     ///     std::cmp::Ordering::Greater,
     /// );
     /// ```
-    #[inline]
-    fn compare(left: &[u8], right: &[u8]) -> Ordering {
-        left.cmp(right)
-    }
+    const KEY_COMPARE: KeyCompare = KeyCompare::Lexicographic;
```
hinto-janai, line 72: ## Key 1/4
`trait Key`'s custom compare function was replaced with an `enum` defined...

storage/database/src/key.rs
```diff
@@ -1,54 +1,173 @@
...
+            Self::Lexicographic => std::cmp::Ord::cmp,
+            Self::Number => |left, right| {
+                let left = <K as Storable>::from_bytes(left);
+                let right = <K as Storable>::from_bytes(right);
+                std::cmp::Ord::cmp(&left, &right)
+            },
+            Self::Custom(f) => f,
+        }
+    }
 }
```
hinto-janai, line 171: ## Key 2/4
...here.

storage/database/src/key.rs
```diff
@@ -1,54 +1,173 @@
...
+                // Just forward the the number comparison function.
+                const KEY_COMPARE: KeyCompare = KeyCompare::Custom(|left, right| {
+                    KeyCompare::Number.as_compare_fn::<$t>()(left, right)
+                });
+            }
+        )*
+    };
+}
+
+impl_custom_numbers_key!(u8, u16, u128, i8, i16, i32, i64, i128, isize);
```
hinto-janai, line 129: ## Key 3/4
Implementation detail: LMDB's `INTEGER_KEY` flag only supports `u32`, `u64` and `usize` if it's either of those.

`redb` supports all numbers out-the-box though, so our workaround is:

1. We implement `Key` for `u32, u64, usize` and set `KeyCompare::Number`
2. For all other numbers we set `KeyCompare::Custom(/* number compare fn */)`

then in the `heed` backend implementation...

storage/database/src/backend/heed/env.rs
```diff
@@ -282,10 +282,32 @@ where
...
+                }
+
+                // Use a custom comparison function if specified.
+                KeyCompare::Custom(_) => {
+                    db.key_comparator::<T::Key>();
+                }
+            }
+
+            db.types().name(T::NAME).create(&mut tx_rw)?
+        };
```
hinto-janai, line 308: ## Key 4/4
...we forward the `KeyCompare::Number` to use the faster flag instead. We only fall back to custom integer comparison if the number isn't supported.

Our only number keys for `cuprate_blockchain` are `u64` so it doesn't really matter to support this for _us_, but I thought it would be nice to maintain the sorted behavior for _all_ numbers, like `redb`.

storage/database/src/backend/heed/database.rs
```diff
@@ -26,19 +28,19 @@ use crate::{
...
 }
 
 /// An opened read/write database associated with a transaction.
 ///
 /// Matches `redb::Table` (read & write).
-pub(super) struct HeedTableRw<'env, 'tx, T: Table> {
+pub(super) struct HeedTableRw<'env, 'tx, T: Table, C: Comparator = DefaultComparator> {
     /// An already opened database table.
-    pub(super) db: HeedDb<T::Key, T::Value>,
+    pub(super) db: HeedDb<T::Key, T::Value, C>,
```
hinto-janai, line 43: Sidenote: we have to now define this `C: heed::Comparator` everywhere now...

hinto-janai (commented):
OK, the previous review is still mostly accurate, just 1 small change to how `trait Key` works.

storage/blockchain/src/types.rs
Boog900: We might be able to use the same technique we currently use for pre-RCT outputs, I'll have to look into it more

storage/database/src/backend/heed/env.rs
```diff
@@ -268,12 +264,16 @@ where
...
-            db: self
-                .open_database(tx_ro, Some(T::NAME))?
-                .ok_or(RuntimeError::TableNotFound)?,
-            tx_ro,
-        })
+        //
+        // INVARIANT:
+        // We only care about setting the flags and/or comparison functions
+        // on write tables, read tables can't/don't add data so it doesn't matter.
+        // The sorted data from write tables will appear for read tables regardless.
```
Boog900, line 271: Uhmm is this defiantly correct? 

How is LMDB supposed to search the data correctly without the correct comparison function?

storage/database/src/key.rs
```diff
@@ -1,54 +1,175 @@
...
+///
+/// See [`Key`] for more info.
+#[derive(Default, Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum KeyCompare {
+    /// [Lexicographical comparison](https://doc.rust-lang.org/1.79.0/std/cmp/trait.Ord.html#lexicographical-comparison),
+    /// i.e. a straight byte comparison.
+    ///
+    /// This is the default.
+    #[default]
+    Lexicographic,
```
Boog900, line 146: I think it might be better to call this `Default`, as lexicographic still puts a requirement on backends.

If this was instead `Default` we can just use any method for the backend, whichever is more performant.

Not really an issue now as no backend actually has another default than lexicographic.

storage/database/src/storable.rs
Boog900: Probably alright to stay.

storage/database/src/key.rs
Boog900: Yeah it's probably better to do it this new way

storage/database/src/backend/heed/env.rs
```diff
@@ -282,10 +282,32 @@ where
...
+                // use this LMDB flag; it is ~10% faster.
+                KeyCompare::Number => {
+                    db.flags(DatabaseFlags::INTEGER_KEY);
+                }
+
+                // Use a custom comparison function if specified.
+                KeyCompare::Custom(_) => {
+                    db.key_comparator::<T::Key>();
+                }
+            }
```
hinto-janai, line 305: @Boog900 There is a problem with this.

Since the comparator function is baked into heed's database _type_, setting a different one means we're returning a different type - we can't do this since `-> impl DatabaseRw<T>` must be _one_ type.

I.e. in the first 2 branches we're returning:
```rust
heed::Database<K, V, heed::DefaultComparator>
```
but in the last, we're returning:
```rust
//            implements heed::Comparator
//                   v
heed::Database<K, V, StorableHeed<Key>>
```

Workarounds I've thought about so far:

Workaround: always call `key_comparator`, the `heed::Comparator` impl forwards to the correct fn at compile time
Con: ~10% slower DB writes, even the default. I believe it's because LMDB has to call our (un-inline-able?) function instead of using its built-in one

Workaround: Only support `u32, u64, usize` for value comparisons
Con: No custom comparison for anything other than those

Workaround (best IMO): Document an invariant that the new `create_db` _must_ be called to "set" the flags / comparison fns. After that, `open_db_r{o,w}` will be like before this PR - open with no flags or comparison function, assume they were set with `create_db`.
Con: Users must follow the invariant for correct behavior.

The last one works since `create_db` doesn't return the DB, so no type worries.

storage/database/src/backend/heed/env.rs
Boog900: Ah ok, LMDB caches the comparison function, that makes sense, thanks.

Users would be required to create these tables/ open them rw on startup for these flags to be set (I forgot if that is already a requirement or if we dropped that), with your other comment though this is already needed.

storage/database/src/backend/heed/env.rs
Boog900: Ah ok, yeah I like the the last option best as well

> Con: ~10% slower DB writes,

LMDB still needs to call the comparison function for reads, so reads should be slower as well, probably more so as the comparison function takes up a bigger percent of the total work for reads compared to writes.

storage/database/src/backend/heed/env.rs
Boog900: I would update the comment to include that LMDB is caching the comparison function

storage/database/src/backend/heed/env.rs
```diff
@@ -282,15 +283,44 @@ where
...
+
+            // Use a custom comparison function if specified.
+            KeyCompare::Custom(_) => {
+                db.key_comparator::<StorableHeed<T::Key>>()
+                    .create(&mut tx_rw)?;
+            }
+        }
+
         Ok(())
     }
```
hinto-janai, line 325: `EnvInner::create_db` is now responsible for setting flags / comparison behavior, this invariant is documented...

storage/database/src/env.rs
```diff
@@ -204,7 +202,13 @@ Implementation detail you should NOT rely on:
...
 /// they must be created first or else it will return error.
 ///
-/// See [`EnvInner::open_db_rw`] and [`EnvInner::create_db`] for creating tables.
+/// Note that when opening tables with [`EnvInner::open_db_ro`],
+/// they must be created first or else it will return error.
+///
+/// See [`EnvInner::create_db`] for creating tables.
+///
+/// # Invariant
+#[doc = doc_heed_create_db_invariant!()]
```
hinto-janai, line 211: ...here on `EnvInner`, `open_db_r{o,w}`, and `create_db`.

storage/database/src/backend/redb/env.rs
```diff
@@ -189,7 +189,10 @@ where
         // 3. So it's not being used to open a table since that needs `&tx_rw`
         //
         // Reader-open tables do not affect this, if they're open the below is still OK.
-        redb::WriteTransaction::delete_table(tx_rw, table)?;
+        if !redb::WriteTransaction::delete_table(tx_rw, table)? {
+            return Err(RuntimeError::TableNotFound);
+        }
```
hinto-janai, line 194: Unrelated: fixed `redb`'s `clear_db` behavior. It is supposed to error if the table didn't exist, test added...

storage/database/src/backend/tests.rs
```diff
@@ -156,6 +156,20 @@ fn non_manual_resize_2() {
...
+fn clear_db_table_not_found() {
+    let (env, _tmpdir) = tmp_concrete_env();
+    let env_inner = env.env_inner();
+    let mut tx_rw = env_inner.tx_rw().unwrap();
+    let err = env_inner.clear_db::<TestTable>(&mut tx_rw).unwrap_err();
+    assert!(matches!(err, RuntimeError::TableNotFound));
+
+    env_inner.create_db::<TestTable>(&tx_rw).unwrap();
+    env_inner.clear_db::<TestTable>(&mut tx_rw).unwrap();
+}
```
hinto-janai, line 171: ...here.

storage/database/src/key.rs
```diff
@@ -1,54 +1,180 @@
...
+/// 2. Implement [`Key`] on that wrapper
+/// 3. Define a custom [`Key::KEY_COMPARE`]
+// FIXME:
+// implement getting values using ranges.
 // <https://github.com/Cuprate/cuprate/pull/117#discussion_r1589378104>
-pub trait Key: Storable + Sized {
-    /// The primary key type.
-    type Primary: Storable;
-
+pub trait Key: Storable + Sized + Ord {
```
hinto-janai, line 33: `Key` must now be `Ord` because `heed`'s `create_db` may call:
- `heed::Comparator::compare` which may call...
- `KeyCompare::as_compare_fn` which may call...
- `Ord::cmp()` for value comparison

Thankfully this is already the case for all our keys but is this going to be a problem in the future? Is there a clearer way to set these bounds?

Boog900: I don't think so, it kinda makes sense for `Key` to require `Ord`

storage/database/src/key.rs
```diff
@@ -1,54 +1,180 @@
...
+///
+/// This default is overridden for numbers, which use a number comparison.
+/// For example, [`u64`] keys are sorted as `{0, 1, 2 ... 999_998, 999_999, 1_000_000}`.
+///
+/// If you would like to re-define this for number types, consider;
+/// 1. Creating a wrapper type around primitives like a `struct SortU8(pub u8)`
+/// 2. Implement [`Key`] on that wrapper
+/// 3. Define a custom [`Key::KEY_COMPARE`]
+// FIXME:
+// implement getting values using ranges.
```
Boog900, line 31: We have this already right? https://github.com/Cuprate/cuprate/blob/7e9891de5b54f9caf11fd1ed57a5c45daeb57a11/storage/database/src/database.rs#L48

```suggestion
```

Boog900 (approved):

PR #203, opened 2024-06-29
ci: add `doc.yml`

hinto-janai:
### What
Uploads `cargo doc` to this repo's GitHub pages.

This allows pre-built, public viewing of Cuprate crate documentation on `main`, example:
- https://github.com/hinto-janai/docs
- https://hinto-janai.github.io/docs
- https://hinto-janai.github.io/docs/cuprate_blockchain/index.html

We can disable the dependencies showing but this unfortunately disables links, e.g.:
```
[`tower::Service`]
```
no longer links. Showing the list of dependencies is not bad either IMO. It makes the list bigger but all of our crates start with `cuprate_` so they are easy to spot. 

This is in this repo unlike our books because code will be updated frequently. It would be painful to manually deploy changes.

### Previous work
- https://github.com/paradigmxyz/reth/blob/d1efe2b19b3c69b5c534d0afaa5fe3eaf7d53a92/.github/workflows/book.yml
- https://reth.rs/docs
- https://doc-internal.zebra.zfnd.org/zebrad

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,0 +1,74 @@
+# This builds `cargo doc` and uploads it to the repo's GitHub Pages.
+
+name: Doc
+
+on:
+  push:
+    branches: [ "main" ] # Only deploy if `main` changes.
+  workflow_dispatch:
+
+env:
+  # Show colored output in CI.
+  CARGO_TERM_COLOR: always
+  # Fail on documentation warnings, generate an index page.
+  RUSTDOCFLAGS: '-D warnings --cfg docsrs --show-type-layout --enable-index-page -Zunstable-options'
+
+jobs:
+  # Build documentation.
+  build:
+    # FIXME: how to build and merge Windows + macOS docs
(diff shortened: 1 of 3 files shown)


hinto-janai:
@Boog900 This will need a DNS entry for `doc.cuprate.org`. Pages on this repo needs to be set-up as well.

Boog900:
This is a really good idea!

However I don't like the index page, we can make a custom one with markdown but it needs to be manually updated and the style is incorrect. Is there any other option there?

hinto-janai:
We can do it `zebrad` style, where the main `index.html` symlinks to `zebrad`'s crate documentation, which acts as a general guide.

Or, we could hijack the generated `index.html` with our own text? A bit hacky but:
```bash
sed -i "s/List of all crates/$OUR_TEXT/" index.html
```

Boog900:
Yeah, probably something to look at for the future, when we have a `cuprated` binary, for now this isn't too bad.

Boog900 (approved):

Issue #177, opened 2024-06-19
Dandelion with no outbound connections will freeze.


Boog900:
We should provide a way for the `outbound_peer_discover` to tell us there are currently not enough peers.

relevant code: 

https://github.com/Cuprate/cuprate/blob/b76042a4e4f364aa2c3b53235f30382ac632d2e8/p2p/dandelion/src/router.rs#L164-L179

PR #204, opened 2024-06-30
P2P: Fix freeze in D++

Boog900:
### What

Fixes: #177 

This prevent freezing in the d++ router by changing the peer discover to a stream that returns an enum with a variant to tell the router there are no more outbound peers.

### DIFF
--- a/p2p/dandelion-tower/src/lib.rs
+++ b/p2p/dandelion-tower/src/lib.rs
@@ -26,9 +26,9 @@
 //! The diffuse service should have a request of [`DiffuseRequest`](traits::DiffuseRequest) and it's error
 //! should be [`tower::BoxError`].
 //!
-//! ## Outbound Peer Discoverer
+//! ## Outbound Peer TryStream
 //!
-//! The outbound peer [`Discover`](tower::discover::Discover) should provide a stream of randomly selected outbound
+//! The outbound peer [`TryStream`](futures::TryStream) should provide a stream of randomly selected outbound
 //! peers, these peers will then be used to route stem txs to.
 //!
 //! The peers will not be returned anywhere, so it is recommended to wrap them in some sort of drop guard that returns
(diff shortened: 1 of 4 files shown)


p2p/dandelion-tower/src/lib.rs
```diff
@@ -37,10 +37,10 @@
 //! ## Peer Service
 //!
 //! This service represents a connection to an individual peer, this should be returned from the Outbound Peer
-//! Discover. This should immediately send the transaction to the peer when requested, i.e. it should _not_ set
+//! TryStream. This should immediately send the transaction to the peer when requested, it should _not_ set
 //! a timer.
 //!
-//! The diffuse service should have a request of [`StemRequest`](traits::StemRequest) and it's error
+//! The peer service should have a request of [`StemRequest`](traits::StemRequest) and it's error
```
hinto-janai, line 43:
```suggestion
//! The peer service should have a request of [`StemRequest`](traits::StemRequest) and its error
```

p2p/dandelion-tower/src/router.rs
```diff
@@ -165,27 +171,41 @@ where
...
-                Change::Insert(key, svc) => {
+                OutboundPeer::Peer(key, svc) => {
                     self.stem_peers.insert(key, svc);
                 }
-                Change::Remove(key) => {
-                    self.stem_peers.remove(&key);
+                OutboundPeer::Exhausted => {
+                    tracing::warn!("Failed to retrieve enough outbound peers for optimal dandelion++, privacy may be degraded.");
+                    return Poll::Ready(Ok(()));
                 }
```
hinto-janai, line 184: This seems... not okay? What does `monerod` here?

Boog900: monerod does the same thing, there isn't really much more we can do here, if we have no outbound peers it is better for transactions to get to the network with less privacy than to be dropped completely.

Even if you fluff immediately you still have some sort of "herd immunity", it shouldn't be easy to tell this was the case.

hinto-janai (approved):

PR #206, opened 2024-07-01
Storage: add blockchain history requests

Boog900:
### What

Adds the DB requests and handlers for blockchain history, this history is sent to peers so they can send us their chain to sync from.

### How

Uses the same technique as monerod, choosing the first 11 blocks from the top of the chain and then power of 2 offsets from there.

### DIFF
--- a/storage/blockchain/src/service/free.rs
+++ b/storage/blockchain/src/service/free.rs
@@ -59,10 +59,10 @@ pub(crate) fn compact_history_genesis_not_included<const INITIAL_BLOCKS: u64>(
     // Otherwise, we use the fact that to reach the genesis block this statement must be true (for a
     // single `i`):
     //
-    // `chain_height - INITIAL_BLOCKS - 2^i + 2 = 0`
+    // `top_block_height - INITIAL_BLOCKS - 2^i + 2 == 0`
     // which then means:
-    // `chain_height - INITIAL_BLOCKS + 2 = 2^i`
-    // So if `chain_height - INITIAL_BLOCKS + 2` is a power of 2 then the genesis block is in
+    // `top_block_height - INITIAL_BLOCKS + 2 == 2^i`
+    // So if `top_block_height - INITIAL_BLOCKS + 2` is a power of 2 then the genesis block is in
     // the compact history already.
     top_block_height > INITIAL_BLOCKS && !(top_block_height - INITIAL_BLOCKS + 2).is_power_of_two()
 }
@@ -71,13 +71,13 @@ pub(crate) fn compact_history_genesis_not_included<const INITIAL_BLOCKS: u64>(
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-
     use proptest::prelude::*;
+
+    use super::*;
 
     proptest! {
         #[test]
-        fn compact_history_always_includes_genesis(top_height in 0_u64..500_000_000) {
+        fn compact_history(top_height in 0_u64..500_000_000) {
             let mut heights = (0..)
                 .map(compact_history_index_to_height_offset::<11>)
                 .map_while(|i| top_height.checked_sub(i))
@@ -87,6 +87,7 @@ mod tests {
                 heights.push(0);
             }
 
+            // Make sure the genesis and top block are always included.
             assert_eq!(*heights.last().unwrap(), 0);
             assert_eq!(*heights.first().unwrap(), top_height);
 
--- a/storage/blockchain/src/service/free.rs
+++ b/storage/blockchain/src/service/free.rs
@@ -33,8 +33,64 @@ pub fn init(config: Config) -> Result<(DatabaseReadHandle, DatabaseWriteHandle),
     Ok((readers, writer))
 }
 
+/// Given a position in the compact history, returns the height offset that should be in that position.
+///
+/// The height offset is the difference between the top block's height and the block height that should be in that position.
+#[inline]
+pub(crate) fn compact_history_index_to_height_offset<const INITIAL_BLOCKS: u64>(i: u64) -> u64 {
+    // If the position is below the initial blocks just return the position back
+    if i <= INITIAL_BLOCKS {
+        i
+    } else {
+        // Otherwise we go with power of 2 offsets, the same as monerod.
+        // So (INITIAL_BLOCKS + 2), (INITIAL_BLOCKS + 2 + 4), (INITIAL_BLOCKS + 2 + 4 + 8)
+        INITIAL_BLOCKS + (2 << (i - INITIAL_BLOCKS)) - 2
+    }
+}
+
+/// Returns if the genesis block was _NOT_ included when calculating the height offsets.
+///
+/// The genesis must always be included in the compact history.
+#[inline]
+pub(crate) fn compact_history_genesis_not_included<const INITIAL_BLOCKS: u64>(
+    top_block_height: u64,
+) -> bool {
+    // If the top block height is less than the initial blocks then it will always be included.
+    // Otherwise, we use the fact that to reach the genesis block this statement must be true (for a
+    // single `i`):
+    //
+    // `chain_height - INITIAL_BLOCKS - 2^i + 2 = 0`
+    // which then means:
+    // `chain_height - INITIAL_BLOCKS + 2 = 2^i`
+    // So if `chain_height - INITIAL_BLOCKS + 2` is a power of 2 then the genesis block is in
+    // the compact history already.
+    top_block_height > INITIAL_BLOCKS && !(top_block_height - INITIAL_BLOCKS + 2).is_power_of_two()
+}
+
 //---------------------------------------------------------------------------------------------------- Tests
+
 #[cfg(test)]
-mod test {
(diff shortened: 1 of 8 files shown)


Boog900:
@hinto-janai this is ready, for reference here is the function in monerod that gets the compact history: https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/cryptonote_core/blockchain.cpp#L727

storage/blockchain/src/service/free.rs
```diff
@@ -33,8 +33,65 @@ pub fn init(config: Config) -> Result<(DatabaseReadHandle, DatabaseWriteHandle),
...
+    }
+}
+
+/// Returns if the genesis block was _NOT_ included when calculating the height offsets.
+///
+/// The genesis must always be included in the compact history.
+#[inline]
+pub(crate) fn compact_history_genesis_not_included<const INITIAL_BLOCKS: u64>(
+    top_block_height: u64,
+) -> bool {
```
hinto-janai, line 57:
```suggestion
pub(super) const fn compact_history_genesis_not_included<const INITIAL_BLOCKS: u64>(
    top_block_height: u64,
) -> bool {
```

storage/blockchain/src/service/read.rs
```diff
@@ -525,3 +530,60 @@ fn key_images_spent(env: &ConcreteEnv, key_images: HashSet<KeyImage>) -> Respons
...
+            Ok(height) => last_height = height,
+            Err(RuntimeError::KeyNotFound) => {
+                return Ok(BCResponse::FindFirstUnknown(Some((i, last_height + 1))))
+            }
+            Err(e) => return Err(e),
+        }
+    }
+
+    Ok(BCResponse::FindFirstUnknown(None))
+}
```
hinto-janai, line 589: Same here, what is `block_ids.len()` typically going to be?

types/src/blockchain.rs
```diff
@@ -83,10 +83,16 @@ pub enum BCReadRequest {
     /// The input is a list of output amounts.
     NumberOutputsWithAmount(Vec<u64>),
 
-    /// Check that all key images within a set arer not spent.
+    /// Check that all key images within a set are not spent.
     ///
     /// Input is a set of key images.
     KeyImagesSpent(HashSet<[u8; 32]>),
+
+    /// A request for the compact chain history
```
hinto-janai, line 91:
```suggestion
    /// A request for the compact chain history.
```

types/src/blockchain.rs
```diff
@@ -164,6 +170,21 @@ pub enum BCResponse {
...
+        /// These blocks should be in reverse chronological order, not every block is needed.
+        block_ids: Vec<[u8; 32]>,
+        /// The current cumulative difficulty of the chain.
+        cumulative_difficulty: u128,
+    },
+
+    /// The response for [`BCReadRequest::FindFirstUnknown`].
+    ///
+    /// Contains the index of the first unknown block and its expected height.
+    FindFirstUnknown(Option<(usize, u64)>),
```
hinto-janai, line 186:
```suggestion
    /// The response for [`BCReadRequest::FindFirstUnknown`].
    ///
    /// Contains the index of the first unknown block and its expected height.
    ///
    /// This will be [`None`] if all blocks were known.
    FindFirstUnknown(Option<(usize, u64)>),
```

hinto-janai (commented):
Unrelated: there's a typo in `storage/blockchain/src/service/read.rs`, `filter_unknown_hahses` -> `filter_unknown_hashes`.

Also maybe link to the compact history reference in `types`, and in `compact_history_index_to_height_offset`.

storage/blockchain/src/service/read.rs
```diff
@@ -525,3 +530,60 @@ fn key_images_spent(env: &ConcreteEnv, key_images: HashSet<KeyImage>) -> Respons
...
+    );
+
+    /// The amount of top block IDs in the compact chain.
+    const INITIAL_BLOCKS: u64 = 11;
+
+    let mut block_ids = (0..)
+        .map(compact_history_index_to_height_offset::<INITIAL_BLOCKS>)
+        .map_while(|i| top_block_height.checked_sub(i))
+        .map(|height| Ok(get_block_info(&height, &table_block_infos)?.block_hash))
+        .collect::<Result<Vec<_>, RuntimeError>>()?;
```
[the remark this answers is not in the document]
Boog900: no the amount of block IDs to lookup is very small

storage/blockchain/src/service/read.rs
Boog900: Max is 25,000 which now that I say it is a lot ... I'll change this to a binary search.

Boog900:
I fixed the typo and added a link to monero core in `compact_history_index_to_height_offset`

storage/blockchain/src/service/read.rs
```diff
@@ -525,3 +530,75 @@ fn key_images_spent(env: &ConcreteEnv, key_images: HashSet<KeyImage>) -> Respons
...
+        block_ids.partition_point(
+            |block_id| match block_exists(block_id, &table_block_heights) {
+                Ok(exists) => exists,
+                Err(e) => {
+                    err.get_or_insert(e);
+                    // if this happens the search is scrapped, just return `false` back.
+                    false
+                }
+            },
+        );
```
hinto-janai, line 589: I'm not sure this is correct? Sorting is not enforced or documented either.

AFAICT this binary search requires the input slice to be partitioned like so:
```
[exists, exists, ..., doesnt_exist, doesnt_exist]
```
but:
1. The input `block_ids: &[BlockHash]` isn't enforced to be sorted like this
2. There is no sort order for "exists", if there were, the caller would know the index already

For example:
```rust
fn main() {
    let x = [0; 32]; // known hash
    let y = [1; 32]; // unknown hash

    let block_ids = [x, y, x, x];
    let blockchain = [x];

    // should return 1
    let idx = block_ids.partition_point(|id| blockchain.contains(id));
    assert_eq!(idx, 4); // hmm
}
```

storage/blockchain/src/service/read.rs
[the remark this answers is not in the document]
hinto-janai: Maybe add this comment in the function:

> // `rayon` is not used here because the amount of block IDs is expected to be small.

storage/blockchain/src/service/read.rs
Boog900: Yeah I should probably document this but the inputted block hashes must be in chronological order, so if we don't know a block we don't know any of its children either.

A peer could send a bogus chain but there are other ways to do that as well and the peer will be banned,

storage/blockchain/src/service/read.rs
```diff
@@ -525,3 +530,75 @@ fn key_images_spent(env: &ConcreteEnv, key_images: HashSet<KeyImage>) -> Respons
...
+    }
+
+    Ok(BCResponse::CompactChainHistory {
+        cumulative_difficulty,
+        block_ids,
+    })
+}
+
+/// [`BCReadRequest::FindFirstUnknown`]
+fn find_first_unknown(env: &ConcreteEnv, block_ids: &[BlockHash]) -> ResponseResult {
```
hinto-janai, line 570:
```suggestion
/// [`BCReadRequest::FindFirstUnknown`]
///
/// # Invariant
/// `block_ids` must be sorted in chronological block order, or else
/// the returned result is unspecified and meaningless, as this function
/// performs a binary search.
fn find_first_unknown(env: &ConcreteEnv, block_ids: &[BlockHash]) -> ResponseResult {
```

types/src/blockchain.rs
```diff
@@ -83,10 +83,16 @@ pub enum BCReadRequest {
...
+    /// Check that all key images within a set are not spent.
     ///
     /// Input is a set of key images.
     KeyImagesSpent(HashSet<[u8; 32]>),
+
+    /// A request for the compact chain history.
+    CompactChainHistory,
+
+    /// A request to find the first unknown block ID in a list of block IDs.
+    FindFirstUnknown(Vec<[u8; 32]>),
```
hinto-janai, line 95:
```suggestion
    /// A request to find the first unknown block ID in a list of block IDs.
    ////
    /// # Invariant
    /// The [`Vec`] containing the block IDs must be sorted in chronological block
    /// order, or else the returned response is unspecified and meaningless,
    /// as this request performs a binary search.
    FindFirstUnknown(Vec<[u8; 32]>),
```

hinto-janai (changes requested):

types/src/blockchain.rs
hinto-janai: Also the binary search isn't guaranteed to return the _first_ unknown block hash, right? Maybe this should be renamed.

Boog900: It finds the partition point between the known and the unknown blocks, so it does return the first

hinto-janai (approved):

hinto-janai:
Wait actually, add this please :)

https://github.com/Cuprate/cuprate/pull/206#discussion_r1663189153

PR #210, opened 2024-07-02
rpc: add feature flags

hinto-janai:
### What
Adds and enforces these feature flags in `cuprate-rpc-types`:

| Feature flag | Does what |
|--------------|-----------|
| `json`       | Enables the `crate::json` module
| `bin`        | Enables the `crate::bin` module
| `other`      | Enables the `crate::other` module
| `serde`      | Implements `serde` on all types
| `epee`       | Implements `epee_encoding` on all types

All are enabled by default.

### DIFF
--- a/rpc/types/Cargo.toml
+++ b/rpc/types/Cargo.toml
@@ -9,14 +9,19 @@ repository  = "https://github.com/Cuprate/cuprate/tree/main/rpc/types"
 keywords    = ["cuprate", "rpc", "types", "monero"]
 
 [features]
-default = []
+default = ["json", "bin", "other", "serde", "epee"]
+json    = []
+bin     = []
+other   = []
+serde   = ["dep:serde"]
+epee    = ["dep:cuprate-epee-encoding"]
 
 [dependencies]
-cuprate-epee-encoding = { path = "../../net/epee-encoding" }
+cuprate-epee-encoding = { path = "../../net/epee-encoding", optional = true }
 
 monero-serai = { workspace = true }
 paste        = { workspace = true }
-serde        = { workspace = true }
+serde        = { workspace = true, optional = true }
 
 [dev-dependencies]
 serde_json = { workspace = true }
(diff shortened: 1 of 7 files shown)


Boog900:
Would it be better to just have `serde` and `epee` be feature flags and enable all types that those feature allow.

-  `serde` enables JSON
-  `epee` enables bin
- types in other are enabled on a per type basis?

hinto-janai:
> Would it be better to just have serde and epee be feature flags and enable all types that those feature allow.

I think it's standard to keep these separate, i.e. modules have flags, (de)serialization have flags.

> types in other are enabled on a per type basis

There's 27 `other` types... that's a lot of feature flags.

Boog900:
> I think it's standard to keep these separate, i.e. modules have flags, (de)serialization have flags.

This crate is just types, and the whole point of these types are for ser/de. For example I can't see a reason to have the bin types without epee?

> There's 27 other types... that's a lot of feature flags.

My bad, I meant enable the JSON other types with `serde` enabled and the bin + JSON other types if both `serde` and `epee` are enabled.

hinto-janai:
> This crate is just types, and the whole point of these types are for ser/de. For example I can't see a reason to have the bin types without epee?

I realize this but having a feature called `serde` do anything other than enable `serde` is unexpected IMO. I'll change it if wanted but I think the current set-up is more clear, it's more modular at no cost.

Boog900:
It might be a good idea to enable all types all the time and just feature flag epee and serde?

hinto-janai:
> It might be a good idea to enable all types all the time

That's... against the point of the feature flags :D

E.g. if someone only needs JSON-RPC, they only set `json, serde` and can ignore the rest (maybe this applies to audits as well).

Boog900:
> That's... against the point of the feature flags :D

I think enabling some type definitions is not too bad, the main issue is the epee dependency.

> E.g. if someone only needs JSON-RPC, they only set json, serde and can ignore the rest (maybe this applies to audits as well).

You could just ignore certain modules?

I don't want to delay this too much longer though, if you think the extra feature flags are worth it.

hinto-janai:
There's a lot of misc types (not in this PR) - those under features is probably too fine-grain, so `serde` + `epee` feature flags + enable all types is probably okay.

hinto-janai:
Closing in favor of https://github.com/Cuprate/cuprate/pull/218.

PR #213, opened 2024-07-03
ci: fix `doc.yml`

hinto-janai:
### What
- Removes non-existent feature usage from `cuprate-types`
- Restore optional `serde` feature in `cuprate-blockchain` (error not shown below, but shows up if `cargo doc` is ran manually)

Fixes https://github.com/Cuprate/cuprate/actions/runs/9785148434/job/27017556001 (probably).

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -10,8 +10,8 @@ on:
 env:
   # Show colored output in CI.
   CARGO_TERM_COLOR: always
-  # Fail on documentation warnings, generate an index page.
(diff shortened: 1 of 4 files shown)


hinto-janai:
@Boog900 I think these warnings only show up here since they're part of the nightly `cargo doc`. We could turn `-D warnings` off in `doc.yml` since it's already enabled in `ci.yml`, what do you think?

Boog900:
Yeah probably ok to remove that

Boog900 (approved):

PR #215, opened 2024-07-05
Storage: fix lifetimes

Boog900:
### What

The `EnvInner` and tx lifetime relation was not correct, while these types are concrete rust can infer some things that it can't if these types are kept generic, which is why this hasn't been a problem.


Boog900:
@hinto-janai  this is ready, I also fixed redb with `cuprate-blockchain`, we were setting the default features which _always_enables heed

PR #218, opened 2024-07-08
rpc: feature flags, macro changes, misc setup

hinto-janai:
### What
This PR:
- Makes changes to the base RPC types
- Adds a documentation macro the whole crate can use
- Adds the `default` and `free` (internal) module for default/conditional serde behavior
- Makes changes to the type generator macro
    - More maintenance friendly (sub-macros instead of duplicated branches)
    - Supports default field values
- Adds the `misc` module containing _every single_ misc type used in other request/response types

---

This is the 1st out of 4 PRs that implement types in `cuprate-rpc-types`.

There are 3 PRs built on-top of this one:
1. https://github.com/Cuprate/cuprate/pull/219
2. https://github.com/Cuprate/cuprate/pull/220
3. https://github.com/Cuprate/cuprate/pull/221

These respectively implement all:
- `/json_rpc` types
- `/*.bin` types
- Other JSON types


rpc/types/Cargo.toml
```diff
@@ -9,14 +9,16 @@ repository  = "https://github.com/Cuprate/cuprate/tree/main/rpc/types"
 keywords    = ["cuprate", "rpc", "types", "monero"]
 
 [features]
-default = []
+default = ["serde", "epee"]
+serde   = ["dep:serde"]
+epee    = ["dep:cuprate-epee-encoding"]
```
hinto-janai, line 14: All types enabled by default, only `serde` and `epee` are feature gated (enabled by default though).

rpc/types/README.md
```diff
@@ -10,13 +10,14 @@ This crate ports the types used in Monero's RPC interface, including:
...
-- [`other`] module; Misc JSON types from other endpoints
-
-Miscellaneous types are found in the root module, e.g. [`crate::Status`].
+| Module | Purpose |
+|--------|---------|
+| The root module | Miscellaneous items, e.g. constants.
+| [`json`] | Contains JSON request/response (some mixed with binary) that all share the common `/json_rpc` endpoint. |
+| [`bin`] | Contains request/response types that are expected to be fully in binary (`cuprate_epee_encoding`) in `monerod` and `cuprated`'s RPC interface. These are called at a custom endpoint instead of `/json_rpc`, e.g. `/get_blocks.bin`. |
+| [`other`] | Contains request/response types that are JSON, but aren't called at `/json_rpc` (e.g. [`crate::other::GetHeightRequest`]). |
+| [`misc`] | Contains miscellaneous types, e.g. [`crate::misc::Status`]. Many of types here are found and used in request/response types, for example, [`crate::misc::BlockHeader`] is used in [`crate::json::GetLastBlockHeaderResponse`]. |
```
hinto-janai, line 19: All misc types are now in the `misc` module instead of root.

rpc/types/src/base.rs
```diff
@@ -10,76 +10,30 @@
 //! - <https://github.com/monero-project/monero/commit/2899379791b7542e4eb920b5d9d58cf232806937>
 //! - <https://github.com/monero-project/monero/issues/8722>
 //! - <https://github.com/monero-project/monero/pull/8843>
+//!
+//! This library doesn't contain [`rpc_access_request_base`](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L114-L122)
+//! found in `monerod` as the type is practically deprecated.
+//!
+//! Although, [`AccessResponseBase`] still exists as to allow
+//! outputting the same JSON fields as `monerod` (even if deprecated).
```
hinto-janai, line 18: There is a `rpc_access_request_base` defined in `monerod` that pairs with the `rpc_access_response_base`.

This contains a `std::string: client` field, but since this is deprecated, I removed the implementation as we'll never use it - however - I kept `AccessResponseBase` as this actually affects the JSON output (i.e. it adds `top_hash: String`, `credits: u64`).

At the end of the day, the response is deprecated too but I thought following `monerod`'s JSON output more than not would be good. We can also remove this if wanted, they're always dead fields anyway.

rpc/types/src/base.rs
```diff
@@ -10,76 +10,30 @@
...
-    ($start:literal..=$end:literal) => {
-        concat!(
-            "[Definition](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L",
-            stringify!($start),
-            "-L",
-            stringify!($end),
-            ")."
-        )
-    };
-}
```
hinto-janai, line 33: Moved to `macros.rs` for the whole crate to use.

rpc/types/src/defaults.rs
```diff
@@ -0,0 +1,78 @@
+//! These functions define the default values
+//! of optional fields in request/response types.
+//!
+//! For example, [`crate::json::GetBlockRequest`]
+//! has a [`crate::json::GetBlockRequest::height`]
+//! field and a [`crate::json::GetBlockRequest::hash`]
+//! field, when the RPC interface reads JSON without
+//! `height`, it will use [`default_height`] to fill that in.
```
hinto-janai, line 8: These functions exist to pass to `serde` and `epee` for default field values.

Our own free functions are used over something like `Default::default` because:
- It's more explicit
- Sometimes we need the non-default, e.g. `default_bool_true`.

There's no example in this PR, but these are extensively used in the other PRs.

rpc/types/src/free.rs
```diff
@@ -0,0 +1,18 @@
+//! Free functions.
+
+//---------------------------------------------------------------------------------------------------- Serde
+// These are functions used for conditionally (de)serialization.
```
hinto-janai, line 4: Same as `defaults.rs`, this allows for conditional (de)serialization in some cases.

Examples are in the other PRs.

rpc/types/src/json.rs
```diff
@@ -3,10 +3,7 @@
...
 
 //---------------------------------------------------------------------------------------------------- Import
-use crate::{
-    base::{EmptyRequestBase, EmptyResponseBase, ResponseBase},
-    macros::define_request_and_response,
-};
+use crate::{base::ResponseBase, macros::define_request_and_response};
 
 //---------------------------------------------------------------------------------------------------- Struct definitions
 // This generates 2 structs:
```
hinto-janai, line 9: Any `{json,bin,other.rs}` changes can be ignored in this PR, they get implemented in the other PRs.

rpc/types/src/lib.rs
```diff
@@ -94,23 +102,25 @@
     )
 )]
 // TODO: remove me after finishing impl
-#![allow(dead_code)]
+#![allow(
+	dead_code,
+    rustdoc::broken_intra_doc_links // TODO: remove after `{bin,json,other}.rs` gets merged
```
hinto-janai, line 107: TODO: keep this unresolved until done.

rpc/types/src/macros.rs
```diff
@@ -17,35 +15,35 @@
...
+/// flatten that into the request type automatically.
 ///
-/// FIXME: there's probably a less painful way to branch here on input
-/// without having to duplicate 80% of the macro. Sub-macros were attempted
-/// but they ended up unreadable. So for now, make sure to fix the other
-/// branch as well when making changes. The only de-duplicated part is
-/// the doc generation with [`define_request_and_response_doc`].
+/// E.g. `Response {/*...*/}` and `ResponseBase {/*...*/}`
+/// would trigger the different branches.
 macro_rules! define_request_and_response {
```
hinto-janai, line 46: The type generator macro no longer has duplicated code, it only has 1 branch and uses sub-macros instead.

Since we got rid of the RPC access `Request` stuff, this macro _always_ takes `Request {}` as input now. No fields will create the `type Request = ()`, where fields will create a normal request.

Responses still use the RPC access syntax, so it's one of:
- `Request` - adds nothing (some requests only have `status`, so we need this)
- `ResponseBase` - adds `base: crate::base::ResponseBase`
- `AccessResponseBase` - adds `base: crate::base::AccessResponseBase`

rpc/types/src/macros.rs
```diff
@@ -59,177 +57,234 @@ macro_rules! define_request_and_response {
...
+        $( #[$type_attr:meta] )*
         $type_name:ident,
 
+        // The request type (and any doc comments, derives, etc).
+        $( #[$request_type_attr:meta] )*
+        Request {
+            // And any fields.
+            $(
+                $( #[$request_field_attr:meta] )*
+                $request_field:ident: $request_field_type:ty $(= $request_field_type_default:expr)?,
```
hinto-janai, line 71: And also, it supports `cuprate_epee_encoding`'s default value syntax.

We can find a way to automatically do this with `serde` too (right now the caller must add the serde attribute) but that's for another PR.

rpc/types/src/macros.rs
```diff
@@ -269,9 +323,34 @@ macro_rules! define_request_and_response_doc {
...
+            "#L",
+            stringify!($start),
+            $(
+                "-L",
+                stringify!($end),
+            )?
             ")."
         )
     };
 }
```
hinto-janai, line 355: This documents the original definition for pretty much every type in `cuprate-rpc-types`.

rpc/types/src/misc/binary_string.rs
```diff
@@ -1,14 +1,14 @@
-//! TODO
+//! JSON string containing binary data.
 
 //---------------------------------------------------------------------------------------------------- Import
 
 //---------------------------------------------------------------------------------------------------- BinaryString
-/// TODO
+/// TODO: we need to figure out a type that (de)serializes correctly, `String` errors with `serde_json`
```
hinto-janai, line 6: For another PR.

rpc/types/src/misc/block_complete_entry.rs
```diff
@@ -0,0 +1,37 @@
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct BlockCompleteEntry {
+    pub pruned: bool,
+    pub block: String,
+    pub block_weight: u64,
+    pub txs: Vec<TxBlobEntry>,
+}
+
+// TODO: custom epee
+// <https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/cryptonote_protocol/cryptonote_protocol_defs.h#L138-L163>
```
hinto-janai, line 29: Custom (de)serialization will be done in another PR.

rpc/types/src/misc/misc.rs
```diff
@@ -0,0 +1,539 @@
+//! Miscellaneous types.
+//!
+//! These are `struct`s that appear in request/response types.
+//! For example, [`crate::json::GetConnectionsResponse`] contains
+//! the [`crate::misc::ConnectionInfo`] struct defined here.
```
hinto-janai, line 5: TODO: I have to document this but `misc.rs` contains all misc types that can be created in a macro-friendly way.

The other files in the `misc/` folder are because they have something special, so the macro doesn't work.

hinto-janai (commented):
Things left that will be done in another PR after this + all type impl PRs are merged:
- Custom (de)serialization
- `Status` fixes (doesn't catch all strings, probably need `Other(String)`
- Field type optimizations
- Macro optimizations (less duplicated input, etc)

rpc/types/src/misc/misc.rs
```diff
@@ -0,0 +1,539 @@
...
+    #[doc = monero_definition_link!(
+        cc73fe71162d564ffda8e549b79a350bca53c454,
+        "rpc/core_rpc_server_commands_defs.h",
+        538..=553
+    )]
+    #[derive(Copy)]
+    /// Used in [`crate::bin::GetOutsRequest`].
+    OutKeyBin {
+        key: u8, // TODO: crypto::public_key,
+        mask: u8, // TODO: rct::key,
```
Boog900, line 395: These are `[u8; 32]`

rpc/types/src/misc/tx_entry.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub prunable_as_hex: String,
+    pub prunable_hash: String,
+    pub pruned_as_hex: String,
+    pub received_timestamp: u64,
+    pub relayed: bool,
+    pub tx_hash: String,
+}
+
+// TODO: custom epee
+// <https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/cryptonote_protocol/cryptonote_protocol_defs.h#L406-427>
```
Boog900, line 41: Is this the correct link?

Boog900 (changes requested):

rpc/types/src/misc/misc.rs
hinto-janai: These were going to be replace with some dedicated types and/or aliases as there's a few more of these (e.g. [`typedef std::string blobdata`](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/cryptonote_basic/blobdatatype.h#L39)), but maybe the type directly is okay?

rpc/types/src/misc/tx_entry.rs
hinto-janai: Oops, fixed to correct link (https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L406-L428).

rpc/types/src/base.rs
hinto-janai: Added back `AccessRequestBase`.

FYI the empty bases (`Request`, `Response`) are special-cased by the macro since flattening an empty struct causes some errors.

Boog900 (approved):

PR #219, opened 2024-07-08
rpc: implement `/json_rpc` types

hinto-janai:
### What
Ports all of `monerod`'s JSON types at the `/json_rpc` endpoint to `cuprate-rpc-types` in the `json` module. 

Requires https://github.com/Cuprate/cuprate/pull/218.


rpc/types/src/json.rs
```diff
@@ -86,6 +95,568 @@ define_request_and_response! {
...
+    /// let x = SubmitBlockResponse { status: String::from("asdf") };
+    /// let x = to_string(&x).unwrap();
+    /// assert_eq!(x, "\"asdf\"");
+    /// ```
+    #[cfg_attr(feature = "serde", serde(transparent))]
+    #[repr(transparent)]
+    Response {
+        status: String,
+    }
+}
```
Boog900, line 182: Is this type correct?

```cpp
  struct COMMAND_RPC_SUBMITBLOCK
  {
    typedef std::vector<std::string> request;
    
    struct response_t: public rpc_response_base
    {
      std::string block_id;

      BEGIN_KV_SERIALIZE_MAP()
        KV_SERIALIZE_PARENT(rpc_response_base)
        KV_SERIALIZE(block_id)
      END_KV_SERIALIZE_MAP()
    };
    typedef epee::misc_utils::struct_init<response_t> response;
  };

```

hinto-janai: No, I think I got confused. [`block_id` is documented as `status`](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#submit_block), are the docs wrong?

Boog900: `status` is in `rpc_response_base`,  `block_id` is a seperate field that was added here: https://github.com/monero-project/monero/commit/e8cac61f4b9a662cbc1b00e46d1f9a3dd991c5f0 I guess the docs were not updated so yes.

hinto-janai: OK, the output would be something like this, right?:
```json
{
  "result": {
    "status": "OK",
    "untrusted": false,
    "block_id": "<hash>"
  },
  "id": "0",
  "jsonrpc": "2.0"
}
```

Boog900: I think so yes

rpc/types/src/json.rs
```diff
@@ -86,6 +95,558 @@ define_request_and_response! {
...
+    /// use cuprate_rpc_types::json::*;
+    ///
+    /// let x = SubmitBlockRequest { block_ids: vec!["a".into(),"b".into()] };
+    /// let x = to_string(&x).unwrap();
+    /// assert_eq!(x, r#"["a","b"]"#);
+    /// ```
+    #[cfg_attr(feature = "serde", serde(transparent))]
+    #[repr(transparent)]
+    Request {
+        block_ids: Vec<String>,
```
Boog900, line 167: This is a block blob:

```suggestion
        // This is `std::vector<std::string>` in `monerod` but
        // it must be a 1 length array or else it will error.
        block_blob: [String; 1],
```

hinto-janai: Yup: https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server.cpp#L2222-L2234, the output only returns 1 block too so yeah.

Documentation unfortunately states:
> array of strings; list of block **blobs** which have been mined

FWIW I expect many optimizations like this to made to the types when the handlers are created and we see how the types are used.

Boog900 (approved):

PR #220, opened 2024-07-08
rpc: implement `.bin` types

hinto-janai:
### What
Ports all of `monerod`'s binary types at `.bin` endpoints to `cuprate-rpc-types` in the `bin` module. 

Requires https://github.com/Cuprate/cuprate/pull/218.


rpc/types/src/bin.rs
```diff
@@ -1,8 +1,121 @@
...
+    }
+}
+
+define_request_and_response! {
+    get_hashesbin,
+    cc73fe71162d564ffda8e549b79a350bca53c454 =>
+    core_rpc_server_commands_defs.h => 309..=338,
+    GetHashes,
+    Request {
+        block_ids: Vec<[u8; 32]>,
```
Boog900, line 70: In monerod this is serialized using the macro `KV_SERIALIZE_CONTAINER_POD_AS_BLOB`, which has slightly different behaviour than default. 

Becuase this is a `Vec<[u8; 32>` you would have to use `ByteArrayVec`, for other types like `Vec<u64>` you would have to use `ContainerAsBlob`

hinto-janai: OK, can this be done in another PR? This will need changes to the macro and other files since `serde` is impl'ed on all things, but neither `ByteArrayVec` or `ContainerAsBlob` implement it.

Either those implement `serde` or we remove it from the macro and add a derive on-top of all types in `{json,other}.rs` manually.

Boog900: `ByteArrayVec` could impl serde, `ContainerAsBlob` is done differently though, the type of the field is still `Vec<_>` it's used like this:

```rust
epee_object!(
    FluffyMissingTransactionsRequest,
    block_hash: ByteArray<32>,
    current_blockchain_height: u64,
    missing_tx_indices: Vec<u64> as ContainerAsBlob<u64>,
);
```

Boog900: but yeah this can be done in another PR

Boog900 (approved):

PR #226, opened 2024-07-10
fixed-bytes: add `serde`, document feature flags

hinto-janai:
### What
Adds a `serde` feature flag (enabled by default) to `cuprate-fixed-bytes` and documents it.

### Why
For https://github.com/Cuprate/cuprate/pull/220#discussion_r1672904392.

### DIFF
--- a/net/fixed-bytes/src/lib.rs
+++ b/net/fixed-bytes/src/lib.rs
@@ -8,7 +8,7 @@ use core::{
 use bytes::{BufMut, Bytes, BytesMut};
 
 #[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 #[cfg_attr(feature = "std", derive(thiserror::Error))]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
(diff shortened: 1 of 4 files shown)


hinto-janai:
@Boog900 For another PR: it seems backwards to have the `{ByteArray,ByteArrayVec}` epee implementations in `epee-encoding` itself, they should be in `fixed-bytes`; `epee-encoding` could expose some things so it's less cumbersome to do this.

net/fixed-bytes/src/lib.rs
```diff
@@ -43,6 +49,7 @@ impl Debug for FixedByteError {
 /// Internally this is just a wrapper around [`Bytes`], with the constructors checking that the length is equal to `N`.
 /// This implements [`Deref`] with the target being `[u8; N]`.
 #[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
```
Boog900, line 52: the `Serialize` is probably ok but the `Deserialize` has to make sure the `ByteArray` is a multiple of `N`

net/fixed-bytes/src/lib.rs
```diff
@@ -88,6 +95,7 @@ impl<const N: usize> TryFrom<Vec<u8>> for ByteArray<N> {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
```
Boog900, line 98: same here

Boog900:
@hinto-janai yeah true, it was from when the trait was sealed

Boog900 (approved):

PR #227, opened 2024-07-11
rpc: use `ByteArrayVec` and `ContainerAsBlob`

hinto-janai:
### What
- Uses `ByteArrayVec` and `ContainerAsBlob` for binary types: https://github.com/Cuprate/cuprate/pull/220#discussion_r1672904392
- Adds derives to `cuprate_fixed_bytes`
- Reduces default value duplication for type generator macro (kinda)

Requires https://github.com/Cuprate/cuprate/pull/226.


net/fixed-bytes/src/lib.rs
```diff
@@ -48,7 +49,7 @@ impl Debug for FixedByteError {
 ///
 /// Internally this is just a wrapper around [`Bytes`], with the constructors checking that the length is equal to `N`.
 /// This implements [`Deref`] with the target being `[u8; N]`.
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Default, Clone, Eq, PartialEq, PartialOrd, Ord, Hash)]
```
hinto-janai, line 52: Needed for the type generator macro. I don't think there's a reason not to have these anyway.

rpc/types/README.md
```diff
@@ -64,6 +64,20 @@ These mixed types are:
...
+
+Some fields within requests/responses are containers, but fixed in size.
+
+For example, [`crate::json::GetBlockTemplateResponse::prev_hash`] is always a 32-byte hash.
+
+In these cases, stack allocated types like `cuprate_fixed_bytes::StrArray`
+will be used instead of a more typical [`String`] for optimization reasons.
+
+-->
+
```
hinto-janai, line 80: @Boog900 re: https://github.com/Cuprate/cuprate/pull/147#discussion_r1654912900.

Do you want to make a `cuprate_fixed_bytes::StrArray(Bytes)` type? Or `Str<const N: usize>([u8; N])`?

If the latter, there's plenty of stack string crates, or we can do a straight copy+paste of [my impl](https://docs.rs/readable/latest/readable/str/struct.Str.html) (or a sub-set of it).

I'm also not sure the overhead of `Bytes` will be worth it in some of the types that are mostly create -> throw away.

rpc/types/src/bin.rs
```diff
@@ -91,15 +94,29 @@ define_request_and_response! {
...
+    core_rpc_server_commands_defs.h => 487..=510,
+    GetOutputIndexes,
+    #[derive(Copy)]
+    Request {
+        txid: [u8; 32],
+    },
+    AccessResponseBase {
+        o_indexes: Vec<u64> as ContainerAsBlob<u64>,
+    }
+}
```
hinto-janai, line 110: Not ideal: there are 2 definitions for these `as` types. If `epee`:
```rust
o_indexes: Vec<u64> as ContainerAsBlob<u64>,
```
else:
```rust
o_indexes: Vec<u64>,
```

rpc/types/src/json.rs
```diff
@@ -49,8 +49,33 @@ define_request_and_response! {
...
+        // The default function must be specified twice:
+        //
+        // 1. As an expression
+        // 2. As a string literal
+        //
+        // For example: `extra_nonce: String /* = default_string(), "default_string" */,`
+        //
+        // This is a HACK since `serde`'s default attribute only takes in
+        // string literals and macros (stringify) within attributes do not work.
+        extra_nonce: String /* = default_expression, "default_literal" */,
```
hinto-janai, line 69: The type generator macro now handles `serde` default values directly.

Not ideal: attributes take in a string literal, so we can't pass `stringify!($default)` to `serde`'s `default(...)`.

So we have to provide:
- An expression for `cuprate_epee_encoding`
- A string literal for `serde`

rpc/types/Cargo.toml
```diff
@@ -10,11 +10,12 @@ keywords    = ["cuprate", "rpc", "types", "monero"]
 
 [features]
 default = ["serde", "epee"]
-serde   = ["dep:serde"]
+serde   = ["dep:serde", "cuprate-fixed-bytes/serde"]
 epee    = ["dep:cuprate-epee-encoding"]
 
 [dependencies]
 cuprate-epee-encoding = { path = "../../net/epee-encoding", optional = true }
+cuprate-fixed-bytes   = { path = "../../net/fixed-bytes" }
```
hinto-janai, line 18: Not ideal: this always gets pulled in but is only used in `crate::bin`.

Boog900: yeah not great but fixed bytes isn't too big of a crate

rpc/types/README.md
Boog900: `Bytes` will not be efficient for parsing from json, it's only efficient is the underlying data is stored in a `Bytes` object and if the parser uses the `Bytes` functions.

So `StrArray` could be useful for epee, however in the p2p code I just use `Bytes` for strings, they aren't common.

I would want a stack string though, I would prefer using a crate, unless there is a good reason for using a homegrown one?

rpc/types/src/bin.rs
```diff
@@ -91,15 +94,29 @@ define_request_and_response! {
...
+#[cfg(feature = "epee")]
+define_request_and_response! {
+    get_o_indexesbin,
+    cc73fe71162d564ffda8e549b79a350bca53c454 =>
+    core_rpc_server_commands_defs.h => 487..=510,
+    GetOutputIndexes,
+    #[derive(Copy)]
+    Request {
+        txid: [u8; 32],
+    },
```
Boog900, line 106: If we really want (memory) efficient parsing this (and other `[u8; N]` in `bin`) can be made into a `ByteArray`

rpc/types/src/bin.rs
Boog900: yeah that's not great, if this is the only type then it's probably acceptable

rpc/types/src/json.rs
Boog900: hmm that is annoying, I can't think of a way to work around this atm though

rpc/types/src/bin.rs
hinto-janai: Can type optimizations be for a different PR? Also, the `cuprate_epee_encoding` blanket impls don't cover `Vec<ByteArray<N>>`.

rpc/types/README.md
hinto-janai: > I would prefer using a crate, unless there is a good reason for using a homegrown one?

1. IIRC there's no stack string crate that everyone uses, the closest I think would be [heapless](https://github.com/rust-embedded/heapless)
1. We'd be exposing a 3rd party crate in our API, not the worst but a bit awkward IMO
2. `epee_encoding` would stop working, we'd have to make a wrapper for impl

We could also enable features to swap certain types, e.g.:
```rust
cfg_if! {
    if #[cfg(feature = "heapless")] else {
        pub type String = heapless::String;
    } else {
        pub type String = String
    }
}
```
We could do this for `ByteArray` stuff too, i.e. just keep all types as lenient as possible but allow type optimizations via features.

Boog900: Fair looking at `heapless` it does seem to be doing a similar thing to your homegrown one, and it wouldn't be too much to maintain.

Are stings common in epee binary types for RPC? I would just use `Bytes` for bin types.

rpc/types/src/bin.rs
Boog900: yeah sure, btw this `Vec<ByteArray<N>>` should always be this `ByteArrayVec<N>`

Boog900 (approved):

PR #201, opened 2024-06-28
update monero-serai

Boog900:
### What

Updates to the new monero-serai API, fixing the issues that have come up.

related: #192

PR #229, opened 2024-07-12
rpc: custom epee for misc types

hinto-janai:
### What
Implement `cuprate_epee_encoding` manually for some types in `cuprate-rpc-types`.

Requires https://github.com/Cuprate/cuprate/pull/227.

### Why
These types have custom epee (de)serialization in `monerod`.


rpc/types/src/misc/tx_entry.rs
```diff
@@ -37,23 +37,118 @@ pub struct TxEntry {
...
+            write_field(self.confirmations, "confirmations", w)?;
+            write_field(self.block_timestamp, "block_timestamp", w)?;
+            write_field(self.output_indices, "output_indices", w)?;
+        } else {
+            write_field(self.relayed, "relayed", w)?;
+            write_field(self.received_timestamp, "received_timestamp", w)?;
+        }
+
+        Ok(())
+    }
```
hinto-janai, line 153: @Boog900 Can you add support for a custom `EpeeObject::write_fields` in the epee macro?

All the custom epee stuff need context from `self`, so the macro's custom `write_field` doesn't work. E.g. here we need access to `self.in_pool`, then serialize fields based off that. Original code: https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L242-L259.

This forces us to do manual epee which gets pretty messy because of all the related boilerplate.

rpc/types/src/misc/tx_entry.rs
```diff
@@ -37,23 +37,118 @@ pub struct TxEntry {
...
+}
+
+#[cfg(feature = "epee")]
+impl EpeeObject for TxEntry {
+    type Builder = __TxEntryEpeeBuilder;
+
+    fn number_of_fields(&self) -> u64 {
+        // TODO: this is either 12 or 10 depending on `self.in_pool`.
+        14
+    }
```
hinto-janai, line 126: How should this work? We sometimes serialize 10 fields, sometimes 12.

rpc/types/src/misc/block_complete_entry.rs
```diff
@@ -25,13 +29,91 @@ pub struct BlockCompleteEntry {
...
+            pruned: self.pruned.unwrap_or(false),
+            block: self.block.ok_or(ELSE)?,
+            block_weight: self.block_weight.unwrap_or(0),
+            txs: self.txs.ok_or(ELSE)?,
+        })
+    }
+}
+
+#[cfg(feature = "epee")]
+impl EpeeObject for BlockCompleteEntry {
```
Boog900, line 82: Sorry for not realizing this earlier but this type is already defined in `cuprate-wire`:

https://github.com/Cuprate/cuprate/blob/0910c0a231b4e0ef48b6cf0b9d7d48a44cc557a7/net/wire/src/p2p/common.rs#L223-L234

I don't really want to move it to `cuprate-types` but that may be the best option?

rpc/types/src/misc/tx_entry.rs
Boog900: This can be dynamic, however it could be less than 10, epee wont serialize empty container.

rpc/types/src/misc/tx_entry.rs
Boog900: I would make this type an enum of 2 structs (epee objects) and call `write_fields`/ `number_of_fields` on whatever object is being written

Boog900: Also is this type even a bin type or just JSON?

hinto-janai: Ah my bad, this is only JSON.

rpc/types/src/bin.rs
```diff
@@ -134,6 +115,396 @@ define_request_and_response! {
...
+    get_blocksbin,
+    cc73fe71162d564ffda8e549b79a350bca53c454,
+    core_rpc_server_commands_defs, h, 162, 262,
+)]
+///
+/// This response's variant depends upon [`PoolInfoExtent`].
+#[allow(dead_code, missing_docs)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum GetBlocksResponse {
```
hinto-janai, line 148: ## Custom epee 1/2
These data types require custom epee.

rpc/types/src/misc/distribution.rs
```diff
@@ -0,0 +1,298 @@
...
+    "rpc/core_rpc_server_commands_defs.h",
+    2468..=2508
+)]
+/// Used in [`crate::json::GetOutputDistributionResponse`].
+///
+/// This enum's variant depends upon the `binary` and `compress` fields.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(untagged))]
+pub enum Distribution {
```
hinto-janai, line 60: ## Custom epee 2/2
These data types require custom epee.

rpc/types/src/misc/distribution.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub base: Option<u64>,
+    pub distribution: Option<Vec<u64>>,
+    pub amount: Option<u64>,
+    pub compressed_data: Option<String>,
+    pub binary: Option<bool>,
+    pub compress: Option<bool>,
+}
+
+#[cfg(feature = "epee")]
+impl EpeeObjectBuilder<Distribution> for __DistributionEpeeBuilder {
```
hinto-janai, line 123: Sidenote: are the custom epee implementations correct in this PR?

Is `if $field.should_write() { write() }` required? Should it be `write_field`? or just `write`?

rpc/types/src/other.rs
```diff
@@ -68,7 +68,8 @@ define_request_and_response! {
         key_images: Vec<String>,
     },
     AccessResponseBase {
-        spent_status: Vec<u8>, // TODO: should be `KeyImageSpentStatus`.
+        /// FIXME: These are [`KeyImageSpentStatus`] in [`u8`] form.
+        spent_status: Vec<u8>,
```
hinto-janai, line 72: For later: `cuprate_epee_encoding` doesn't blanket impl `Vec<impl EpeeValue>` so although this really should be `Vec<KeyImageSpentStatus>`, it's `Vec<u8>` for now.

rpc/types/src/misc/tx_entry.rs
```diff
@@ -18,42 +20,127 @@ use cuprate_epee_encoding::{
...
+    type Builder = ();
+
+    fn number_of_fields(&self) -> u64 {
+        unreachable!()
+    }
+
+    fn write_fields<B: BufMut>(self, w: &mut B) -> error::Result<()> {
+        unreachable!()
+    }
 }
```
hinto-janai, line 146: Invariant documented in `lib.rs` - some types will be unreachable and/or not impl for wrong context, e.g. `TxEntry` here is only used in a JSON endpoint, so its epee impl just panics.

This is a workaround because the macro requires all types impl serde + epee.

rpc/types/src/macros.rs
```diff
@@ -16,11 +16,11 @@
...
 /// This macro uses:
-/// - [`__define_request`]
-/// - [`__define_response`]
-/// - [`__define_request_and_response_doc`]
+/// - [`define_request`]
+/// - [`define_response`]
+/// - [`define_request_and_response_doc`]
 ///
-/// # `__define_request`
+/// # `define_request`
```
hinto-janai, line 23: With custom epee impls, we actually need to use some of these macros individually elsewhere, so no `__` prefix.

rpc/types/src/misc/distribution.rs
```diff
@@ -0,0 +1,298 @@
...
+/// TODO: for handler code. This should already be provided in the field.
+#[doc = crate::macros::monero_definition_link!(
+    cc73fe71162d564ffda8e549b79a350bca53c454,
+    "rpc/core_rpc_server_commands_defs.h",
+    57..=72
+)]
+#[allow(clippy::needless_pass_by_value)] // TODO: remove after impl
+fn decompress_integer_array(array: Vec<u64>) -> Vec<u64> {
+    todo!()
+}
```
hinto-janai, line 46: `Distribution::CompressedBinary` has a `compressed_data` field that gets populated _during_ serialization: https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L2487.

I think this is a bit weird since this is a field in a data structure, it should be provided by the constructor themselves - i.e. I think the handler should handle this, otherwise we have to comment something like:

```markdown
# Invariant
Leave the `compressed_data` field empty, the serialization will handle it...?
```

rpc/types/src/misc/status.rs
```diff
@@ -86,17 +86,12 @@ pub enum Status {
...
-    /// The reason this isn't `Unknown(String)` is because that
-    /// disallows [`Status`] to be [`Copy`], and thus other types
-    /// that contain it.
-    #[cfg_attr(feature = "serde", serde(other))]
-    #[cfg_attr(feature = "serde", serde(rename = "UNKNOWN"))]
-    Unknown,
+    /// This exists to act as a catch-all for all of
+    /// `monerod`'s other strings it puts in the `status` field.
+    #[cfg_attr(feature = "serde", serde(rename = "OTHER"), serde(untagged))]
+    Other(String),
```
hinto-janai, line 94: `enum Status` now has a catch-all and is no longer `Copy`.

Turns out `monerod` emits a lot of `status` strings other than the ones listed...

rpc/types/src/misc/distribution.rs
Boog900: For the JSON version of this type we will need a custom function that writes this type out right?

I think you could use the `serde` from/into attributes to go between a type that has all fields and the type we expose to users.

However if you would rather the handler handle this then I am ok with that

rpc/types/src/misc/distribution.rs
```diff
@@ -0,0 +1,291 @@
...
+#[cfg_attr(feature = "serde", serde(untagged))]
+pub enum Distribution {
+    /// Distribution data will be (de)serialized as either JSON or binary (uncompressed).
+    Uncompressed {
+        start_height: u64,
+        base: u64,
+        /// TODO: this is a binary JSON string if `binary == true`.
+        ///
+        /// Considering both the `binary` field and `/get_output_distribution.bin`
+        /// endpoint are undocumented in the first place, Cuprate could just drop support for this.
```
Boog900, line 79: If this is used in the core wallet(s) then I would want to support them

rpc/types/src/misc/distribution.rs
Boog900: only `write_field` needs to be called, `should_write` is checked in that function.

Everything else looks ok

rpc/types/src/other.rs
Boog900: blanket impl `Vec<impl EpeeValue>` we can't do this until specialization btw.

rpc/types/src/misc/distribution.rs
Boog900, line 79: If this is used in the core wallet(s) then I would want to support them

rpc/types/src/misc/distribution.rs
hinto-janai: > For the JSON version of this type we will need a custom function that writes this type out right?

Probably yes, but because of the binary string, not the compression.

There is no JSON compressed type, since it must be `binary && compress`: https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L2481-L2504.

> However if you would rather the handler handle this then I am ok with that

Or maybe we could hide the (de)compression in the (de)serialization? I.e. there is no `compressed_data` field, only `distribution`. It is always uncompressed from the caller's perspective because the epee (de)serialization handles it internally.

rpc/types/src/bin.rs
```diff
@@ -134,6 +115,285 @@ define_request_and_response! {
...
+            start_height,
+            current_height,
+            output_indices,
+            daemon_time,
+            pool_info_extent
+        }
+
+        self.added_pool_txs = read_epee_value(r).ok().unwrap_or_default();
+        self.remaining_added_pool_txids = read_epee_value(r).ok().unwrap_or_default();
+        self.removed_pool_txids = read_epee_value(r).ok().unwrap_or_default();
```
Boog900, line 299: is this correct? these should be in `read_epee_field!` right?

hinto-janai: I think this was incorrect actually, since these fields are only _sometimes_ serialized (e.g. `struct GetBlocksResponsePoolInfoNone`), they won't always exist.

In that case, the fields are left as `None` where `finish()` will do the rest of the work, right?

Boog900: > they won't always exist

This should be fine in `read_epee_field` right? there is no requirement for the field the exist there.

`add_field` is called after every field name has been read, by putting these lines after `read_epee_field` we are tying to read these fields after each field has been read which isn't correct. 

> In that case, the fields are left as None where finish() will do the rest of the work, right?

This is what would happen if they were in  `read_epee_field`

rpc/types/src/misc/distribution.rs
Boog900: > Or maybe we could hide the (de)compression in the (de)serialization?

Yeah this is what I was getting at, up to you what you think would be better

rpc/types/src/bin.rs
```diff
@@ -291,13 +291,12 @@ impl EpeeObjectBuilder<GetBlocksResponse> for __GetBlocksResponseEpeeBuilder {
             current_height,
             output_indices,
             daemon_time,
-            pool_info_extent
+            pool_info_extent,
+            added_pool_txs,
+            remaining_added_pool_txids,
+            removed_pool_txids
```
hinto-janai, line 297: Fields were put back in `read_epee_field!`.

rpc/types/src/misc/distribution.rs
```diff
@@ -15,30 +15,39 @@ use cuprate_epee_encoding::{
...
-    todo!()
+fn decompress_integer_array(mut array: &[u8]) -> Vec<u64> {
+    let capacity = array.len();
+    let mut vec = Vec::<u64>::with_capacity(capacity);
+
+    while let Ok(varint) = read_varint(&mut array) {
+        vec.push(varint);
+    }
+
+    vec
```
hinto-janai, line 50: (De)compression functions.

Ref: https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h#L45-L72.

rpc/types/src/misc/distribution.rs
```diff
@@ -109,7 +118,16 @@ epee_object! {
...
+    #[cfg_attr(
+        feature = "serde",
+        serde(serialize_with = "serialize_distribution_as_compressed_data")
+    )]
+    #[cfg_attr(
+        feature = "serde",
+        serde(deserialize_with = "deserialize_compressed_data_as_distribution")
+    )]
+    #[cfg_attr(feature = "serde", serde(rename = "compressed_data"))]
+    pub distribution: Vec<u64>,
```
hinto-janai, line 130: `distribution` is now always uncompressed from the constructor's perspective.

It gets (de)compressed during (de)serialization in `epee/serde`.

rpc/types/src/misc/distribution.rs
Boog900: Monero has 2 varint formats, the one we are using is the epee varint, monerod seems to be using the consensus serialisation varint, which is defined in `monero-serai`: https://github.com/serai-dex/serai/blob/6f34c2ff77ef8a9f8839dbf63f35dcd6dc33f4b3/networks/monero/io/src/lib.rs#L60

Boog900: might need to leave this as a todo until #201

hinto-janai: Changed bodies to `todo!()` for now.

Boog900 (approved):

PR #230, opened 2024-07-16
types: add `BlockCompleteEntry`

hinto-janai:
### What
Moves `cuprate_p2p::common::{BlockCompleteEntry,TransactionBlobs,PrunedTxBlobEntry}` to `cuprate_types` and fixes imports in other crates.

### Why
https://github.com/Cuprate/cuprate/pull/229#discussion_r1676635223.


Boog900 (changes requested):
I think I would rather `monero-wire` reexport these types under `p2p/common` than require users to pull in `cuprate-types` as well

Boog900 (approved):

PR #231, opened 2024-07-17
test-utils: add `crate::rpc::data` module

hinto-janai:
### What
Adds a `crate::rpc::data` module to `cuprate-test-utils`.

This module contains raw RPC request/response data, for example:
```rust
const GET_BLOCK_TEMPLATE_REQUEST: &str = r#"{"jsonrpc":"2.0","id":"0","method":"get_block_template","params":{"wallet_address":"44GBHzv6ZyQdJkjqZje6KLZ3xSyN1hBSFAnLP6EAqJtCRVzMzZmeXTC2AHKDS9aEDTRKmo6a6o9r9j86pYfhCWDkKjbtcns","reserve_size":60}"#;
const GET_BLOCK_TEMPLATE_RESPONSE: &str = "...";
```

Only raw data is defined and not real types as those will be tested in `rpc/` crates, i.e. the full JSON-RPC 2.0 wrapped `GetBlockTemplateRequest` type will be tested in `cuprate-rpc-types`.

The only tests in `cuprate-test-utils` will be to make sure JSON data is valid JSON.

### Why
To act as test data for `rpc/` crates. Defined here so other crates can use it as well.

### TODO
The `crate::rpc::data::bin` module will be implemented later.


test-utils/src/rpc/mod.rs
```diff
@@ -1,25 +1,6 @@
...
-//! ```
-
-mod client;
-pub use client::HttpRpcClient;
+//! This module has a `monerod` RPC [`client`] and some real request/response [`data`].
 
-mod constants;
-pub use constants::LOCALHOST_RPC_URL;
+pub mod client;
+pub mod data;
```
hinto-janai, line 6: The old `rpc` module is now inside `rpc::client`.

test-utils/src/rpc/data/macros.rs
```diff
@@ -0,0 +1,168 @@
...
+//---------------------------------------------------------------------------------------------------- define_request_and_response
+/// A template for generating the RPC request and response `const` data.
+///
+/// See the [`crate::json`] module for example usage.
+///
+/// # Macro internals
+/// This macro uses:
+/// - [`define_request_and_response_doc`]
+/// - [`define_request_and_response_test`]
+macro_rules! define_request_and_response {
```
hinto-janai, line 12: Macro that generates all the `const` request/response data.

test-utils/src/rpc/data/bin.rs
```diff
@@ -0,0 +1,55 @@
+//! Binary data from [`.bin` endpoints](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_blocksbin).
+//!
+//! TODO: Not implemented yet.
```
hinto-janai, line 3: Will be implemented some other time since we need binary request inputs.

test-utils/src/rpc/data/json.rs
```diff
@@ -0,0 +1,1268 @@
...
+    "prev_hash": "9d648e741d85ca0e7acb4501f051b27e9b107d3cd7a3f03aa7f776089117c81a",
+    "reserved_offset": 131,
+    "seed_hash": "e2aa0b7b55042cd48b02e395d78fa66a29815ccc1584e38db2d1f0e8485cd44f",
+    "seed_height": 3194880,
+    "status": "OK",
+    "untrusted": false,
+    "wide_difficulty": "0x41f64bf3ff"
+  }
+}"#;
+}
```
hinto-janai, line 57: Macro example usage.

test-utils/src/rpc/data/json.rs
```diff
@@ -0,0 +1,1268 @@
...
+// r#"{
+//   "id": "0",
+//   "jsonrpc": "2.0",
+//   "result": {
+//     "backlog": "...Binary...",
+//     "status": "OK",
+//     "untrusted": false
+//   }
+// }"#;
+// }
```
hinto-janai, line 1089: TODO: will be implemented when we figure out binary strings.

test-utils/src/rpc/data/macros.rs
```diff
@@ -0,0 +1,168 @@
...
+    };
+}
+pub(super) use define_request_and_response_doc;
+
+//---------------------------------------------------------------------------------------------------- define_request_and_response_test
+/// Generate documentation for the types generated
+/// by the [`define_request_and_response`] macro.
+///
+/// See it for more info on inputs.
+macro_rules! define_request_and_response_doc_test {
```
hinto-janai, line 104: All `/json_rpc` and other JSON types have tests that assert they are valid JSON.

Tests that assert correct (de)serialization of values will be in the `rpc/` crates.

test-utils/src/rpc/data/json.rs
```diff
@@ -0,0 +1,1292 @@
...
+r#"{
+  "id": "0",
+  "jsonrpc": "2.0",
+  "result": {
+    "txids": "",
+    "status": "OK",
+    "untrusted": false
+  }
+}"#;
+}
```
hinto-janai, line 1286: TODO: this has to be moved in `rpc` as well, it should be in `json.rs` not `other.rs`.

test-utils/src/rpc/client/mod.rs
Boog900, line 1: Couldn't we keep `client.rs` where it is and have that act as the `mod` for `constants`?

Boog900: it would also allow us to remove `#![allow(clippy::module_inception)]`

Boog900 (approved):

PR #232, opened 2024-07-19
rpc: add JSON doc-tests

hinto-janai:
### What
Adds doc-tests to most JSON types (`{json,other}.rs`) in `cuprate-rpc-types`, with some left as TODOs.

Also fixes some data in `cuprate-test-utils` and types in `cuprate-rpc-types`.

### How
Both `{json,other}.rs` define a file-private macro `serde_doc_test!()` which takes in a:
- `const _: &str` from `cuprate-test-utils::rpc::data` added in https://github.com/Cuprate/cuprate/pull/231
- expected value after deserialization

The doc-test asserts that the expected value is deserialized from the `const _: &str`.

Some `const: &str` are huge so some tests only test that (de)serialization into the type works, not that values are correct, e.g. `GET_TRANSACTIONS_RESPONSE`.

This is set-up as an optional `#[doc]` + macro as to make it more modular. Some types are harder to test and baking this into the type generator macro makes it even more complicated.

### Example
This is placed above `GetBlockTemplate`'s request type:

```rust
#[doc = serde_doc_test!(
    GET_BLOCK_TEMPLATE_REQUEST => GetBlockTemplateRequest {
        extra_nonce: String::default(),
        prev_block: String::default(),
        reserve_size: 60,
        wallet_address: "44GBHzv6ZyQdJkjqZje6KLZ3xSyN1hBSFAnLP6EAqJtCRVzMzZmeXTC2AHKDS9aEDTRKmo6a6o9r9j86pYfhCWDkKjbtcns".into(),
    }
)]
```


rpc/types/src/json.rs
```diff
@@ -92,6 +169,23 @@ define_request_and_response! {
...
+            height: 3195018,
+            next_seed_hash: "".into(),
+            prev_hash: "9d648e741d85ca0e7acb4501f051b27e9b107d3cd7a3f03aa7f776089117c81a".into(),
+            reserved_offset: 131,
+            seed_hash: "e2aa0b7b55042cd48b02e395d78fa66a29815ccc1584e38db2d1f0e8485cd44f".into(),
+            seed_height: 3194880,
+            wide_difficulty: "0x41f64bf3ff".into(),
+        }
+    )]
     ResponseBase {
```
hinto-janai, line 189: Used on responses as well.

I had second thoughts about putting the tests directly on-top and in the same file as the types themselves since it's a bit noisy and makes the files quite long but:

1. There's only 1 doc-test per type
2. The test is as close as possible to the definition site; better for maintenance

I think I'd be okay with a separate file for tests as well, downside is that it separates types and tests and it won't be a doc-test, just a normal test.

Boog900: I think this is ok, although the type definitions are starting to get pretty confusing...

Boog900 (approved):

PR #233, opened 2024-07-21
rpc: impl `cuprate-rpc-interface`

hinto-janai:
### What
Implements the entirety of the `cuprate-rpc-interface` crate (with a few TODOs).

### How
See below review.


rpc/interface/src/route/fallback.rs
```diff
@@ -0,0 +1,18 @@
...
+//---------------------------------------------------------------------------------------------------- Import
+use axum::http::StatusCode;
+
+//---------------------------------------------------------------------------------------------------- Routes
+/// Fallback route function.
+///
+/// This is used as the fallback endpoint in [`crate::RouterBuilder`].
+pub(crate) async fn fallback() -> StatusCode {
+    StatusCode::NOT_FOUND
+}
```
hinto-janai, line 12: Fallback route function, i.e. the behavior to do on `/unknown_endpoints`. This currently follows `monerod`, it replies with a 404 with no body.

We can change this later.

rpc/interface/src/router_builder.rs
```diff
@@ -0,0 +1,198 @@
...
+    bin_getblocks                   => "/getblocks.bin"                   => bin::get_blocks                  => (get, post),
+    bin_get_blocks_by_height        => "/get_blocks_by_height.bin"        => bin::get_blocks_by_height        => (get, post),
+    bin_getblocks_by_height         => "/getblocks_by_height.bin"         => bin::get_blocks_by_height        => (get, post),
+    bin_get_hashes                  => "/get_hashes.bin"                  => bin::get_hashes                  => (get, post),
+    bin_gethashes                   => "/gethashes.bin"                   => bin::get_hashes                  => (get, post),
+    bin_get_o_indexes               => "/get_o_indexes.bin"               => bin::get_o_indexes               => (get, post),
+    bin_get_outs                    => "/get_outs.bin"                    => bin::get_outs                    => (get, post),
+    bin_get_transaction_pool_hashes => "/get_transaction_pool_hashes.bin" => bin::get_transaction_pool_hashes => (get, post),
+    bin_get_output_distribution     => "/get_output_distribution.bin"     => bin::get_output_distribution     => (get, post),
+}
```
hinto-janai, line 191: Route map.

builder function name => actual route literal => async route function => `$(supported HTTP methods)*`.

`monerod` seems to be pretty loose with what HTTP methods are supported, e.g. this works:
```bash
curl -v http://127.0.0.1:18081/get_limit --request DELETE -H 'Content-Type: application/json'
```
Some of these also don't make sense as a `get` since they require body content. Will fix alongside creating the handlers.

rpc/interface/src/rpc_handler.rs
```diff
@@ -0,0 +1,57 @@
...
+//---------------------------------------------------------------------------------------------------- RpcHandler
+/// An RPC handler.
+///
+/// This trait represents a type that can turn [`RpcRequest`]s into [`RpcResponse`]s.
+///
+/// Implementors of this trait must be [`tower::Service`]s that use:
+/// - [`RpcRequest`] as the generic `Request` type
+/// - [`RpcResponse`] as the associated `Response` type
+/// - [`RpcError`] as the associated `Error` type
+/// - `InfallibleOneshotReceiver<Result<RpcResponse, RpcError>>` as the associated `Future` type
```
Boog900, line 25: I don't think we should put this bound `InfallibleOneshotReceiver` on the implementer as it's quite restrictive and pretty much prevents the use of middleware.

Boog900 (changes requested):

rpc/interface/src/route/other.rs
```diff
@@ -0,0 +1,138 @@
...
+    stop_mining => StopMining,
+    mining_status => MiningStatus,
+    save_bc => SaveBc,
+    get_transaction_pool => GetTransactionPool,
+    get_transaction_pool_stats => GetTransactionPoolStats,
+    stop_daemon => StopDaemon,
+    get_limit => GetLimit,
+    get_net_stats => GetNetStats,
+    get_transaction_pool_hashes => GetTransactionPoolHashes
+}
```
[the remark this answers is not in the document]
Boog900: I can't think of one

Boog900:
great work btw!

rpc/interface/src/route/json_rpc.rs
```diff
@@ -0,0 +1,68 @@
...
+pub(crate) async fn json_rpc<H: RpcHandler>(
+    State(handler): State<H>,
+    Json(request): Json<cuprate_json_rpc::Request<JsonRpcRequest>>,
+) -> Result<Json<cuprate_json_rpc::Response<JsonRpcResponse>>, StatusCode> {
+    // TODO: <https://www.jsonrpc.org/specification#notification>
+    //
+    // JSON-RPC notifications (requests without `id`)
+    // must not be responded too, although, the request's side-effects
+    // must remain. How to do this considering this function will
+    // always return and cause `axum` to respond?
```
hinto-janai, line 31: This is (maybe) one of the differences that will be documented in the architecture book: https://www.jsonrpc.org/specification#notification.

On one hand, `monerod` ignores this and following that behavior is easier, on the other hand, the specification very clearly states this is incorrect:

> The Server MUST NOT reply to a Notification

My first thought is to test bitcoin/eth/zcash and see if they follow the spec, if so, I think we should as well.

Boog900: Yeah I think I agree, the spec does seem pretty firm there.

However it is probably not worth the effort if monerod is already breaking spec.

Boog900 (approved):

PR #241, opened 2024-08-01
rpc-types: add traits and `enum` requests/responses

hinto-janai:
### What
This adds some types/traits to `cuprate_rpc_types`.

This adds 6 enums for encapsulating RPC requests/responses:
- `crate::json::JsonRpcRequest` & `crate::json::JsonRpcResponse`
- `crate::bin::BinRequest` & `crate::bin::BinResponse`
- `crate::other::OtherRequest` & `crate::other::OtherResponse`

2 traits that contains some metadata about RPC request/responses:
- `RpcCall`
- `RpcCallValue`

and changes to the type generator macro to auto implement the traits.

### Why
Both enum types and metadata is needed for actual RPC usage: https://github.com/Cuprate/cuprate/pull/233.

### How
See below review.

### DIFF
--- a/rpc/types/src/json.rs
+++ b/rpc/types/src/json.rs
@@ -3,6 +3,9 @@
 //! All types are originally defined in [`rpc/core_rpc_server_commands_defs.h`](https://github.com/monero-project/monero/blob/cc73fe71162d564ffda8e549b79a350bca53c454/src/rpc/core_rpc_server_commands_defs.h).
 
 //---------------------------------------------------------------------------------------------------- Import
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
 use crate::{
     base::{AccessResponseBase, ResponseBase},
     defaults::{
@@ -16,6 +19,7 @@ use crate::{
         GetMinerDataTxBacklogEntry, HardforkEntry, HistogramEntry, OutputDistributionData, SetBan,
         Span, Status, SyncInfoPeer, TxBacklogEntry,
     },
+    rpc_call::RpcCallValue,
 };
 
 //---------------------------------------------------------------------------------------------------- Macro
@@ -93,7 +97,17 @@ define_request_and_response! {
     cc73fe71162d564ffda8e549b79a350bca53c454 => core_rpc_server_commands_defs.h => 943..=994,
 
     // The base type name.
-    GetBlockTemplate,
+    //
+    // After the type name, 2 optional idents are allowed:
+    // - `restricted`
+    // - `empty`
+    //
+    // These have to be within `()` and will affect the
+    // [`crate::RpcCall`] implementation on the request type.
+    //
+    // This type is not either restricted or empty so nothing is
+    // here, but the correct syntax is shown in a comment below:
+    GetBlockTemplate /* (restricted, empty) */,
 
     // The request type.
     //
@@ -218,7 +232,7 @@ define_request_and_response! {
     get_block_count,
     cc73fe71162d564ffda8e549b79a350bca53c454 =>
     core_rpc_server_commands_defs.h => 919..=933,
-    GetBlockCount,
+    GetBlockCount (empty),
 
     // There are no request fields specified,
     // this will cause the macro to generate a
@@ -300,7 +314,7 @@ define_request_and_response! {
     cc73fe71162d564ffda8e549b79a350bca53c454 =>
     core_rpc_server_commands_defs.h => 1130..=1161,
 
-    GenerateBlocks,
+    GenerateBlocks (restricted),
 
     #[doc = serde_doc_test!(
         GENERATE_BLOCKS_REQUEST => GenerateBlocksRequest {
@@ -633,7 +647,7 @@ define_request_and_response! {
     cc73fe71162d564ffda8e549b79a350bca53c454 =>
     core_rpc_server_commands_defs.h => 1734..=1754,
 
-    GetConnections,
+    GetConnections (restricted, empty),
 
     Request {},
 
@@ -708,7 +722,7 @@ define_request_and_response! {
     get_info,
     cc73fe71162d564ffda8e549b79a350bca53c454 =>
     core_rpc_server_commands_defs.h => 693..=789,
-    GetInfo,
+    GetInfo (empty),
     Request {},
 
     #[doc = serde_doc_test!(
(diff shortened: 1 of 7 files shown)


rpc/types/src/rpc_call.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Is `true` if this RPC method has no inputs, i.e. it is a `struct` with no fields.
+    ///
+    /// ```rust
+    /// use cuprate_rpc_types::{RpcCall, json::*};
+    ///
+    /// assert!(GetBlockCountRequest::IS_EMPTY);
+    /// assert!(!OnGetBlockHashRequest::IS_EMPTY);
+    /// ```
+    const IS_EMPTY: bool;
+}
```
hinto-janai, line 45: This trait is used within `cuprate_rpc_interface` in the following manner:
```rust
async fn route(...) {
    if RequestType::IS_RESTRICTED && rpc_server_is_restricted {
        /* ... */
    } else {
        /* ... */
    }
}
```

rpc/types/src/rpc_call.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Same as [`RpcCall::IS_EMPTY`].
+    ///
+    /// ```rust
+    /// use cuprate_rpc_types::{RpcCallValue, json::*};
+    ///
+    /// assert!(GetBlockCountRequest::default().is_empty());
+    /// assert!(!OnGetBlockHashRequest::default().is_empty());
+    /// ```
+    fn is_empty(&self) -> bool;
+}
```
hinto-janai, line 78: This trait is a runtime mirror of `RpcCall`. It exists because sometimes we can only know the RPC call at runtime, for example, we must deserialize methods in `/json_rpc` before we know what it is.

Once we have some method variable, we do:
```rust
async fn route(...) {
    if request.is_restricted() && rpc_server_is_restricted {
        /* ... */
    } else {
        /* ... */
    }
}
```
`RpcCall` is used when we _know_ the type at runtime, e.g. `/get_height` must lead to a `GetHeightRequest` type, so we can use `GetHeightRequest::IS_RESTRICTED` instead.

rpc/types/src/other.rs
```diff
@@ -930,6 +952,178 @@ define_request_and_response! {
...
+    SetLimit(SetLimitRequest),
+    OutPeers(OutPeersRequest),
+    InPeers(InPeersRequest),
+    GetNetStats(GetNetStatsRequest),
+    GetOuts(GetOutsRequest),
+    Update(UpdateRequest),
+    PopBlocks(PopBlocksRequest),
+    GetTransactionPoolHashes(GetTransactionPoolHashesRequest),
+    GetPublicNodes(GetPublicNodesRequest),
+}
```
hinto-janai, line 1006: Example of a request enum.

These request/response enums are used within `cuprate_rpc_interface`'s `tower::Service` API, not directly, but something like:
```rust
// inside `cuprate_rpc_interface`
enum RpcRequest {
    JsonRpc(cuprate_json_rpc::Request<JsonRpcRequest>),
    Other(OtherRequest),
    Binary(BinRequest),
}
```
then this broad `RpcRequest` type is used like such: `tower::Service<RpcRequest>`.

rpc/types/src/json.rs
```diff
@@ -93,7 +97,17 @@ define_request_and_response! {
...
+    // After the type name, 2 optional idents are allowed:
+    // - `restricted`
+    // - `empty`
+    //
+    // These have to be within `()` and will affect the
+    // [`crate::RpcCall`] implementation on the request type.
+    //
+    // This type is not either restricted or empty so nothing is
+    // here, but the correct syntax is shown in a comment below:
+    GetBlockTemplate /* (restricted, empty) */,
```
hinto-janai, line 110: The way types can easily implement `RpcCall` is by using this syntax when defining a type.

`GetBlockTemplate` is neither empty or restricted so it has nothing. The type generator macro will detect this and implement:
```rust
impl crate::RpcCall for GetBlockTemplate {
    const IS_RESTRICTED: bool = false;
    const IS_EMPTY: bool = false;
}
```

rpc/types/src/macros.rs
```diff
@@ -152,22 +222,7 @@ macro_rules! define_request {
...
-
-    //------------------------------------------------------------------------------
-    // This branch of the macro expects fields within the `{}`,
-    // and will generate a `struct`
-    (
-        // Any doc comments, derives, etc.
-        $( #[$attr:meta] )*
-        // The response type.
-        $t:ident {
+        $t:ident $(($restricted:ident $(, $empty:ident)?))? {
```
hinto-janai, line 225: Empty requests being aliased to `()` unfortunately has to go since each request must have a concrete type so we can implement `RpcCall` on it.

rpc/types/src/json.rs
```diff
@@ -1534,6 +1552,203 @@ define_request_and_response! {
...
+///
+/// TODO: document and test (de)serialization behavior after figuring out `method/params`.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[cfg_attr(
+    feature = "serde",
+    serde(rename_all = "snake_case", tag = "method", content = "params")
+)]
+#[allow(missing_docs)]
+pub enum JsonRpcRequest {
```
hinto-janai, line 1586: TODO: this is the problematic `method/params` enum btw. Will fix in another PR.

rpc/types/src/macros.rs
hinto-janai: Although it's worth noting `()` aliases didn't matter for `cuprate_rpc_interface` usage.

rpc/types/src/json.rs
```diff
@@ -1534,6 +1552,203 @@ define_request_and_response! {
     }
 }
 
+define_request_and_response! {
+    UNDOCUMENTED_METHOD,
+    cc73fe71162d564ffda8e549b79a350bca53c454 =>
+    core_rpc_server_commands_defs.h => 2798..=2823,
+
+    GetTxIdsLoose,
+
```
Boog900, line 1561: This is going to be an annoying endpoint to support, it may require DB changes, depending on what guarantees we need to make

Boog900: I would probably be ok dropping support for this end point.

Boog900 (approved):

PR #249, opened 2024-08-06
ci: add book build tests

hinto-janai:
### What
Adds CI that builds `/books` with `mdbook` if they were modified.

### DIFF
--- a/.github/workflows/monero-book.yml
+++ b/.github/workflows/monero-book.yml
@@ -1,0 +1,30 @@
+# This action attempts to build the Monero book, if changed.
+
+name: Monero mdBook
+
+on:
+  push:
+    paths:
+      - 'books/protocol/**'
+
+env:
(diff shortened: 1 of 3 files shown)


.github/workflows/monero-book.yml
```diff
@@ -0,0 +1,30 @@
...
+name: Monero mdBook
+
+on:
+  push:
+    paths:
+      - 'books/protocol/**'
+
+env:
+  # Version of `mdbook` to install.
+  MDBOOK_VERSION: 0.4.36
```
Boog900, line 12: `monero-book` uses `svgbob` which needs to be installed as well: https://github.com/Cuprate/monero-book/blob/main/.github/workflows/mdbook.yml

Boog900 (changes requested):

Boog900 (approved):

PR #255, opened 2024-08-09
rpc: remove temporary lints

hinto-janai:
### What
- Removes temporary lints from `rpc/`
- Fixes dead code, unused imports, etc


rpc/interface/src/lib.rs
```diff
@@ -76,21 +76,12 @@
 	// TODO
 	rustdoc::bare_urls,
 
+    clippy::multiple_crate_versions,
```
Boog900, line 79: the indenting is different here, I think the other lints are incorrect though not this line.

hinto-janai: I guess `rustfmt` doesn't work inside `#![]`, fixed in other crates too.

Boog900 (approved):

Issue #236, opened 2024-07-25
Usage of `LazyLock` over `OnceLock` + `fn`


hinto-janai:
## What
[Rust 1.80 stablized `LazyLock`](https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#lazycell-and-lazylock).

This can replace any `std::sync::OnceLock` + `fn` usage.

## Where
Any place using `OnceLock` + `fn` for `LazyLock` behavior, e.g.:

https://github.com/Cuprate/cuprate/blob/929d19c4508a84d886ece03009a6fcdc5edea5c2/helper/src/fs.rs#L86-L90

## How
Replace `OnceLock` + `fn` with:
```rust
static _: LazyLock<T> = LazyLock::new(|| {
    /* old OnceLock::get_or_init function */
});
```

PR #256, opened 2024-08-09
Replace `OnceLock` + `fn` with `LazyLock`

hinto-janai:
### What
Closes https://github.com/Cuprate/cuprate/issues/236.

This replaces the `OnceLock` + `fn`s that were used in a `LazyLock` type of way for the entire codebase. There's other `OnceLock`s remaining but they are used differently.


consensus/rules/src/decomposed_amount.rs
```diff
@@ -1,13 +1,9 @@
...
-///
-static DECOMPOSED_AMOUNTS: OnceLock<[u64; 172]> = OnceLock::new();
+use std::sync::LazyLock;
 
 #[rustfmt::skip]
-pub fn decomposed_amounts() -> &'static [u64; 172] {
-    DECOMPOSED_AMOUNTS.get_or_init(|| {
-        [
+/// Decomposed amount table.
+pub static DECOMPOSED_AMOUNTS: LazyLock<[u64; 172]> = LazyLock::new(|| {
```
Boog900, line 5: Can we remove the `LazyLock` here, keep it a static though. I don't know why I made this a `OnceLock`, the data is still in the binary.

Boog900 (changes requested):

consensus/rules/src/decomposed_amount.rs
SyntheticBird45: I must say I'm confused on why this isn't a `const`

hinto-janai: `static` is a ref to 1 memory address containing the data, `const` inlines the data everywhere it gets used.

I assume @Boog900 doesn't want it inlined in the binary everywhere it gets used.

SyntheticBird45: I was confused, I thought const weren't inlined. Thanks for the heads-up

Boog900 (approved):

PR #257, opened 2024-08-13
dandelion-tower: improve API

Boog900:
### What

This PR makes the dandelion crate's tx-pool abstraction more efficient.

### How
by separating new tx handling from the pool maintenance.


p2p/dandelion-tower/src/pool/incoming_tx.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// The transaction ID.
+    pub(crate) tx_id: TxID,
+    /// The routing state of the transaction.
+    pub(crate) routing_state: TxState<PID>,
+}
+
+/// An [`IncomingTx`] builder.
+///
+// The const generics here are used to restrict what methods can be called.
+pub struct IncomingTxBuilder<const RS: bool, const DBS: bool, Tx, TxID, PID> {
```
Boog900, line 17: I created an incoming tx builder to force people to use the dandelion pool correctly.

I also tried a new technique to force people to use the builder correctly.

p2p/dandelion-tower/src/pool/manager.rs
```diff
@@ -0,0 +1,294 @@
...
+
+#[derive(Copy, Clone, Debug, thiserror::Error)]
+#[error("The dandelion pool was shutdown")]
+pub struct DandelionPoolShutDown;
+
+/// The dandelion++ pool manager.
+///
+/// See the [module docs](super) for more.
+pub struct DandelionPoolManager<P, R, Tx, TxID, PID> {
+    /// The dandelion++ router
```
Boog900, line 32: I also renamed this to a more fitting name

Boog900 (commented):
@hinto-janai I decided to go another way than what I first set out to do. Instead of making the dandelion pool handle more of the tx-pool, I have changed it so it now handles less.

Making it handle more while keeping it generic got really messy.

p2p/dandelion-tower/src/pool/incoming_tx.rs
```diff
@@ -0,0 +1,110 @@
+//! Contains [`IncomingTx`] and [`IncomingTxBuilder`]
+use crate::{State, TxState};
+
+/// An incoming transaction that has gone through the preprocessing stage.
+pub struct IncomingTx<Tx, TxID, PID> {
```
hinto-janai, line 5: `TxId`, `Pid`?

Instead of enforcing this myself, can we add this to `deny`?: https://rust-lang.github.io/rust-clippy/master/index.html#/upper_case_acronyms

p2p/dandelion-tower/src/pool/incoming_tx.rs
hinto-janai, line 17: Can you document what `RS` and `DBS` are or change them to more obvious names like `ROUTING_STATE` and `DATABASE_STATE` or `DB_STATE`? I'm not sure what `PID` is either, pool ID?

If these are used elsewhere maybe add a section in the `lib.rs` docs instead.

hinto-janai (changes requested):

p2p/dandelion-tower/src/pool/incoming_tx.rs
hinto-janai: Generic would fall under `Type parameters` so:

> concise `UpperCamelCase`, usually single uppercase letter: `T`

I'm guilty too...

https://github.com/Cuprate/cuprate/blob/59adf6dcf8a1adfbfbf05583866f6c5b0d85b374/storage/database/src/tables.rs#L275-L280

Boog900: Yeah sure we could test it

p2p/dandelion-tower/src/pool/incoming_tx.rs
Boog900: I added docs above, these are just an implementation detail so probably don't need to be in public docs

p2p/dandelion-tower/src/pool/incoming_tx.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// The routing state of the transaction.
+    pub(crate) routing_state: TxState<PId>,
+}
+
+/// An [`IncomingTx`] builder.
+///
+// The const generics here are used to restrict what methods can be called.
+// RS: a bool for if the routing state is set.
+// DBS: a bool for if the state in DB is set.
+pub struct IncomingTxBuilder<const RS: bool, const DBS: bool, Tx, TxId, PId> {
```
hinto-janai, line 19: > these are just an implementation detail so probably don't need to be in public docs

Isn't it part of the public API? It's not a big deal here but I don't think it hurts to add a couple doc lines to make it immediately clear what these are.

```suggestion
/// An [`IncomingTx`] builder.
///
/// The const generics here are used to restrict what methods can be called.
///
/// - `RS`: routing state; a `bool` for if the routing state is set
/// - `DBS`: database state; a `bool` for if the state in the DB is set
pub struct IncomingTxBuilder<const RS: bool, const DBS: bool, Tx, TxId, PId> {
```

p2p/dandelion-tower/src/pool/incoming_tx.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// The transaction ID.
+    tx_id: TxId,
+    /// The routing state of the transaction.
+    routing_state: Option<TxState<PId>>,
+    /// The state of this transaction in the DB.
+    state_in_db: Option<State>,
+}
+
+impl<Tx, TxId, PId> IncomingTxBuilder<false, false, Tx, TxId, PId> {
+    /// Crates a new [`IncomingTxBuilder`].
```
hinto-janai, line 31:
```suggestion
    /// Creates a new [`IncomingTxBuilder`].
```

p2p/dandelion-tower/src/pool/incoming_tx.rs
```diff
@@ -0,0 +1,112 @@
+//! Contains [`IncomingTx`] and [`IncomingTxBuilder`]
+use crate::{State, TxState};
+
+/// An incoming transaction that has gone through the preprocessing stage.
+pub struct IncomingTx<Tx, TxId, PId> {
```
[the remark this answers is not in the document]
Boog900: > peer is never abbreviated as P...

generics are often very short so this should not matter.

> This would trigger the lint

True but that's more an issue with the lint, I'll change it anyway

hinto-janai (approved):
> It's only 2 letters being saved

*3 letters. BTW if you don't think it matters you can just say no, I push for this because I don't think names like this are immediately obvious if you aren't the original author.

PR #258, opened 2024-08-16
Binaries: add cuprated skeleton

Boog900:
### What

Adds a rough file structure for `cuprated`.


binaries/cuprated/src/blockchain.rs
```diff
@@ -0,0 +1,6 @@
+//! Blockchain
```
hinto-janai, line 1: Do you prefer `{a.rs, a/}` over `a/mod.rs` for any specific reasons?

I use `a/mod.rs` because it encapsulates exporting and means all module files are in 1 directory. I rather have the entire repo stick to 1 standard though.

hinto-janai (approved):

binaries/cuprated/src/blockchain.rs
Boog900: Yeah I prefer not having lots of `mod.rs`, although I don't mind that much.

It's also "encouraged" here: https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

I don't mind having a mix in the repo

Issue #194, opened 2024-06-26
Alt Block/ Chain Handling


Boog900:
## What

Adding alt block handling to Cuprate's components, the consensus and DB code needs to be able to handle reorgs.

The block downloader can already handle downloading an alt chain.

## How

TODO

PR #260, opened 2024-08-29
Blockchain: add alt-block handling 

Boog900:
### What

Adds alt block handling to the blockchain storage.

### How
Follows #194


helper/src/tx_utils.rs
```diff
@@ -0,0 +1,34 @@
+//! Utils for working with [`Transaction`]
+
+use monero_serai::transaction::{Input, Transaction};
+
+/// Calculates the fee of the [`Transaction`].
+///
+/// # Panics
+/// This will panic if the inputs overflow or the transaction outputs too much, so should only
+/// be used on known to be valid txs.
+pub fn tx_fee(tx: &Transaction) -> u64 {
```
Boog900, line 10: I moved this function from `test-utils` as it's also now needed in the DB

storage/blockchain/src/ops/alt_block/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+//! [`get_alt_chain_history_ranges`] covers this and is the method to get the ranges of heights needed from each [`ChainID`](cuprate_types::ChainId)
+//! to get all the alt-blocks in a given [`ChainID`](cuprate_types::ChainId).
+//!
+//! Although this should be kept in mind as a possibility, because Cuprate's block downloader will only track a single chain it is
+//! unlikely that we will be tracking [`ChainID`](cuprate_types::ChainId) that don't immediately connect to the main-chain.
+//!
+//! ## Why not use block's previous field?
+//!
+//! Although that would be easier, it makes getting a range of block extremely slow, as we have to build the weight cache to verify
+//! blocks, roughly 100,000 block headers needed, this cost was seen as too high.
```
Boog900, line 48: Here is some information on how alt-blocks are stored

storage/blockchain/src/ops/block.rs
```diff
@@ -33,11 +40,6 @@ use crate::{
 /// This function will panic if:
 /// - `block.height > u32::MAX` (not normally possible)
 /// - `block.height` is not != [`chain_height`]
-///
-/// # Already exists
-/// This function will operate normally even if `block` already
-/// exists, i.e., this function will not return `Err` even if you
-/// call this function infinitely with the same block.
```
Boog900, line 40: I also removed this as we panic if the same block gets passed in again

storage/blockchain/src/ops/block.rs
```diff
@@ -129,23 +130,24 @@ pub fn add_block(
...
 /// The removed block's data is returned.
+///
+/// If a [`ChainId`] is specified the popped block will be added to the alt block tables under
+/// that [`ChainId`]. Otherwise, the block will be completely removed from the DB.
 #[doc = doc_error!()]
 ///
 /// In `pop_block()`'s case, [`RuntimeError::KeyNotFound`]
 /// will be returned if there are no blocks left.
 // no inline, too big
 pub fn pop_block(
```
Boog900, line 141: I also changed `pop_block` to add the block to the alt-blocks if a `ChainId` is specified

storage/blockchain/src/service/write.rs
```diff
@@ -55,7 +64,167 @@ fn write_block(env: &ConcreteEnv, block: &VerifiedBlockInformation) -> ResponseR
...
+        }
+    }
+}
+
+/// [`BlockchainWriteRequest::ReverseReorg`].
+fn reverse_reorg(env: &ConcreteEnv, chain_id: ChainId) -> ResponseResult {
+    let env_inner = env.env_inner();
+    let tx_rw = env_inner.tx_rw()?;
+
+    // TODO: turn this function into a try block once stable.
```
hinto-janai, line 159:
```suggestion
    // FIXME: turn this function into a try block once stable.
```

storage/blockchain/src/service/write.rs
```diff
@@ -55,7 +64,167 @@ fn write_block(env: &ConcreteEnv, block: &VerifiedBlockInformation) -> ResponseR
...
+        let chain_info = tables_mut.alt_chain_infos().get(&chain_id.into())?;
+        // Although this doesn't guarantee the chain was popped from the main-chain, it's an easy
+        // thing for us to check.
+        assert_eq!(Chain::from(chain_info.parent_chain), Chain::Main);
+
+        let tob_block_height =
+            crate::ops::blockchain::top_block_height(tables_mut.block_heights())?;
+
+        // pop any blocks that were added as part of a re-org.
+        for _ in chain_info.common_ancestor_height..tob_block_height {
```
hinto-janai, line 172:
```suggestion
        let top_block_height =
            crate::ops::blockchain::top_block_height(tables_mut.block_heights())?;

        // pop any blocks that were added as part of a re-org.
        for _ in chain_info.common_ancestor_height..top_block_height {
```

hinto-janai: Also, shouldn't this be `..=`?

storage/blockchain/src/tables.rs
```diff
@@ -129,6 +130,40 @@ cuprate_database::define_tables! {
...
+    18 => AltBlockBlobs,
+    AltBlockHeight => BlockBlob,
+
+    /// Alt-Block transactions blobs.
+    ///
+    /// Contains the raw bytes of alt transactions, if those transactions are not in the main-chain.
+    19 => AltTransactionBlobs,
+    TxHash => TxBlob,
+
+    /// Alt-Block transactions information.
```
hinto-janai, line 162:
```suggestion
    /// Alt-block blobs.
    ///
    /// Contains the raw bytes of all alt-blocks.
    18 => AltBlockBlobs,
    AltBlockHeight => BlockBlob,

    /// Alt-block transaction blobs.
    ///
    /// Contains the raw bytes of alt transactions, if those transactions are not in the main-chain.
    19 => AltTransactionBlobs,
    TxHash => TxBlob,

    /// Alt-block transaction information.
```

storage/blockchain/src/types.rs
```diff
@@ -324,6 +326,259 @@ pub struct RctOutput {
...
+    fn from(value: ChainId) -> Self {
+        Self(value.0.get())
+    }
+}
+
+impl From<RawChainId> for ChainId {
+    fn from(value: RawChainId) -> Self {
+        Self(NonZero::new(value.0).expect("RawChainId mut not have a value of `0`"))
+    }
+}
```
hinto-janai, line 419:
```suggestion
impl From<RawChainId> for ChainId {
    fn from(value: RawChainId) -> Self {
        Self(NonZero::new(value.0).expect("RawChainId cannot have a value of `0`"))
    }
}
```

hinto-janai (commented):
Will review again.

storage/blockchain/src/service/write.rs
Boog900: `common_ancestor_height` is one less than the range start so by using the `top_block_height` as the end it makes the count correct, a bit confusing but should be ok.

storage/blockchain/src/ops/alt_block/tx.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let tx_blob = match tables.alt_transaction_blobs().get(tx_hash) {
+        Ok(blob) => blob.0,
+        Err(RuntimeError::KeyNotFound) => {
+            let tx_id = tables.tx_ids().get(tx_hash)?;
+
+            let blob = tables.tx_blobs().get(&tx_id)?;
+
+            blob.0
+        }
```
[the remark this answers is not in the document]
Boog900: An alt block is a block that was mined at the same height as another block, if we add a block to the chain and then another (alt)block gets broadcasted at the same height then they could share txs

storage/blockchain/src/ops/alt_block/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+pub use tx::*;
+
+mod block;
+mod chain;
+mod tx;
+
+/// Flush all alt-block data from all the alt-block tables.
+///
+/// This function completely empties the alt block tables.
+pub fn flush_alt_blocks<'a, E: cuprate_database::EnvInner<'a>>(
```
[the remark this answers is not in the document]
Boog900: It's such a good place for this function though :) where would you prefer it?

types/src/types.rs
```diff
@@ -97,7 +99,7 @@ pub struct VerifiedBlockInformation {
 ///
 /// The inner value is meaningless.
 #[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
-pub struct ChainId(pub u64);
+pub struct ChainId(pub NonZero<u64>);
```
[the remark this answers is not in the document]
Boog900: It could but it would change what this type represents and IMO would be less Rusty

types/src/blockchain.rs
```diff
@@ -197,12 +218,24 @@ pub enum BlockchainResponse {
...
+
+    //------------------------------------------------------ Writes
+    /// A generic Ok response to indicate a request was successfully handled.
+    ///
+    /// currently the response for:
+    /// - [`BlockchainWriteRequest::WriteBlock`]
+    /// - [`BlockchainWriteRequest::WriteAltBlock`]
+    /// - [`BlockchainWriteRequest::ReverseReorg`]
+    /// - [`BlockchainWriteRequest::FlushAltBlocks`]
+    Ok,
```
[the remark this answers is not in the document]
Boog900: I added a test that tests all the new requests

storage/blockchain/src/service/read.rs
```diff
@@ -254,7 +291,38 @@ fn block_extended_header_in_range(
...
+                                &AltBlockHeight {
+                                    chain_id: chain_id.into(),
+                                    height,
+                                },
+                                tables,
+                            ),
+                        }
+                    })
+                })
+                .flatten()
```
hinto-janai, line 323: nit: `.map().flatten()` -> `.flat_map()`.

[This lint](https://rust-lang.github.io/rust-clippy/master/index.html#/map_flatten) would trigger here but it isn't because it's `rayon`'s iter functions (I think).

storage/blockchain/src/service/write.rs
```diff
@@ -55,7 +64,170 @@ fn write_block(env: &ConcreteEnv, block: &VerifiedBlockInformation) -> ResponseR
...
         Ok(()) => {
             TxRw::commit(tx_rw)?;
-            Ok(BlockchainResponse::WriteBlockOk)
+            Ok(BlockchainResponse::Ok)
+        }
+        Err(e) => {
+            // INVARIANT: ensure database atomicity by aborting
+            // the transaction on `add_block()` failures.
+            TxRw::abort(tx_rw)
+                .expect("could not maintain database atomicity by aborting write transaction");
```
hinto-janai, line 73: We should probably do this now that there's >1 write function.

`const` can be defined somewhere in this file, probably right after `use` statements.

```rust
/// Write functions within this module abort if the write transaction
/// could not be aborted successfully to maintain atomicity.
///
/// This is the panic message if the `abort()` fails.
const TX_RW_ABORT_FAIL: &str =
    "Could not maintain blockchain database atomicity by aborting write transaction";
```
```suggestion
            TxRw::abort(tx_rw).expect(TX_RW_ABORT_FAIL);
```

storage/blockchain/src/service/write.rs
```diff
@@ -55,7 +64,170 @@ fn write_block(env: &ConcreteEnv, block: &VerifiedBlockInformation) -> ResponseR
...
+
+        // Update the alt_chain_info with the correct information.
+        tables_mut.alt_chain_infos_mut().put(
+            &old_main_chain_id.into(),
+            &AltChainInfo {
+                parent_chain: Chain::Main.into(),
+                common_ancestor_height: last_block_height - 1,
+                chain_height: last_block_height + numb_blocks,
+            },
+        )?;
```
hinto-janai, line 134: It seems wrong to do direct DB mutation within `service`. Shouldn't this `put()` be in `crate::ops::block::pop_block` anyway?

storage/blockchain/src/service/write.rs
```diff
@@ -55,7 +64,170 @@ fn write_block(env: &ConcreteEnv, block: &VerifiedBlockInformation) -> ResponseR
...
+            .collect::<Vec<_>>();
+
+        // Add the old main chain blocks back to the main chain.
+        for res_alt_block in alt_blocks {
+            let alt_block = res_alt_block?;
+
+            let verified_block = map_valid_alt_block_to_verified_block(alt_block);
+
+            crate::ops::block::add_block(&verified_block, &mut tables_mut)?;
+        }
```
hinto-janai, line 196: Is this comment true?
```suggestion
        // Add the old main chain blocks back to the main chain.
        for height in (chain_info.common_ancestor_height + 1)..chain_info.chain_height {
            let alt_block = crate::ops::alt_block::get_alt_block(
                &AltBlockHeight {
                    chain_id: chain_id.into(),
                    height,
                },
                &tables_mut,
            )?;

            let verified_block = map_valid_alt_block_to_verified_block(alt_block);

            crate::ops::block::add_block(&verified_block, &mut tables_mut)?;
        }
```

storage/blockchain/src/types.rs
```diff
@@ -187,7 +189,7 @@ pub struct BlockInfo {
     /// The adjusted block size, in bytes.
     ///
     /// See [`block_weight`](https://monero-book.cuprate.org/consensus_rules/blocks/weights.html#blocks-weight).
-    pub weight: u64,
+    pub weight: usize,
```
hinto-janai, line 192: Is there a reason things are mostly `usize` instead of `u64`? Here but also across the codebase in general, e.g. DB service messages.

They are required to be `u64` elsewhere (especially in RPC) and these `usize`s permeate and cause `usize_to_u64()` to be everywhere. I understand indices but these aren't that. Can we start using `u64` instead?

test-utils/Cargo.toml
```diff
@@ -7,7 +7,7 @@ authors = ["Boog900", "hinto-janai"]
 
 [dependencies]
 cuprate-types             = { path = "../types" }
-cuprate-helper            = { path = "../helper", features = ["map"] }
+cuprate-helper            = { path = "../helper", features = ["map", "tx-utils"] }
```
hinto-janai, line 10:
```suggestion
cuprate-helper            = { path = "../helper", features = ["map", "tx"] }
```

hinto-janai (changes requested):

storage/blockchain/src/ops/alt_block/mod.rs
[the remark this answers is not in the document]
hinto-janai: `alt_block/block.rs` to be specific

storage/blockchain/src/service/write.rs
Boog900: I added some logic to `update_alt_chain_info` to handle adding blocks in reverse

storage/blockchain/src/service/write.rs
Boog900: oops, I retrieved the alt_block in an iterator, this is a better solution though

storage/blockchain/src/types.rs
Boog900: It's a trade of between what type is needed more, I used to use `u64`s everywhere but I changed it here: #201 and I do remember there being an improvement.

This is a weight which is often calculated by getting the size of a Vec, which is given in a `usize`. Plus you could change the JSON types to be `usize`s right?

storage/blockchain/src/service/free.rs
```diff
@@ -81,6 +84,44 @@ pub(super) const fn compact_history_genesis_not_included<const INITIAL_BLOCKS: u
...
+//---------------------------------------------------------------------------------------------------- Map Block
+/// Maps [`AltBlockInformation`] to [`VerifiedBlockInformation`]
+///
+/// # Panics
+/// This will panic if the block is invalid, so should only be used on blocks that have been popped from
+/// the main-chain.
+pub(super) fn map_valid_alt_block_to_verified_block(
+    alt_block: AltBlockInformation,
+) -> VerifiedBlockInformation {
+    let total_fees = alt_block.txs.iter().map(|tx| tx.fee).sum::<u64>();
```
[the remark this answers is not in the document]
Boog900: IMHO I don't think this would be helpful. When I map between types I am often mapping for the whole type not for a particular function. Also these types are specialized, for example you can't get `tx_fees` from a `monero_serai::Block`.

I would like some more unified "extra data" tx/block types, but I don't think that would solve the issue you mention.

storage/blockchain/src/types.rs
Boog900: Changing this actually removed multiple conversions in the DB

hinto-janai: The types are defined as 64-bit in `monerod`, I don't think defining public fields within public RPC types as `usize` is a good tradeoff. Probably won't affect many machines but not worth it IMO. I'm okay with using the cast functions if these need to be `usize` elsewhere.

storage/blockchain/src/service/write.rs
```diff
@@ -69,8 +76,7 @@ fn write_block(env: &ConcreteEnv, block: &VerifiedBlockInformation) -> ResponseR
         Err(e) => {
             // INVARIANT: ensure database atomicity by aborting
             // the transaction on `add_block()` failures.
-            TxRw::abort(tx_rw)
-                .expect("could not maintain database atomicity by aborting write transaction");
+            TxRw::abort(tx_rw).expect(TX_RW_ABORT_FAIL);
```
hinto-janai, line 79: These invariant comments (all of them) should be removed since the `const` documents the panic reasoning now.

hinto-janai (approved):

PR #261, opened 2024-09-01
workspace: add naming convention lints

hinto-janai:
### What
Adds:

- https://rust-lang.github.io/rust-clippy/master/index.html#/upper_case_acronyms
- https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#non-camel-case-types

to enforce the [Rust naming convention](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

---

Changes `RandomXVM` -> `RandomXVm` and some other things. To be fair `RandomX` is an uncommon case but here are some other VM examples:

- https://reth.rs/docs/reth/revm/struct.Evm.html
- https://reth.rs/docs/reth/core/rpc/types/trace/parity/struct.VmTrace.html

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+upper-case-acronyms-aggressive = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -250,6 +250,7 @@ empty_structs_with_brackets = "deny"
 empty_enum_variants_with_brackets = "deny"
 empty_drop = "deny"
 clone_on_ref_ptr = "deny"
+upper_case_acronyms  = "deny"
 
 # Hot
 # inline_always = "deny"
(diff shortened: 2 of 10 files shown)


clippy.toml
```diff
@@ -0,0 +1 @@
+upper-case-acronyms-aggressive = true
```
hinto-janai, line 1: This is a clippy option for `upper_case_acronyms`:

> Triggers if there is more than one uppercase char next to each other

So it catches things like `CamelCASE`.

Boog900: How does `RandomXVm` pass this lint, the `XV` are uppercase characters next to each other.

Boog900 (approved):
Approved just the one question.

clippy.toml
hinto-janai: [There's special conditions for acronyms](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/upper_case_acronyms.rs#L59-L72), either:

- 3 uppercase chars back-to-back
- 2 ending uppercase chars

In `RandomXVm`'s case, the last `m` resets the uppercase char count to 0...

PR #262, opened 2024-09-01
cuprated: initial RPC module skeleton

hinto-janai:
### What
Adds the initial skeleton for the RPC module in `cuprated` and many function signatures filled with `todo!()`.


binaries/cuprated/Cargo.toml
```diff
@@ -8,6 +8,68 @@ authors = ["Boog900", "hinto-janai", "SyntheticBird45"]
 repository = "https://github.com/Cuprate/cuprate/tree/main/binaries/cuprated"
 
 [dependencies]
+# TODO: after v1.0.0, remove unneeded dependencies.
```
hinto-janai, line 11: Added all the workspace dependencies to `cuprated` for now for convenience. We can remove them after v1.0.0.

binaries/cuprated/src/main.rs
```diff
@@ -1,3 +1,16 @@
...
+#![allow(
+    unused_imports,
+    unreachable_pub,
+    unused_crate_dependencies,
+    dead_code,
+    unused_variables,
+    clippy::needless_pass_by_value,
+    clippy::unused_async,
+    reason = "TODO: remove after v1.0.0"
+)]
```
hinto-janai, line 12: Same here, remove after impl is complete.

binaries/cuprated/src/rpc/handler.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// TODO
+#[derive(Clone)]
+pub struct CupratedRpcHandlerState {
+    /// Read handle to the blockchain database.
+    pub blockchain: BlockchainReadHandle,
+
+    /// Read handle to the transaction pool database.
+    pub txpool: TxpoolReadHandle,
+}
```
hinto-janai, line 44: This is `cuprated`'s `RpcHandler`.

I separated the state into another struct so the handler only needs to clone that instead of all of `CupratedRpcHandler`. It only save a single `bool` for now though.

binaries/cuprated/src/rpc/json.rs
```diff
@@ -0,0 +1,293 @@
...
+};
+use tower::ServiceExt;
+
+use crate::rpc::CupratedRpcHandlerState;
+
+/// Map a [`JsonRpcRequest`] to the function that will lead to a [`JsonRpcResponse`].
+pub(super) async fn map_request(
+    state: CupratedRpcHandlerState,
+    request: JsonRpcRequest,
+) -> Result<JsonRpcResponse, RpcError> {
```
hinto-janai, line 31: There is a `{json,bin,other}.rs` inside `rpc/`. One file filled with the handlers for each request type.

The pattern in each file is:
- `map_request` function that maps request enum -> handler function -> response enum
- all the handler functions

rpc/interface/src/rpc_error.rs
```diff
@@ -16,9 +16,16 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, thiserror::Error)]
+pub enum RpcError {
+    /// A [`std::io::Error`] from the database.
+    #[error("database I/O error: {0}")]
+    DatabaseIo(#[from] std::io::Error),
+
+    /// A (non-I/O related) database error.
+    #[error("database error: {0}")]
+    DatabaseError(RuntimeError),
+}
```
hinto-janai, line 28: Change to `cuprate_rpc_interface::RpcError`. The database(s) return `RuntimeError` in their `Service` signatures so we forward that.

This also means that `cuprate_rpc_interface` depends on `cuprate_database`, it doesn't seem ideal to pull all of it in for just the error type though, especially since none of the errors other than IO can happen at this level.

`cuprated` already pulls everything in anyway, but this means other `RpcHandler` impls will have to pull in `cuprate-database` too.

rpc/interface/src/rpc_service.rs
```diff
@@ -32,7 +32,7 @@ pub trait RpcService<Request, Response>:
         Request,
         Response = Response,
         Error = RpcError,
-        Future: Future<Output = Result<Response, RpcError>> + Send + Sync + 'static,
+        Future: Future<Output = Result<Response, RpcError>> + Send + 'static,
```
hinto-janai, line 35: https://github.com/Cuprate/cuprate/pull/266#discussion_r1745799658 and also `BoxFuture` isn't `Sync` so we have to remove this.

binaries/cuprated/src/rpc/handler.rs
```diff
@@ -0,0 +1,102 @@
...
+};
+use futures::{channel::oneshot::channel, future::BoxFuture};
+use serde::{Deserialize, Serialize};
+use tower::Service;
+
+use cuprate_blockchain::service::BlockchainReadHandle;
+use cuprate_helper::asynch::InfallibleOneshotReceiver;
+use cuprate_json_rpc::Id;
+use cuprate_rpc_interface::{RpcError, RpcHandler};
+use cuprate_txpool::service::TxpoolReadHandle;
```
Boog900, line 18: `cuprate_rpc_types` should be with the second group here

binaries/cuprated/src/rpc/json.rs
```diff
@@ -0,0 +1,293 @@
...
+    GetInfoResponse, GetLastBlockHeaderRequest, GetLastBlockHeaderResponse, GetMinerDataRequest,
+    GetMinerDataResponse, GetOutputHistogramRequest, GetOutputHistogramResponse,
+    GetTransactionPoolBacklogRequest, GetTransactionPoolBacklogResponse, GetTxIdsLooseRequest,
+    GetTxIdsLooseResponse, GetVersionRequest, GetVersionResponse, HardForkInfoRequest,
+    HardForkInfoResponse, JsonRpcRequest, JsonRpcResponse, OnGetBlockHashRequest,
+    OnGetBlockHashResponse, PruneBlockchainRequest, PruneBlockchainResponse, RelayTxRequest,
+    RelayTxResponse, SetBansRequest, SetBansResponse, SubmitBlockRequest, SubmitBlockResponse,
+    SyncInfoRequest, SyncInfoResponse,
+};
+use tower::ServiceExt;
```
Boog900, line 23: the tower import here should be moved as well

rpc/interface/src/rpc_error.rs
Boog900: Would it worth just making this a `BoxError`? that's what I have done elsewhere, like in the consensus crate.

hinto-janai: If we're going to be (ab)using opaque errors internally, what about using `anyhow::Error`?

[It's basically a better `Box<dyn Error + Send + Sync + 'static>`](https://www.reddit.com/r/rust/comments/17neomp/comment/k7rhrss/).

Boog900: Yeah sure that's probably a good idea.

rpc/interface/src/rpc_service.rs
```diff
@@ -31,8 +29,8 @@ pub trait RpcService<Request, Response>:
     + Service<
         Request,
         Response = Response,
-        Error = RpcError,
-        Future: Future<Output = Result<Response, RpcError>> + Send + 'static,
+        Error = anyhow::Error,
+        Future: Future<Output = Result<Response, anyhow::Error>> + Send + 'static,
```
hinto-janai, line 33: `RpcError` deleted, `RpcService`s must now return `anyhow::Error`.

rpc/interface/src/route/bin.rs
```diff
@@ -66,8 +73,16 @@ macro_rules! generate_endpoints_inner {
     ($variant:ident, $handler:ident, $request:expr) => {
         paste::paste! {
             {
+                // Check if restricted.
+                if [<$variant Request>]::IS_RESTRICTED && $handler.restricted() {
+                    // TODO: mimic `monerod` behavior.
+                    return Err(StatusCode::FORBIDDEN);
+                }
```
hinto-janai, line 80: Unrelated: binary requests now check `restricted`.

All the binary requests are `!restricted` but adding this for the future is probably a good idea.

Boog900 (approved):

Issue #167, opened 2024-06-17
Casting into/from pointer-sized integers without `unwrap()`


hinto-janai:
## What
If Cuprate only supports 64-bit hardware, it means that `usize` can be losslessly casted into `u64` and vice-versa. Same goes for `isize` <-> `i64`.

We could also allow `u32` -> `usize`, as Rust doesn't impl anything higher than `From<u16> for usize`.

## Why
Prevents needless `unwrap()`; is more explicit.

## Where
This will replace any code where `try_{into,from}()` -> `unwrap()` is used, e.g.:

https://github.com/Cuprate/cuprate/blob/c837f2f48e2f99e10c81e790afe5d5e8d866df3f/helper/src/map.rs#L80

https://github.com/Cuprate/cuprate/blob/c837f2f48e2f99e10c81e790afe5d5e8d866df3f/consensus/src/context/hardforks.rs#L82

https://github.com/Cuprate/cuprate/blob/c837f2f48e2f99e10c81e790afe5d5e8d866df3f/helper/src/map.rs#L100

https://github.com/Cuprate/cuprate/blob/c837f2f48e2f99e10c81e790afe5d5e8d866df3f/consensus/src/context/weight.rs#L154

## How
Implementation can be free functions:
```rust
#[inline(always)]
fn u64_to_usize(u: u64) -> usize {
    u as usize
}

#[inline(always)]
fn usize_to_u64(u: usize) -> u64 {
    u as u64
}

// repeat for isize, i64
```
Safety is upheld by using a `#[cfg]` option where the functions are implemented:
```rust
// in {lib.mod}.rs
#[cfg(not(target_pointer_width = "64"))]
compile_error!("Cuprate is only compatible with 64-bit CPUs");
```
This prevents any user of this function from compiling on any non-64-bit architecture.

Using `as` directly would also work, but it means `as` will be more prevalent in the codebase. `as` can be used in a lossly manner, so it muddies things and is less explicit, e.g.:
```rust
let i: i64 = -1;
let u = i as u64; // lossy
let u = usize_to_u64(i); // compile error
```

PR #264, opened 2024-09-02
helper: add and use `cast` module

hinto-janai:
### What
Adds a `cuprate_helper::cast` module and uses it throughout the codebase.

Closes https://github.com/Cuprate/cuprate/issues/167.

### DIFF
--- a/net/wire/src/p2p.rs
+++ b/net/wire/src/p2p.rs
@@ -99,7 +99,7 @@ impl LevinCommandTrait for LevinCommand {
             LevinCommand::FluffyMissingTxsRequest => 1024 * 1024,  // 1 MB
             LevinCommand::GetTxPoolCompliment => 1024 * 1024 * 4,  // 4 MB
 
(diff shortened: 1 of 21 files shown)


net/wire/src/p2p.rs
```diff
@@ -99,7 +99,7 @@ impl LevinCommandTrait for LevinCommand {
             LevinCommand::FluffyMissingTxsRequest => 1024 * 1024,  // 1 MB
             LevinCommand::GetTxPoolCompliment => 1024 * 1024 * 4,  // 4 MB
 
-            LevinCommand::Unknown(_) => usize::MAX.try_into().unwrap_or(u64::MAX),
+            LevinCommand::Unknown(_) => u64::MAX,
```
hinto-janai, line 102: Not a cast but if we are only supporting 64-bit this is always `u64::MAX`.

Boog900 (approved):

PR #267, opened 2024-09-03
books/architecture: port database design document

hinto-janai:
### What
Ports [`storage/blockchain/DESIGN.md`](https://github.com/Cuprate/cuprate/blob/0941f68efcd7dfe66124ad0c1934277f47da9090/storage/blockchain/DESIGN.md) to the architecture book and updates some things.


books/architecture/src/storage/db/issues/unaligned.md
```diff
@@ -0,0 +1,26 @@
...
+```
+but this would require supporting types that must be copied regardless with the occasional `&[u8]` that can be returned without casting. This was hard to do so in a generic way, thus all `[u8]`'s are copied and returned as owned `StorableVec`s.
+
+This is a _tradeoff_ `cuprate_database` takes as:
+- `bytemuck::pod_read_unaligned` is cheap enough
+- The main API, `service`, needs to return owned value anyway
+- Having no references removes a lot of lifetime complexity
+
+The alternative is either:
+- Using proper (de)serialization instead of casting (which comes with its own costs)
```
Boog900, line 25: I don't think this is a solution - it would still require copying the data

Boog900 (commented):
Just one comment, sorry for taking so long

Boog900 (approved):

Issue #176, opened 2024-06-19
Handle inbound connection pings.


Boog900:
Currently if we don't have an available permits for inbound connections we drop the connection completely, we should wait to see if the connection just wanted to send us a ping to see if we are reachable.

Here is where this should happen:

https://github.com/Cuprate/cuprate/blob/4653ac58849c81b6ab993a1d23f061a97962524b/p2p/p2p/src/inbound_server.rs#L106

We spawn a separate task to wait for the first message to see if it is a ping, then respond if it is and close the connection.

We should also limit the number of concurrent pings to 2 and we should also have a reasonable timeout so we are not waiting for too long.

PR #268, opened 2024-09-05
book/architecture: add resource index

hinto-janai:
### What
Adds an index section for common resources in the `Resources` section, similar to the crates index we already have.

Setting this up now and gradually editing it is probably easier than doing it all at once.


books/architecture/src/SUMMARY.md
```diff
@@ -93,17 +93,20 @@
...
-        - [⚪️ The block downloader](resource-model/concurrency-and-parallelism/the-block-downloader.md)
-        - [⚪️ The verifier](resource-model/concurrency-and-parallelism/the-verifier.md)
-        - [⚪️ Thread exit](resource-model/concurrency-and-parallelism/thread-exit.md)
+- [⚪️ Resources](resources/intro.md)
+    - [⚪️ File system](resources/fs/intro.md)
+        - [🟡 Index of PATHs](resources/fs/paths.md)
+    - [⚪️ Sockets](resources/sockets/index.md)
+        - [🔴 Index of ports](resources/sockets/ports.md)
+    - [⚪️ Memory](resources/memory.md)
+    - [🟡 Concurrency and parallelism](resources/cap/intro.md)
```
hinto-janai, line 102: `concurrency-and-parallelism/` is very long so -> `cap/`

books/architecture/src/resources/cap/threads.md
```diff
@@ -0,0 +1,2 @@
+# Index of threads
+This is an index of all of the system threads Cuprate actively uses.
```
hinto-janai, line 2: Currently, is this just `global rayon` + `global tokio` + `blockchain` + `txpool`?

@Boog900 are there any threads (ephemeral/persistent) spawned in any of your code? Or any code that we call out to?

books/architecture/src/resources/sockets/ports.md
```diff
@@ -0,0 +1,2 @@
+# Index of ports
+This is an index of all of the network sockets Cuprate actively uses.
```
hinto-janai, line 2: Don't think we open any right now, right? Maybe networking ones in the alpha binary?

books/architecture/src/resources/fs/paths.md
```diff
@@ -0,0 +1,87 @@
+# Index of PATHs
```
hinto-janai, line 1: Do we only save database files currently?

books/architecture/src/resources/cap/threads.md
Boog900: We are going to have a thread that listens to Stdin and handles commands. Also the `blockchain` + `txpool` will use the same threadpool

books/architecture/src/resources/fs/paths.md
Boog900: we also save a `p2p_state` file, which I should add to `helper`

books/architecture/src/resources/sockets/ports.md
Boog900: Yeah not yet, we would need #176. For the initial alpha binary I would be ok with not accepting inbound connections

Boog900 (approved):

PR #273, opened 2024-09-09
P2P: Implement incoming ping request handling over maximum inbound limit

Asurar0:
### What

Closes: #176 

This PR implements the following changes:

- If the maximum inbound connection semaphore reach its limit, `inbound_server` fn will open a tokio task to check if the node wanted to ping us. If it is the case we respond, otherwise drop the connection.
- Added some documentation to the `inbound_server` fn.

### DIFF
--- a/p2p/p2p/src/inbound_server.rs
+++ b/p2p/p2p/src/inbound_server.rs
@@ -4,9 +4,10 @@
 //! them to the handshaker service and then adds them to the client pool.
 use std::{pin::pin, sync::Arc};
 
-use futures::StreamExt;
+use futures::{SinkExt, StreamExt};
 use tokio::{
     sync::Semaphore,
+    task::JoinSet,
     time::{sleep, timeout},
 };
 use tower::{Service, ServiceExt};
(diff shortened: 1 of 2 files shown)


p2p/p2p/src/inbound_server.rs
```diff
@@ -102,8 +114,38 @@ where
...
+                                    )).into())
+                                    .await
+                                    .inspect_err(|err: &BucketError| tracing::debug!("Unable to respond to ping request from peer ({addr}): {err}"));
+                            }
+                        }
+                    }
+                    .instrument(Span::current())
+                );
+                
+                task.await.expect("A tokio task responsible for handling ping requests panicked!");
```
Boog900, line 147: By awaiting the task here we block the inbound server from processing other connections.

I would recommend spawning these tasks on a [JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#) and checking the length to limit concurrent pings instead of using a semaphore.

Boog900 (changes requested):
Thanks for working on this!

p2p/p2p/src/inbound_server.rs
Asurar0: Fixed in latest commit.

p2p/p2p/src/inbound_server.rs
```diff
@@ -2,12 +2,12 @@
 //!
 //! This module contains the inbound connection server, which listens for inbound connections, gives
 //! them to the handshaker service and then adds them to the client pool.
-use std::{pin::pin, sync::Arc};
+use std::{pin::pin, sync::Arc, time::Duration};
 
-use futures::StreamExt;
+use cuprate_wire::{admin::{PingResponse, PING_OK_RESPONSE_STATUS_TEXT}, AdminRequestMessage, AdminResponseMessage, BucketError, Message};
```
Boog900, line 7: cna you move this import new to `cuprate_p2p_core` please

p2p/p2p/src/inbound_server.rs
```diff
@@ -102,8 +114,36 @@ where
...
+            // Otherwise check if the node is simply pinging us.
             tracing::debug!("No permit free for incoming connection.");
-            // TODO: listen for if the peer is just trying to ping us to see if we are reachable.
+            
+            // We only handle 2 ping request conccurently. Otherwise we drop the connection immediately.
+            if ping_join_set.len() < 2 {
+                ping_join_set.spawn(
+                    async move {
+                        // Await first message from node. If it is a ping request we respond back, otherwise we drop the connection.
+                        let fut = timeout(Duration::from_secs(5), peer_stream.next());
```
Boog900, line 125: Also can you move the 2 & 5 here to `constants`  with appropriate names for each please

PR #275, opened 2024-09-09
Storage: use `saturating_add` for `cumulative_generated_coins`

Boog900:
### Why

This amount overflows due to tail emission, the consensus cache already does this.


hinto-janai (approved):
Hmm... what does `monerod` actually store here? `u64::MAX`?

Then it returns something like `u64::MAX + calculate_cumulative_tail_emission()`?

PR #287, opened 2024-09-18
p2p/dandelion-tower: enable workspace lints

hinto-janai:
### What
Enables and fixes workspace lints for `cuprate-dandelion-tower`.

### DIFF
--- a/p2p/dandelion-tower/src/pool/manager.rs
+++ b/p2p/dandelion-tower/src/pool/manager.rs
@@ -88,8 +88,7 @@ where
                 .insert(peer.clone());
         }
 
-        let state = from
-            .map_or(TxState::Local, |from| TxState::Stem { from });
+        let state = from.map_or(TxState::Local, |from| TxState::Stem { from });
 
         let fut = self
             .dandelion_router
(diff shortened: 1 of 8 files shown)


p2p/dandelion-tower/src/pool/manager.rs
```diff
@@ -280,11 +278,13 @@ where
...
 
                     if let Err(e) = self.handle_incoming_tx(tx, routing_state, tx_id).await {
+                        #[expect(clippy::let_underscore_must_use, reason = "TODO")]
                         let _ = res_tx.send(());
 
                         tracing::error!("Error handling transaction in dandelion pool: {e}");
                         return;
                     }
+                    #[expect(clippy::let_underscore_must_use, reason = "TODO")]
                     let _ = res_tx.send(());
```
hinto-janai, line 288: What is the reason channel errors can be ignored here?

Boog900: An error just means the requestor dropped the response channel, it doesn't effect the behaviour of the service.

Boog900 (approved):

PR #288, opened 2024-09-18
p2p/p2p-core: enable workspace lints

hinto-janai:
### What
Enables and fixes workspace lints for `cuprate-p2p-core`.


hinto-janai:
@Boog900 do you know why this error is happening? https://github.com/Cuprate/cuprate/actions/runs/10913364837/job/30289741378?pr=288#step:11:261

`p2p/p2p-core/tests` was moved to `p2p/p2p-core/src/tests`.

Boog900:
It's because `cuprate-test-utils` depends on `cuprate-p2p-core`: https://doc.rust-lang.org/cargo/reference/resolver.html#dev-dependency-cycles

> In the case of building library unit tests, there are actually two copies of the library linked into the final test binary

Basically `cuprate-p2p-core` is first complied by it self, without the tests, in this crate a trait is defined: `NetworkZone`. `cuprate-test-utils` is then built with `TestNetZone` impl'ing the trait from `cuprate-p2p-core`. Next `cuprate-p2p-core` is built again but this time including the tests, because it is built again technically the trait `NetworkZone` is a different trait than the `NetworkZone` in the original build.

hinto-janai:
Moved `tests/` back to an outer folder.

Boog900 (approved):

PR #292, opened 2024-09-18
levin: enable workspace lints

hinto-janai:
### What
Enables and fixes workspace lints for `cuprate-levin`.


net/levin/src/lib.rs
```diff
@@ -33,6 +33,16 @@
...
 
+// Used in `tests/`.
+#[cfg(test)]
+use futures as _;
+#[cfg(test)]
+use proptest as _;
+#[cfg(test)]
+use rand as _;
+#[cfg(test)]
+use tokio as _;
```
hinto-janai, line 44: Tests being laid out like they are in `consensus/`, `net/`, `p2p/` is pretty painful for the unused dependency lint since `tests/` can't be detected. 

For now I've done this and `#![expect(clippy::tests_outside_test_module, unused_crate_dependencies)]` in the `tests/` files for now but surely there's a better solution that isn't removing the lint.

Boog900: I wouldn't want to lose the ability of doing proper integration tests.

Sadly I think this is going to be the cost of using semi-stable lints. IMO `tests_outside_test_module` could probably be disabled, `unused_crate_dependencies` is nice though.

You could put this in brackets which would make this slightly better:

```suggestion
// Used in `tests/`.
#[cfg(test)]
use futures as _;
#[cfg(test)]
use proptest as _;
#[cfg(test)]
use rand as _;
#[cfg(test)]
use tokio as _;
```
```suggestion
// Used in `tests/`.
#[cfg(test)]
{
    use futures as _;
    use proptest as _;
    use rand as _;
    use tokio as _;
}
```

Boog900: Or we remove certain lints for tests, so remove the `--all-targets` flag on our normal clippy run and allow warnings for tests?

hinto-janai: Removed `tests_outside_test_module`.

Boog900 (approved):

PR #295, opened 2024-09-18
consensus: enable workspace lints

hinto-janai:
### What
Enables and fixes workspace lints for the crates inside `consensus/`:
- `cuprate-consensus`
- `cuprate-consensus-rules` 
- `cuprate-fast-sync`

Also makes `cuprate-fast-sync` use `cuprate_helper::u64_to_usize` instead of `as usize`.


consensus/fast-sync/Cargo.toml
```diff
@@ -9,19 +9,22 @@ name = "cuprate-fast-sync-create-hashes"
...
-tokio = { workspace = true, features = ["full"] }
-tower.workspace = true
+cuprate-types           = { path = "../../types" }
+cuprate-helper          = { path = "../../helper", features = ["cast"] }
+
+clap         = { workspace = true, features = ["derive", "std"] }
+hex          = { workspace = true }
+hex-literal  = { workspace = true }
+monero-serai = { workspace = true }
+sha3         = { version = "0.10.8" }
```
hinto-janai, line 22: TODO: this can be shared by `cuprate-cryptonight`

consensus/fast-sync/src/fast_sync.rs
```diff
@@ -6,8 +6,6 @@ use std::{
     task::{Context, Poll},
 };
 
-#[allow(unused_imports)]
-use hex_literal::hex;
```
hinto-janai, line 10: This is used in the `hashes_of_hashes` file but `clippy` can't detect external files like this.

consensus/rules/src/blocks.rs
```diff
@@ -122,10 +122,10 @@ pub fn check_block_pow(hash: &[u8; 32], difficulty: u128) -> Result<(), BlockErr
 /// Returns the penalty free zone
 ///
 /// <https://cuprate.github.io/monero-book/consensus_rules/blocks/weight_limit.html#penalty-free-zone>
-pub fn penalty_free_zone(hf: &HardFork) -> usize {
-    if hf == &HardFork::V1 {
+pub fn penalty_free_zone(hf: HardFork) -> usize {
+    if hf == HardFork::V1 {
```
hinto-janai, line 126: Probably >50% of the diffs in this PR are from changing `&HardFork` signatures to `HardFork`.

`HardFork` is 1 byte so it's cheaper to copy it rather than creating/using an 8 byte pointer to it.

consensus/src/context/difficulty.rs
```diff
@@ -232,26 +232,22 @@ impl DifficultyCache {
...
             let last_cum_diff = cumulative_difficulties.back().copied().unwrap_or(1);
             cumulative_difficulties.push_back(last_cum_diff + *difficulties.last().unwrap());
 
             if timestamps.len() > self.config.total_block_count() {
-                diff_info_popped.push((
-                    timestamps.pop_front().unwrap(),
-                    cumulative_difficulties.pop_front().unwrap(),
-                ));
+                timestamps.pop_front().unwrap();
+                cumulative_difficulties.pop_front().unwrap();
```
hinto-janai, line 243: This `diff_info_popped` was removed by clippy because it isn't being read/used after creation.

types/src/block_complete_entry.rs
```diff
@@ -1,7 +1,6 @@
 //! Contains [`BlockCompleteEntry`] and the related types.
 
 //---------------------------------------------------------------------------------------------------- Import
-#[cfg(feature = "epee")]
 use bytes::Bytes;
```
hinto-janai, line 4: `Bytes` is used in a field that _isn't_ gated by `epee`, so this needs to be removed.

consensus/src/block/alt_block.rs
```diff
@@ -36,8 +36,8 @@ use crate::{
 ///
 /// Returns [`AltBlockInformation`], which contains the cumulative difficulty of the alt chain.
 ///
-/// This function only checks the block's PoW and its weight.
-pub async fn sanity_check_alt_block<C>(
+/// This function only checks the block's `PoW` and its weight.
```
Boog900, line 39:
```suggestion
/// This function only checks the block's proof-of-work and its weight.
```

consensus/src/context/difficulty.rs
Boog900: Ah this must have been left over from something

consensus/src/context/rx_vms.rs
```diff
@@ -1,6 +1,6 @@
-//! RandomX VM Cache
+//! `RandomX` VM Cache
 //!
-//! This module keeps track of the RandomX VM to calculate the next blocks PoW, if the block needs a randomX VM and potentially
+//! This module keeps track of the `RandomX` VM to calculate the next blocks `PoW`, if the block needs a randomX VM and potentially
```
Boog900, line 3: Should PoW be in ``?

hinto-janai: No but unfortunately it triggers the CamelCase doc lint, changed to proof-of-work.

Boog900 (approved):

PR #296, opened 2024-09-18
`cargo upgrade`

hinto-janai:
### What
Updates yanked versions of:
- `bytes`
- `bytemuck`

notified from `cargo deny` @ https://github.com/Cuprate/cuprate/actions/runs/10929670932/job/30340882757.

The rest of the semver compatible updates are from a normal [`cargo upgrade`](https://github.com/killercup/cargo-edit/blob/master/README.md#cargo-upgrade).


hinto-janai:
@Boog900 this doesn't fix the `cargo deny` error at https://github.com/Cuprate/cuprate/actions/runs/10929670932/job/30340882757#step:5:2064.

It stems from an unmaintained dependency used by `borsh`, not sure what we can/should do.

Boog900:
Either we can add it to the ignore list or wait for borsh to update I guess

Boog900 (approved):
I'll merge this as a one off, however for the future the way we specify versions actually means `>=` (roughly anyway), so this isn't really needed as none of these have actually been bumped: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

Issue #279, opened 2024-09-11
Internal data needed for `cuprated`'s RPC


hinto-janai:
## What
This is a discussion for what data is required for `cuprated`'s RPC system.

Notably, this is:
- Blockchain data
- Transaction pool data
- P2P data
- General process state

The following tables (except the general process state) are newly proposed requests/responses to be added to each system's `tower::Service` API such that the RPC system can fulfill the request. The RPC system is assumed to have read access (and write access where needed) to all systems via `tower::Service` handles.

## Blockchain

| Request                   | Response                       | Needed for              | Description |
|---------------------------|--------------------------------|-------------------------|-------------|
| `PopBlocks(u64)`          | `u64`                          | `/pop_blocks`           | Pop `n` blocks, return the new height
| `Prune`                   | TODO                           | `prune_blockchain`      | Prune the blockchain
| `Pruned`                  | `bool`                         | `prune_blockchain`      | Is the blockchain pruned?
| `Block(u64)`              | `Block`                        | `get_block`             | Retrieve block information
| `BlockByHash([u8; 32])`   | `Block`                        | `get_block`             | Same as `Block`; for convenience
| `TopBlock`                | `Block`                        |                         | Get the latest `Block`, Same as `Block`; for convenience
| `TopBlockExtendedHeader`  | `ExtendedBlockHeader`          |                         | Get the latest `ExtendedBlockHeader`; for convenience
| `KeyImageSpent([u8; 32])` | `bool`                         | `/is_key_image_spent`   | Singular version of already existing `KeyImagesSpent`
| `TopBlockFull`            | `(Block, ExtendedBlockHeader)` | `get_last_block_header` | Get information on the last block to map to `cuprate_rpc_types::misc::BlockHeader`
| `TopBlockHash`            | `[u8; 32]`                     | `get_info`              | Get the top block hash.
| `TotalTxCount`            | `u64`                          | `get_info`              | Get the total amount of non-coinbase transactions; `cuprate_rpc_types::json::GetInfoResponse::tx_count`
| `DatabaseSize`            | `(u64, u64)`                   | `get_info` | Return byte size of database file and available space for `cuprate_rpc_types::json::GetInfoResponse::{database_size,free_space}`
| `BlockWeightLimit`        | `u64`                          | `get_info` | `cuprate_rpc_types::json::GetInfoResponse::block_weight_limit`
[...]

PR #274, opened 2024-09-09
cuprated: Blockchain Manager

Boog900:
### What

This PR contains the blockchain manager, there are still quite a few TODOs however I think it is in a state to be reviewed and merged.

More details in comments

PR #297, opened 2024-09-23
cuprated: internal signatures required for RPC

hinto-janai:
### What
Adds the `{function, request, response}` signatures needed by `cuprated`'s RPC system to the `{blockchain, txpool, p2p, blockchain_context}` APIs.

Described in https://github.com/Cuprate/cuprate/issues/279.

This covers the signatures needed for all RPC requests (JSON-RPC, binary, other).

### DIFF
--- a/binaries/cuprated/src/rpc/handler.rs
+++ b/binaries/cuprated/src/rpc/handler.rs
@@ -7,7 +7,7 @@ use futures::{channel::oneshot::channel, future::BoxFuture};
 use serde::{Deserialize, Serialize};
 use tower::Service;
 
-use cuprate_blockchain::service::BlockchainReadHandle;
+use cuprate_blockchain::service::{BlockchainReadHandle, BlockchainWriteHandle};
 use cuprate_helper::asynch::InfallibleOneshotReceiver;
 use cuprate_json_rpc::Id;
 use cuprate_rpc_interface::RpcHandler;
@@ -16,20 +16,13 @@ use cuprate_rpc_types::{
     json::{JsonRpcRequest, JsonRpcResponse},
     other::{OtherRequest, OtherResponse},
 };
-use cuprate_txpool::service::TxpoolReadHandle;
+use cuprate_txpool::service::{TxpoolReadHandle, TxpoolWriteHandle};
 
 use crate::rpc::{bin, json, other};
 
 /// TODO
 #[derive(Clone)]
 pub struct CupratedRpcHandler {
-    /// Should this RPC server be [restricted](RpcHandler::restricted)?
-    //
-    // INVARIANT:
-    // We don't need to include this in `state` and check for
-    // `self.is_restricted()` because `cuprate-rpc-interface` handles that.
-    pub restricted: bool,
-
     /// State needed for request -> response mapping.
     pub state: CupratedRpcHandlerState,
 }
@@ -37,11 +30,24 @@ pub struct CupratedRpcHandler {
 /// TODO
 #[derive(Clone)]
 pub struct CupratedRpcHandlerState {
+    /// Should this RPC server be [restricted](RpcHandler::restricted)?
+    //
+    // INVARIANT:
+    // We don't need to include this in `state` and check for
+    // `self.is_restricted()` because `cuprate-rpc-interface` handles that.
+    pub restricted: bool,
+
     /// Read handle to the blockchain database.
-    pub blockchain: BlockchainReadHandle,
+    pub blockchain_read: BlockchainReadHandle,
+
+    /// Write handle to the blockchain database.
+    pub blockchain_write: BlockchainWriteHandle,
 
     /// Read handle to the transaction pool database.
-    pub txpool: TxpoolReadHandle,
+    pub txpool_read: TxpoolReadHandle,
+
+    /// Write handle to the transaction pool database.
+    pub txpool_write: TxpoolWriteHandle,
 }
 
 impl CupratedRpcHandler {
@@ -53,7 +59,7 @@ impl CupratedRpcHandler {
 
 impl RpcHandler for CupratedRpcHandler {
     fn restricted(&self) -> bool {
-        self.restricted
+        self.state.restricted
     }
 }
 
--- a/binaries/cuprated/src/rpc/request.rs
+++ b/binaries/cuprated/src/rpc/request.rs
@@ -1,0 +1,18 @@
+//! Convenience functions for requests/responses.
+//!
+//! This module implements many methods on
+//! [`CupratedRpcHandlerState`](crate::rpc::CupratedRpcHandlerState)
+//! that are simple wrappers around the request/response API provided
+//! by the multiple [`tower::Service`]s.
+//!
+//! These exist to prevent noise like `unreachable!()`
+//! from being everywhere in the actual handler functions.
+//!
+//! Each module implements methods for a specific API, e.g.
+//! the [`blockchain`] modules contains methods for the
+//! blockchain database [`tower::Service`] API.
+
+mod blockchain;
+mod blockchain_manager;
+mod p2p;
+mod txpool;
--- a/storage/blockchain/src/service/mod.rs
+++ b/storage/blockchain/src/service/mod.rs
@@ -21,10 +21,6 @@
 //!
 //! The 2nd allows any caller to send [`WriteRequest`][req_w]s.
 //!
-//! The `DatabaseReadHandle` can be shared as it is cheaply [`Clone`]able, however,
-//! the `DatabaseWriteHandle` cannot be cloned. There is only 1 place in Cuprate that
-//! writes, so it is passed there and used.
-//!
 //! ## Initialization
 //! The database & thread-pool system can be initialized with [`init()`].
 //!
--- a/storage/txpool/src/service/interface.rs
+++ b/storage/txpool/src/service/interface.rs
@@ -29,6 +29,7 @@ pub enum TxpoolReadResponse {
 
 //---------------------------------------------------------------------------------------------------- TxpoolWriteRequest
 /// The transaction pool [`tower::Service`] write request type.
+#[derive(Clone)]
 pub enum TxpoolWriteRequest {
     /// Add a transaction to the pool.
     ///
@@ -49,7 +50,7 @@ pub enum TxpoolWriteRequest {
 
 //---------------------------------------------------------------------------------------------------- TxpoolWriteResponse
 /// The transaction pool [`tower::Service`] write response type.
-#[derive(Debug, Ord, PartialOrd, Eq, PartialEq)]
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub enum TxpoolWriteResponse {
     /// A [`TxpoolWriteRequest::AddTransaction`] response.
     ///
--- a/binaries/cuprated/src/rpc.rs
+++ b/binaries/cuprated/src/rpc.rs
@@ -6,5 +6,6 @@ mod bin;
 mod handler;
 mod json;
 mod other;
+mod request;
 
 pub use handler::{CupratedRpcHandler, CupratedRpcHandlerState};
--- a/binaries/cuprated/src/rpc/request/blockchain.rs
+++ b/binaries/cuprated/src/rpc/request/blockchain.rs
@@ -1,0 +1,366 @@
+//! Functions for [`BlockchainReadRequest`] and [`BlockchainWriteRequest`].
+
+use std::{
+    collections::{HashMap, HashSet},
+    ops::Range,
+    sync::Arc,
+};
+
+use anyhow::{anyhow, Error};
+use futures::StreamExt;
+use monero_serai::block::Block;
+use tower::{Service, ServiceExt};
+
+use cuprate_consensus::BlockchainResponse;
+use cuprate_helper::{
+    cast::{u64_to_usize, usize_to_u64},
+    map::split_u128_into_low_high_bits,
+};
+use cuprate_types::{
+    blockchain::{BlockchainReadRequest, BlockchainWriteRequest},
+    Chain, ExtendedBlockHeader, HardFork, OutputOnChain, VerifiedBlockInformation,
+};
+
+use crate::rpc::CupratedRpcHandlerState;
+
+impl CupratedRpcHandlerState {
+    /// [`BlockchainReadRequest::BlockExtendedHeader`].
+    pub(super) async fn block_extended_header(
+        &mut self,
+        height: u64,
+    ) -> Result<ExtendedBlockHeader, Error> {
+        let BlockchainResponse::BlockExtendedHeader(header) = self
+            .blockchain_read
+            .ready()
+            .await?
+            .call(BlockchainReadRequest::BlockExtendedHeader(u64_to_usize(
+                height,
+            )))
+            .await?
+        else {
+            unreachable!();
+        };
+
+        Ok(header)
+    }
+
+    /// [`BlockchainReadRequest::BlockHash`].
+    pub(super) async fn block_hash(
+        &mut self,
+        height: u64,
+        chain: Chain,
+    ) -> Result<[u8; 32], Error> {
+        let BlockchainResponse::BlockHash(hash) = self
+            .blockchain_read
+            .ready()
+            .await?
+            .call(BlockchainReadRequest::BlockHash(
+                u64_to_usize(height),
+                chain,
+            ))
+            .await?
+        else {
+            unreachable!();
+        };
+
+        Ok(hash)
+    }
+
+    /// [`BlockchainReadRequest::FindBlock`].
+    pub(super) async fn find_block(
+        &mut self,
+        block_hash: [u8; 32],
+    ) -> Result<Option<(Chain, usize)>, Error> {
+        let BlockchainResponse::FindBlock(option) = self
+            .blockchain_read
+            .ready()
+            .await?
+            .call(BlockchainReadRequest::FindBlock(block_hash))
+            .await?
+        else {
+            unreachable!();
+        };
+
+        Ok(option)
+    }
+
+    /// [`BlockchainReadRequest::FilterUnknownHashes`].
+    pub(super) async fn filter_unknown_hashes(
+        &mut self,
+        block_hashes: HashSet<[u8; 32]>,
+    ) -> Result<HashSet<[u8; 32]>, Error> {
+        let BlockchainResponse::FilterUnknownHashes(output) = self
+            .blockchain_read
+            .ready()
+            .await?
+            .call(BlockchainReadRequest::FilterUnknownHashes(block_hashes))
+            .await?
+        else {
+            unreachable!();
+        };
+
+        Ok(output)
+    }
+
+    /// [`BlockchainReadRequest::BlockExtendedHeaderInRange`]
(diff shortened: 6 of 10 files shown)


binaries/cuprated/src/rpc/handler.rs
```diff
@@ -3,51 +3,136 @@
...
+    /// Handle to the blockchain manager.
+    pub blockchain_manager: BlockchainManagerHandle,
 
     /// Read handle to the transaction pool database.
-    pub txpool: TxpoolReadHandle,
+    pub txpool_read: TxpoolReadHandle,
+
+    /// Write handle to the transaction pool database.
+    pub txpool_write: TxpoolWriteHandle,
 }
```
hinto-janai, line 116: ### Misc
We actually need access to `restricted` within the handlers which means there's no reason to separate `CupratedRpcHandler` into a `state: CupratedRpcHandlerState` anymore for cheaper cloning since we need the whole thing.

You can ignore all the changes in `binaries/cuprated/src/rpc/{json,bin,other}.rs`, they're just updating the input from `CupratedRpcHandlerState` -> `CupratedRpcHandler`.

binaries/cuprated/src/rpc/request.rs
```diff
@@ -0,0 +1,19 @@
...
+//!
+//! Each module implements methods for a specific API, e.g.
+//! the [`blockchain`] modules contains methods for the
+//! blockchain database [`tower::Service`] API.
+
+mod address_book;
+mod blockchain;
+mod blockchain_context;
+mod blockchain_manager;
+mod txpool;
```
hinto-janai, line 19: ### Misc

You can also mostly ignore `binaries/cuprated/src/rpc/request/`, these are just convenience function that just call + unwrap request/responses so there isn't this type of code everywhere:
```rust
let Response::Variant(data) = blockchain_read
    .ready()
    .await?
    .call(Request::Variant)
    .await?
else {
    unreachable!();
};

Ok(data)
```

consensus/src/context.rs
```diff
@@ -221,24 +221,29 @@ pub struct NewBlockData {
 #[derive(Debug, Clone)]
 pub enum BlockChainContextRequest {
     /// Get the current blockchain context.
-    GetContext,
+    Context,
+
     /// Gets the current `RandomX` VM.
-    GetCurrentRxVm,
+    CurrentRxVm,
```
hinto-janai, line 227: ### Misc
Renamed these to match other variants.

storage/blockchain/src/service/mod.rs
```diff
@@ -21,10 +21,6 @@
 //!
 //! The 2nd allows any caller to send [`WriteRequest`][req_w]s.
 //!
-//! The `DatabaseReadHandle` can be shared as it is cheaply [`Clone`]able, however,
-//! the `DatabaseWriteHandle` cannot be cloned. There is only 1 place in Cuprate that
-//! writes, so it is passed there and used.
-//!
```
hinto-janai, line 27: ### Misc
Not true anymore, we need `Clone` on `CupratedRpcHandler` and thus the `DatabaseWriteHandle`.

binaries/cuprated/src/rpc/request/address_book.rs
```diff
@@ -0,0 +1,104 @@
...
+        .expect("TODO")
+        .call(AddressBookRequest::PeerlistSize)
+        .await
+        .expect("TODO")
+    else {
+        unreachable!();
+    };
+
+    Ok((usize_to_u64(white), usize_to_u64(grey)))
+}
```
hinto-janai, line 30: ### Misc
The services that use `tower::BoxError` don't work with `anyhow::Error` when using `?`, can we swap them to `anyhow`?

The services are:
- address book
- blockchain context
- transaction pool

binaries/cuprated/src/rpc/handler.rs
```diff
@@ -3,51 +3,136 @@
...
+
+    /// Response to [`BlockchainManagerRequest::TargetHeight`]
+    TargetHeight { height: usize },
+}
+
+/// TODO: use real type when public.
+pub type BlockchainManagerHandle = cuprate_database_service::DatabaseReadService<
+    BlockchainManagerRequest,
+    BlockchainManagerResponse,
+>;
```
hinto-janai, line 92: ### Signatures (blockchain manager)
Since this doesn't exist in-tree yet, there are just temporary types so signatures line up.

consensus/src/context.rs
```diff
@@ -248,8 +253,22 @@ pub enum BlockChainContextRequest {
...
+    FeeEstimate {
+        /// TODO
+        grace_blocks: u64,
+    },
+
     /// Clear the alt chain context caches.
     ClearAltCache,
+
+    /// Get information on all the current alternate chains.
+    AltChains,
```
hinto-janai, line 270: ### Signatures (blockchain context request)
Added to the existing `BlockChainContextRequest`.

consensus/src/context.rs
```diff
@@ -310,22 +333,49 @@ pub enum BlockChainContextRequest {
...
     AltChainDifficultyCache(DifficultyCache),
+
     /// A randomX VM for an alt chain.
     AltChainRxVM(Arc<RandomXVm>),
+
     /// A weight cache for an alt chain
     AltChainWeightCache(BlockWeightsCache),
-    /// A generic Ok response.
-    Ok,
 }
```
hinto-janai, line 379: ### Signatures (blockchain context response)
Added to the existing `BlockChainContextResponse`.

p2p/p2p-core/src/services.rs
```diff
@@ -86,34 +90,77 @@ pub enum AddressBookRequest<Z: NetworkZone> {
...
+
+    /// (Un)ban a peer.
+    SetBan(SetBan<Z::Addr>),
+
     /// Checks if the given peer is banned.
-    IsPeerBanned(Z::Addr),
+    GetBan(Z::Addr),
+
+    /// Get the state of all bans.
+    GetBans,
```
hinto-janai, line 131: ### Signatures (address book request)
Added to the existing `AddressBookRequest`.

`IsPeerBanned` can be replaced by `GetBan` (see below)

p2p/p2p-core/src/services.rs
```diff
@@ -86,34 +90,77 @@ pub enum AddressBookRequest<Z: NetworkZone> {
...
+
+    /// Response to [`AddressBookRequest::GetBan`].
+    ///
+    /// This returns [`None`] if the peer is not banned,
+    /// else it returns how long the peer is banned for.
+    GetBan { unban_instant: Option<Instant> },
+
+    /// Response to [`AddressBookRequest::GetBans`].
+    GetBans(Vec<BanState<Z::Addr>>),
 }
```
hinto-janai, line 166: ### Signatures (address book response)
Added to the existing `AddressBookResponse`.

RPC's `GetBan` needs the UNIX time where a peer is unbanned, since the peer might not be banned, this call can also function as a `IsPeerBanned` with `unban_instant.is_some()`.

storage/txpool/src/service/interface.rs
```diff
@@ -5,30 +5,53 @@ use std::sync::Arc;
...
+    ///
+    /// The inner `Vec` contains information on all
+    /// the transactions currently in the pool.
+    Backlog(Vec<TxEntry>),
+
+    /// Response to [`TxpoolReadRequest::Size`].
+    ///
+    /// The inner value is the amount of
+    /// transactions currently in the pool.
+    Size(usize),
```
hinto-janai, line 49: ### Signatures (txpool request/request)
Added to the existing `TxpoolRead{Request,Response}`.

types/src/blockchain.rs
```diff
@@ -103,6 +106,40 @@ pub enum BlockchainReadRequest {
...
+    /// `N` last blocks starting at particular height.
+    ///
+    /// TODO: document fields after impl.
+    CoinbaseTxSum {
+        height: usize,
+        count: u64,
+    },
+
+    /// Get the necessary data to create a custom block template.
+    MinerData,
```
hinto-janai, line 142: ### Signatures (blockchain request)
Added to the existing `BlockchainReadRequest`.

types/src/blockchain.rs
```diff
@@ -215,18 +253,47 @@ pub enum BlockchainResponse {
...
+    Difficulty(u128),
+
+    /// Response to [`BlockchainReadRequest::OutputHistogram`].
+    OutputHistogram(Vec<OutputHistogramEntry>),
+
+    /// Response to [`BlockchainReadRequest::CoinbaseTxSum`].
+    CoinbaseTxSum(CoinbaseTxSum),
+
+    /// Response to [`BlockchainReadRequest::MinerData`].
+    MinerData(MinerData),
```
hinto-janai, line 295: ### Signatures (blockchain response)
Added to the existing `BlockchainResponse`.

p2p/p2p-core/src/ban.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub ban: bool,
+    pub duration: Duration,
+}
+
+/// Data within [`crate::services::AddressBookResponse::GetBans`].
+pub struct BanState<A: NetZoneAddress> {
+    pub address: A,
+    pub banned: bool,
+    pub unban_instant: Instant,
+}
```
hinto-janai, line 19: ### Types (address book)
Some of the requests/responses in this PR require defining some types.

A lot of these are 1-1 with the RPC types but I decided on making more specific versions because:
- Pulling in `cuprate-rpc-types` everywhere seems wrong
- The RPC types are JSON-like, so a `[u8; 32]` field will be something like `String`, adding these types lets us use the canonical type internally

Some of the types are defined in the crate where the request/response is defined, many are defined in `cuprate-types`.

storage/txpool/src/tx.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Used in [`TxpoolReadResponse::Backlog`](crate::service::interface::TxpoolReadResponse::Backlog).
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct TxEntry {
+    /// The transaction's weight.
+    pub weight: u64,
+    /// The transaction's fee.
+    pub fee: u64,
+    /// How long the transaction has been in the pool.
+    pub time_in_pool: std::time::Duration,
+}
```
hinto-janai, line 14: ### Types (txpool)

types/src/types.rs
```diff
@@ -155,6 +147,104 @@ pub struct OutputOnChain {
...
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ChainInfo {
+    pub block_hash: [u8; 32],
+    pub block_hashes: Vec<[u8; 32]>,
+    pub difficulty: u128,
+    pub height: u64,
+    pub length: u64,
+    pub main_chain_parent_block: [u8; 32],
+    pub wide_difficulty: u128,
+}
```
hinto-janai, line 246: ### Types (everything else)

storage/blockchain/src/service/read.rs
```diff
@@ -601,3 +617,46 @@ fn alt_blocks_in_chain(env: &ConcreteEnv, chain_id: ChainId) -> ResponseResult {
 
     Ok(BlockchainResponse::AltBlocksInChain(blocks))
 }
+
+/// [`BlockchainReadRequest::Block`]
+fn block(env: &ConcreteEnv, block_height: BlockHeight) -> ResponseResult {
+    Ok(BlockchainResponse::Block(todo!()))
+}
```
hinto-janai, line 624: ### TODO
All the new request/responses are fully typed but they will all panic with `todo!()`.

The exception is `AddressBookRequest::GetBan` which I added a new function for.

Handlers can be added in a different PR.

consensus/src/context.rs
```diff
@@ -310,22 +333,49 @@ pub enum BlockChainContextRequest {
...
+    /// - [`BlockChainContextRequest::AddAltChainContextCache`]
+    Ok,
+
+    /// Response to [`BlockChainContextRequest::Context`]
     Context(BlockChainContext),
+
+    /// Response to [`BlockChainContextRequest::CurrentRxVm`]
+    ///
     /// A map of seed height to `RandomX` VMs.
     RxVms(HashMap<usize, Arc<RandomXVm>>),
```
hinto-janai, line 352: ### TODO
Why does this return a `HashMap` when the request is `CurrentRxVm`?

I might be misremembering but I think somewhere in RPC we need access to a RandomX VM, would calling `.last()` on this `HashMap` always give us the latest in a `Some`?

hinto-janai (commented):
This review is split like such:

| Title       | Purpose |
|-------------|---------|
| Misc        | Miscellaneous changes added and/or needed in the future   
| Signatures  | New requests/responses added to our `tower::Service`s 
| Types       | New types added (for usage in the request/responses)
| TODO        | Things to do after this PR

Questions that need answering:
- Are any of the requests/responses impossible to fulfill?
- Can any of the requests/responses be done in a better way?
- Do any of the requests/responses belong to a different `Service`?

binaries/cuprated/src/rpc/handler.rs
```diff
@@ -3,51 +3,136 @@
...
+
+    /// Handle to the blockchain manager.
+    pub blockchain_manager: BlockchainManagerHandle,
 
     /// Read handle to the transaction pool database.
-    pub txpool: TxpoolReadHandle,
+    pub txpool_read: TxpoolReadHandle,
+
+    /// Write handle to the transaction pool database.
+    pub txpool_write: TxpoolWriteHandle,
```
hinto-janai, line 115: @Boog900 some questions about https://github.com/Cuprate/cuprate/issues/279#issuecomment-2351034124?:

> The transaction pool is going to have a specific interface for requests that mutate the pool, just to keep in mind, it wont directly use the write handle.

- What should be here in `txpool_write`?
- Is "specific interface" just going to be a different `tower::Service`?
- Why does the txpool need this special interface?

binaries/cuprated/src/rpc/handler.rs
```diff
@@ -3,51 +3,136 @@
...
+    PopBlocks { amount: usize },
+
+    /// Start pruning the blockchain.
+    Prune,
+
+    /// Is the blockchain pruned?
+    Pruned,
+
+    /// Relay a block to the network.
+    RelayBlock(Block),
```
Boog900, line 37: This should be add block, this also happens to be the one request I already have in the blockchain manager PR (`handle_incoming_block`).

binaries/cuprated/src/rpc/handler.rs
Boog900: > What should be here in txpool_write?
> Is "specific interface" just going to be a different tower::Service?

Yes, so a handle to that other service.

> Why does the txpool need this special interface?

To synchronise tx-pool state in a way where we aren't verifying the same tx across multiple connections before adding them to the pool. There are other operations that require more synchronization than a database write handle can provide.

binaries/cuprated/src/rpc/request/address_book.rs
Boog900: Yeah, there are other as well, probably should be done in a separate PR

consensus/src/context.rs
Boog900: > Why does this return a HashMap when the request is CurrentRxVm?

It should probably be `CurrentRxVms`, it needs to return more than one so we can compute PoW concurrently.

> I might be misremembering but I think somewhere in RPC we need access to a RandomX VM

You will probably just need to generate a new VM - I don't think you will need the current active one, I may be wrong though I just can't see why you would.

p2p/p2p-core/src/ban.rs
Boog900: Yeah separating them is a good idea IMO

Boog900: Those types seem weird though, wouldn't this make more sense: 

```suggestion
//! Data structures related to bans.

use std::time::{Duration, Instant};

use crate::NetZoneAddress;

/// Data within [`crate::services::AddressBookRequest::SetBan`].
pub struct SetBan<A: NetZoneAddress> {
    pub address: A,
    pub ban: Option<Duration>,
}

/// Data within [`crate::services::AddressBookResponse::GetBans`].
pub struct BanState<A: NetZoneAddress> {
    pub address: A,
    pub banned: Option<Instant>,
}
```

storage/blockchain/src/service/mod.rs
Boog900: This does need to be removed for the tx-pool handle, however the database is still only written to in 1 place, the blockchain manager.

You shouldn't need to hold a write handle to the blockchain in the RPC

types/src/blockchain.rs
Boog900: `Difficulty` is contained in the blockchain context so shouldn't need a new request (my bad for missing this).

`MinerData` will not be able to be a single request. Looking at its data it would need a request to the blockchain cache and to the tx-pool, both of these requests are already there/in this PR so this could just be removed I think

binaries/cuprated/src/rpc/handler.rs
hinto-janai: OK, can these be converted after both this and #274 are merged? Are all the signatures here possible?

binaries/cuprated/src/rpc/handler.rs
hinto-janai: Does it have the same semantics though? It adds a block to our local chain _and_ broadcasts it?

The usecase is [`submit_block`](https://github.com/monero-project/monero/blob/9866a0e9021e2422d8055731a586083eb5e2be67/src/rpc/core_rpc_server.cpp#L2218).

binaries/cuprated/src/rpc/handler.rs
```diff
@@ -102,36 +102,31 @@ pub struct CupratedRpcHandler {
...
     /// Handle to the blockchain manager.
     pub blockchain_manager: BlockchainManagerHandle,
 
     /// Read handle to the transaction pool database.
     pub txpool_read: TxpoolReadHandle,
 
-    /// Write handle to the transaction pool database.
-    pub txpool_write: TxpoolWriteHandle,
+    /// TODO: handle to txpool service.
+    pub txpool_manager: std::convert::Infallible,
```
hinto-janai, line 112: TODO: replace with real interface when public.

binaries/cuprated/src/rpc/handler.rs
Boog900: yep

p2p/p2p-core/src/ban.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// - If [`None`], the peer will be unbanned
+    pub ban: Option<Duration>,
+}
+
+/// Data within [`crate::services::AddressBookResponse::GetBans`].
+pub struct BanState<A: NetZoneAddress> {
+    /// Address of the peer.
+    pub address: A,
+    /// - If [`Some`], when this peer will be unbanned
+    /// - If [`None`], the peer is not currently banned
```
Boog900, line 21:
```suggestion
    /// - If [`Some`], the peer is banned until this [`Instant`]
    /// - If [`None`], the peer is not currently banned
```

Boog900 (approved):
The blockchain manager API should change but for now is fine

Boog900 (approved):

PR #299, opened 2024-09-25
P2P: remove peer sync service

Boog900:
### What

Removes the peer sync service and instead keeps core sync data stored in `PeerInformation`.

### Why

Removes the need for a separate service.


p2p/p2p/src/client_pool.rs
```diff
@@ -127,6 +127,32 @@ impl<N: NetworkZone> ClientPool<N> {
     ) -> impl Iterator<Item = ClientPoolDropGuard<N>> + sealed::Captures<(&'a (), &'b ())> {
         peers.iter().filter_map(|peer| self.borrow_client(peer))
     }
+
+    /// Borrows all [`Client`]s from the pool that have claimed a higher cumulative difficulty than
+    /// the amount passed in.
+    ///
+    /// The [`Client`]s are wrapped in [`ClientPoolDropGuard`] which
+    /// will return the clients to the pool when they are dropped.
+    pub fn clients_with_more_cumulative_difficulty(
```
Boog900, line 136: This is the new way for the block downloader to get peers to sync from.

This function is horribly expensive, however in a future PR I have some ideas on how to make it cheaper

p2p/p2p-core/src/client.rs
```diff
@@ -59,8 +60,13 @@ pub struct PeerInformation<A> {
...
     /// The direction of this connection (inbound|outbound).
     pub direction: ConnectionDirection,
-    /// The peers pruning seed.
+    /// The peer's [`PruningSeed`].
     pub pruning_seed: PruningSeed,
+    /// The [`CoreSyncData`] of this peer.
+    ///
+    /// Data across fields are not necessarily related, so [`CoreSyncData::top_id`] is not always the
+    /// block hash for the block at height one below [`CoreSyncData::current_height`].
+    pub core_sync_data: Arc<Mutex<CoreSyncData>>,
```
hinto-janai, line 69: If this lock isn't temporary I would really prefer if it were accompanied with a brief description of its scope; how, where, and who the lock should be used by.

p2p/p2p/src/client_pool.rs
Boog900: > Is it expensive because of the Mutex?

Yeah and the `DashMap` which is a bunch of `RwLocks`

> Also, would the below have the same lifetime semantics? I don't think Captures is needed anymore right?

It's not however I would rather keep it as it is as the current way allows more uses.

p2p/p2p-core/src/client/handshaker.rs
```diff
@@ -477,19 +454,19 @@ where
         id: addr,
         handle,
         direction,
-        pruning_seed,
+        pruning_seed: PruningSeed::decompress_p2p_rules(peer_core_sync.pruning_seed)?,
```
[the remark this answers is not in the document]
Boog900: oops ..

Issue #199, opened 2024-06-27
ZMQ Pub/Sub design proposal


Boog900:
## What

The proposal covers integrating [`monerod`'s ZMQ Pub/Sub system](https://github.com/monero-project/monero/blob/master/docs/ZMQ.md#zmq-pubsub), it does not cover the seperate ZMQ RPC system.

## How

`monerod` uses ZMQ `XPUB` sockets, to allow clients to filter the data they do not need and so `monerod` only has to create data that has active subscribers for.

This means we can't use the `zeromq` crate (the ZMQ Rust rewrite) as it does not yet support XPUB sockets, so we will have to use the `zmq` crate (Rust bindings). 

### Message types

The message types from Monero will need to be defined, this should be done in its own crate `cuprate-zmq-types` in `zmq/types`.

[This file](https://github.com/monero-project/monero/blame/master/src/rpc/zmq_pub.cpp) contains some of these types.

### Filtering

As described in [monerod's ZMQ docs](https://github.com/monero-project/monero/blob/master/docs/ZMQ.md#zmq-pubsub) filtering is done on multiple levels, format, context, and event.

`zmq` will handle routing the messages to the right clients, however we need a way to keep track of the amount of subscribers for each type so we know if we actually need to send a certain message or not.

This will be done by a `Subscribers` struct which will look something like this:

```rust
struct Subscribers {
    numb_subscribers: BTreeMap<MessageType, u64>,
}

#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
enum MessageType {
    // The order of these variants must not change

    JsonMinimalChainMain,
    JsonMinimalTxPoolAdd,

    JsonFullChainMain,
    JsonFullTxPoolAdd,
    JsonFullMinerData,
}

#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum Filter {
    All,

    JsonMinimal,
    JsonMinimalChainMain,
    JsonMinimalTxPoolAdd,

    JsonFull,
    JsonFullChainMain,
    JsonFullTxPoolAdd,
    JsonFullMinerData,
}

// by implementing `RangeBounds<MessageType>` it allows us to do `.range(Filter::XX)` on a btree
// tests should be added to make sure every Filter returns the correct range. 

impl RangeBounds<MessageType> for Filter {
    fn start_bound(&self) -> Bound<&MessageType> {
        match self {
            Filter::All => Unbounded,
            Filter::JsonMinimal => Unbounded,
[...]

PR #300, opened 2024-09-25
types: JSON representation types

hinto-janai:
### What
Adds types that allow easier JSON representations of a block, transaction, and other common types.

### Why
These JSON representations are used in RPC/ZMQ.

https://github.com/Cuprate/cuprate/issues/199#issuecomment-2373982398

### Where
`cuprate-types`.


types/src/json/block.rs
```diff
@@ -0,0 +1,299 @@
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Block {
+    pub major_version: u8,
+    pub minor_version: u8,
+    pub timestamp: u64,
+    pub prev_id: String,
+    pub nonce: u32,
+    pub miner_tx: MinerTransaction,
+    pub tx_hashes: Vec<String>,
+}
```
hinto-janai, line 22: JSON block type.

types/src/json/tx.rs
```diff
@@ -0,0 +1,526 @@
...
+        signatures: Vec<String>,
+    },
+    V2 {
+        /// This field is [flattened](https://serde.rs/field-attrs.html#flatten).
+        #[serde(flatten)]
+        prefix: TransactionPrefix,
+        rct_signatures: RctSignatures,
+        rctsig_prunable: RctSigPrunable,
+    },
+}
```
hinto-janai, line 35: JSON transaction type.

types/src/json/block.rs
```diff
@@ -0,0 +1,299 @@
...
+        prefix: MinerTransactionPrefix,
+        signatures: [(); 0],
+    },
+    V2 {
+        /// This field is [flattened](https://serde.rs/field-attrs.html#flatten).
+        #[serde(flatten)]
+        prefix: MinerTransactionPrefix,
+        rct_signatures: MinerTransactionRctSignatures,
+    },
+}
```
hinto-janai, line 41: A separate type exists for miner transactions because:

- Fields are slightly different
- The `Input` JSON is different
- It allows us to enforce `signature` always being empty in miner transactions

rpc/types/src/json.rs
```diff
@@ -635,7 +635,9 @@ define_request_and_response! {
     AccessResponseBase {
         blob: String,
         block_header: BlockHeader,
-        json: String, // FIXME: this should be defined in a struct, it has many fields.
+        /// `cuprate_rpc_types::json::block::Block` should be used
+        /// to create this JSON string in a type-safe manner.
+        json: String,
```
hinto-janai, line 640: I thought we would be replacing embedded JSON within RPC responses with these types but I realized: it's not actual JSON, it's a string promising it contains JSON.

I.e. the RPC response is:
```json
{
  "json": "{\n  \"version\": 1, \n  \"unlock_time\": 0}"
}
```
not:
```json
{
  "json": { "version": 1, "unlock_time": 0 }
}
```

I don't think there's any reason it shouldn't be actual JSON other than some internal `monerod` detail making that hard to do. Anyway, we have to keep this a `String` but we can still use the JSON types to be more type-safe.

helper/README.md
```diff
@@ -6,12 +6,16 @@ This allows all workspace crates to share, and aids compile times.
 If a 3rd party's crate/functions/types are small enough, it could be moved here to trim dependencies and allow easy modifications.
 
 ## Features
-Code can be selectively used/compiled with cargo's `--feature` or `features = ["..."]`.
+Modules can be selectively used/compiled with cargo's `--feature` or `features = ["..."]`.
```
hinto-janai, line 9: Unrelated: `cuprate-helper` doc update.

rpc/types/src/misc/tx_entry.rs
```diff
@@ -71,32 +71,24 @@ use cuprate_epee_encoding::{
 pub enum TxEntry {
     /// This entry exists in the transaction pool.
     InPool {
-        as_hex: String,
-        as_json: String,
+        /// This field is [flattened](https://serde.rs/field-attrs.html#flatten).
+        #[cfg_attr(feature = "serde", serde(flatten))]
+        prefix: TxEntryPrefix,
```
hinto-janai, line 76: Unrelated: putting common `TxEntry` variants fields in a prefix struct.

rpc/types/src/json.rs
hinto-janai: Or we could add:
- `#[serde(serialize_with = "type_to_string")]`
- `#[serde(deserialize_with = "string_to_type")]`

to automatically do this extra step?

Boog900: It would probably be easier to just keep this as a `String`

types/src/json/tx.rs
```diff
@@ -0,0 +1,618 @@
...
+                })
+                .collect();
+
+            let unlock_time = match prefix.additional_timelock {
+                transaction::Timelock::None => height,
+                transaction::Timelock::Block(height_lock) => height + usize_to_u64(height_lock),
+                transaction::Timelock::Time(seconds) => {
+                    height + (seconds / usize_to_u64(monero_serai::BLOCK_TIME))
+                }
+            } + usize_to_u64(monero_serai::DEFAULT_LOCK_WINDOW);
```
Boog900, line 114: `monero_serai` calls `unlock_time` `additional_timelock` so the values should be exactly the same - like how it's done in the DB.

The way these are differentiated is `Timelock::Time` is any value above `500_000_000`

Boog900: just to be clear this means that the inner values are what is used you don't need to do any processing

Boog900: and for `None` this should be `0`

types/src/json/tx.rs
```diff
@@ -0,0 +1,1447 @@
...
...a3d0f56a2d6302106e47b6b1b50d","08a9283d8b34426eb7b7547fa8fb1573430b99f1c119f2ff9612e82acee98e03","250d2ac44e26782f293eca3deb70fc5c52cb942166b1efb2f78ec32640e02d06","1bc1bcc3de357a4652c03815e59e14cb13668946366746dca3dad2f4c44c9000","9f8b446e373e3e19871f22b9bc95017d4411e555477afb34114b428c8296470a","e49d0313e969fb8c4e451388309280a96b8e3216fa1e28ab2efab49f38e86f07","0cee07c99293507ad558838f2fa07af1c4ddc86886658c6207c1f25f343afb06","39bd17be3aaaeda4fb8aa8dafcf5748581f7bb8b92b0dfe3add14a8481570003","0048e1ca905806551cd210c40356cc80935a98f63163a087ea89585915e8770d","3c46eea5308dbdff7376d89378998376cb722d08604d6ecb2b3cb795f91dc607","7d13be56b2e858d2fca81b3a6b0312943d33e501b4e09814818edb96fb28aa0c","313a2021350abd25bd79c22ea33fe071367da2e208913d788d50101c90f0e702"],"c1":"9d96220cd0d49340e06b915f7204cd1f68c4c2069389bf4c011b4fa6c24c0d02","D":"91d00727ba3655729ce88981e803967946403e970f0a6749625f59d4e6d7ebc9"}],"pseudoOuts":["a6785a3aca529db1da40944bb1826519d7caaa31f4549e6854cb97e5234d3e8e","f5cda4db5f83f1c1edea0b66461d1848daf01054c24a690e1438add59dc4f206","dff30968b66355b9c7890db508692e9620c999e0025ca9395fa53732e6432606","6b78d37b63714ebe1d09981766c61a07bf0bfbc9fc7f7a8998396aa99d43e0cc"]}}"#;
```
hinto-janai, line 1020: A rough estimate of stack usage within these types: this transaction contains `HexBytes32` 104 times, 86 of them are in `Vec`, so only `[u8; 32 * 18]` bytes (576) are on the stack.

A single block containing 200 transactions puts `~1.152` kilobytes of hexadecimal data on the stack.

I don't think we'll run into stack overflows unless we're somehow holding many many blocks (500-1000) in a stack frame at once? I think it'll be safe to use `HexBytes` within RPC types as well.

Maybe a bigger problem is that `size_of::<Transaction>() == 240`.

types/src/hex.rs
```diff
@@ -0,0 +1,70 @@
...
+			impl Default for [<HexBytes $array_len>] {
+				fn default() -> Self {
+					Self([0; $array_len])
+				}
+			}
+		)*
+	}};
+}
+
+generate_hex_array!(1, 8, 32, 64);
```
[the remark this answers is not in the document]
hinto-janai: nvm re: `HexBytes<const N: usize>`.

I fixed it, see below.

types/src/hex.rs
```diff
@@ -0,0 +1,74 @@
...
+    [u8; N]: hex::FromHex,
+    <[u8; N] as hex::FromHex>::Error: std::fmt::Display,
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(Self(hex::serde::deserialize(deserializer)?))
+    }
+}
```
hinto-janai, line 36: `HexBytes` now uses `<const N: usize>`, deserialization is only blanket implemented if the array supports `hex::serde::deserialization`.

types/src/json/tx.rs
```diff
@@ -0,0 +1,1455 @@
...
+                .outputs
+                .into_iter()
+                .map(|o| {
+                    let amount = o.amount.unwrap_or(0);
+
+                    let target = match o.view_tag {
+                        Some(view_tag) => {
+                            let tagged_key = TaggedKey {
+                                key: HexBytes::<32>(o.key.0),
+                                view_tag: HexBytes::<1>([view_tag]),
```
Boog900, line 96: Nit: you shouldn't need to specify the const generic on every use Rust should be able to infer it

types/src/json/tx.rs
Boog900: > size_of::<Transaction>() == 240

This is the size of the stack data right? 

> A single block containing 200 transactions puts ~1.152 kilobytes of hexadecimal data on the stack.

Surely the data should be in `Vec`s right? so the amount of txs wouldn't effect the stack usage of `Block`

Boog900 (approved):

types/src/json/tx.rs
hinto-janai: > This is the size of the stack data right?

Yes.

> Surely the data should be in Vecs right?

Yeah I don't think there's any place we would be handling that many blocks without some type of heap container, so it should be fine.

PR #301, opened 2024-09-27
cuprated: add `constants` & `statics` modules

hinto-janai:
### What
Adds a `constants` and `statics` module to `cuprated`.

- `constants` contains general constants used by  `cuprated`, currently it contains version strings roughly equivalent to [`version.h`](https://github.com/monero-project/monero/blob/a1dc85c5373a30f14aaf7dcfdd95f5a7375d3623/src/version.h)
- `statics` contains global `static` data used throughout `cuprated`

Some data (e.g. `const DEBUG: bool`) will be available in [`cuprate_constants`](https://github.com/Cuprate/cuprate/pull/280) so it isn't defined in `constants`.

### Why
- `statics::START_INSTANT_UNIX` is needed by RPC for [`get_block` -> start_time](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_block)
- `constants::VERSION_BUILD` is needed by RPC for [`get_block` -> version](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_block)


binaries/cuprated/src/statics.rs
```diff
@@ -0,0 +1,53 @@
...
+define_init_lazylock_statics! {
+    /// The start time of `cuprated`.
+    START_INSTANT: SystemTime = SystemTime::now();
+
+    /// Start time of `cuprated` as a UNIX timestamp.
+    START_INSTANT_UNIX: u64 = START_INSTANT
+        .duration_since(UNIX_EPOCH)
+        .expect("Failed to set `cuprated` startup time.")
+        .as_secs();
+}
```
hinto-janai, line 41: I have little faith that people will manually update things so newly added `static`s that should be initialized before anything else go in this macro, which will generate a function that initializes everything.

binaries/cuprated/src/main.rs
```diff
@@ -13,10 +13,16 @@
...
 mod config;
+mod constants;
 mod p2p;
 mod rpc;
+mod statics;
 mod txpool;
 
 fn main() {
+    // Initialize global static `LazyLock` data.
+    statics::init_lazylock_statics();
```
hinto-janai, line 24: The statics get initialized here.

We could add this to a broader `init_everything()` later that setups up logging as too.

binaries/cuprated/src/statics.rs
hinto-janai: (people including us in the far future btw)

binaries/cuprated/src/main.rs
Boog900: for now this will be ok

binaries/cuprated/src/statics.rs
```diff
@@ -0,0 +1,53 @@
...
+		$( #[$attr:meta] )*
+		$name:ident: $t:ty = $init_fn:expr;
+	)*) => {
+    	/// Initialize global static `LazyLock` data.
+		pub fn init_lazylock_statics() {
+			$(
+				LazyLock::force(&$name);
+			)*
+		}
+
```
Boog900, line 24: tabs here should be changed to spaces

Boog900 (changes requested):

Boog900 (approved):

PR #302, opened 2024-09-27
P2P: give the protocol handler access to the peer info

Boog900:
### What
Changes the p2p crates to allow giving the protocol handler access to the peer info.

### Why
It is useful to know the peers sync state when handling protocol requests.

### Where
Mainly the handshaker/handshaker builder.

### How
Uses the `tower::make::*` stuff, this allows us to provide an async interface for creating request handlers, which is going to be useful for validating core sync data when handshaking is complete


Cargo.toml
```diff
@@ -81,7 +81,7 @@ thread_local          = { version = "1.1.8", default-features = false }
 tokio-util            = { version = "0.7.12", default-features = false }
 tokio-stream          = { version = "0.1.16", default-features = false }
 tokio                 = { version = "1.40.0", default-features = false }
-tower                 = { version = "0.4.13", default-features = false }
+tower                 = { git = "https://github.com/Cuprate/tower.git", rev = "6c7faf0", default-features = false }
```
hinto-janai, line 84:
```suggestion
tower                 = { git = "https://github.com/Cuprate/tower.git", rev = "6c7faf0", default-features = false } // <https://github.com/tower-rs/tower/pull/796>
```

p2p/p2p-core/src/client/handshaker/builder.rs
```diff
@@ -60,7 +62,10 @@ impl<N: NetworkZone> HandshakerBuilder<N> {
             address_book: DummyAddressBook,
             core_sync_svc: DummyCoreSyncSvc::static_mainnet_genesis(),
             peer_sync_svc: DummyPeerSyncSvc,
-            protocol_request_svc: DummyProtocolRequestHandler,
+            protocol_request_svc_maker: MapErr::new(
+                Shared::new(DummyProtocolRequestHandler),
+                tower::BoxError::from,
+            ),
```
hinto-janai, line 68: This infallible is being wrapped and returned in an error to satisfy trait bounds, right?

hinto-janai (approved):

PR #304, opened 2024-10-06
cuprated: config & args

Boog900:
### What
Starts the config and arguments to cuprated.


binaries/cuprated/src/config/Cuprate.toml
Boog900, line 1: This file contains a config file that `cuprated` can generate to make it easier for users to define their own config files.

The type of this file is `.toml` although technically it's not as we do Rust string formatting to change the paths so they are correct for all platforms. I think making it a `.toml` anyway is probably ok.

binaries/cuprated/src/config/args.rs
```diff
@@ -0,0 +1,45 @@
+use std::{io::Write, path::PathBuf};
+
+use clap::builder::TypedValueParser;
+
+use cuprate_helper::network::Network;
+
+use crate::config::{default::create_default_config_file, Config, DEFAULT_CONFIG_FILE_NAME};
+
+#[derive(clap::Parser, Debug)]
+pub struct Args {
```
Boog900, line 10: The args for Cuprate - only a few are defined for now.

binaries/cuprated/src/config/default.rs
```diff
@@ -0,0 +1,69 @@
...
+        blockchain = CUPRATE_BLOCKCHAIN_DIR.to_string_lossy()
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::config::{default::generate_config_text, Config};
+
+    #[test]
+    fn generate_config_text_covers_all_values() {
```
Boog900, line 51: Some tests to make sure the generated config file is correct and contains all values.

Boog900 (commented):
Example generated config file:

```toml
#     ____                      _
#    / ___|   _ _ __  _ __ __ _| |_ ___
#   | |  | | | | '_ \| '__/ _` | __/ _ \
#   | |__| |_| | |_) | | | (_| | ||  __/
#    \____\__,_| .__/|_|  \__,_|\__\___|
#              |_|
#

## The network to run on, valid values: "Mainnet", "Testnet", "Stagenet".
network = "Mainnet"

## Tracing config.
[tracing]
## The minimum level for log events to be displayed.
level = "info"

## Clear-net config.
[p2p.clear_net]
## The number of outbound connections we should make and maintain.
outbound_connections = 64
## The number of extra connections we should make under load from the rest of Cuprate, i.e. when syncing.
extra_outbound_connections = 8
## The maximum number of incoming we should allow.
max_inbound_connections = 128
## The percent of outbound connections that should be to nodes we have not connected to before.
gray_peers_percent = 0.7
## The port to accept connections on, if left `0` no connections will be accepted.
p2p_port = 0
## The IP address to listen to connections on.
server.ip = "0.0.0.0"

## The Clear-net addressbook config.
[p2p.clear_net.address_book_config]
## The size of the white peer list, which contains peers we have made a connection to before.
max_white_list_length = 1_000
## The size of the gray peer list, which contains peers we have not made a connection to before.
max_gray_list_length = 5_000
## The folder to store the address book.
peer_store_folder = "/home/user/.cache/cuprate"
## The amount of time between address book saves.
peer_save_period = { secs = 90, nanos = 0 }

## The block downloader config.
[p2p.block_downloader]
## The size of the buffer of sequential blocks waiting to be verified and added to the chain (bytes).
buffer_size = 50_000_000
## The size of the queue of blocks which are waiting for a parent block to be downloaded (bytes).
in_progress_queue_size = 50_000_000
## The target size of a batch of blocks (bytes), must not exceed 100MB.
target_batch_size = 5_000_000
## The number of blocks in the first bacth (you probably shouldn't change this).
initial_batch_len = 1
## The amount of time between checking the pool of connected peers for free peers to download blocks.
check_client_pool_interval = { secs = 30, nanos = 0 }

## Storage config
[storage]
## The amount of reader threads to spawn.
reader_threads = "OnePerThread"

## Txpool storage config.
[storage.txpool]
## The txpool storage location.
path = "/home/user/.local/share/cuprate/txpool"
## The database sync mode for the txpool.
sync_mode = "Async"
## The maximum size of all the txs in the pool (bytes).
max_txpool_size = 100_000_000

## Blockchain storage config.
[storage.blockchain]
## The blockchain storage location.
path = "/home/user/.local/share/cuprate/blockchain"
## The database sync mode for the blockchain.
sync_mode = "Async"
```

storage/blockchain/src/config.rs
```diff
@@ -59,8 +62,10 @@ pub use cuprate_database_service::ReaderThreads;
 #[derive(Debug, Clone, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct ConfigBuilder {
+    network: Network,
+
     /// [`Config::db_directory`].
-    db_directory: Option<Cow<'static, Path>>,
+    db_directory: Option<PathBuf>,
```
Boog900, line 68: I also made changes to the blockchain & txpool database so we can handle switching networks more easily.

binaries/cuprated/Cargo.toml
```diff
@@ -72,13 +72,8 @@ tokio-stream          = { workspace = true }
...
+toml = "0.8"
 
 [lints]
 workspace = true
-
-[profile.dev]
-panic = "abort"
-
-[profile.release]
-panic = "abort"
```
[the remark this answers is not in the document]
SyntheticBird45: see https://github.com/kayabaNerve/fcmp-plus-plus/blob/develop/Cargo.toml for more exciting compilation profiles

binaries/cuprated/src/config/Cuprate.toml
SyntheticBird45: TOML is the superior format and everyone disagreeing must be extinguished. I'm more than okay with it

helper/src/network.rs
```diff
@@ -38,3 +44,29 @@ impl Network {
...
+
+impl FromStr for Network {
+    type Err = ParseNetworkError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "mainnet" => Ok(Self::Mainnet),
+            "testnet" => Ok(Self::Testnet),
+            "stagenet" => Ok(Self::Stagenet),
+            _ => Err(ParseNetworkError),
```
SyntheticBird45, line 59: really nit but could we add `main` `test` and `stage` as possible values ?

p2p/address-book/src/lib.rs
```diff
@@ -29,12 +31,24 @@ pub struct AddressBookConfig {
...
 }
 
+#[cfg(feature = "serde_config")]
+impl Default for AddressBookConfig {
+    fn default() -> Self {
+        Self {
+            max_white_list_length: 1000,
+            max_gray_list_length: 5000,
+            peer_store_folder: cuprate_helper::fs::CUPRATE_CACHE_DIR.clone(),
+            peer_save_period: Duration::from_secs(90),
```
SyntheticBird45, line 47: if its the default is there any corresponding constant in `cuprate-constants` ?

helper/src/network.rs
Boog900: We could but then we would also have to allow it in the clap args, which would then display 6 possible values to users which could be confusing.

p2p/address-book/src/lib.rs
Boog900: Nope, should I? there are a lot of constant values that I just define in the default method as that's the only place they are used, defining them elsewhere is more annoying IMO

binaries/cuprated/src/config.rs
```diff
@@ -1 +1,145 @@
...
+/// Reads the args & config file, returning a [`Config`].
+pub fn read_config_and_args() -> Config {
+    let args = args::Args::parse();
+
+    let config: Config = if let Some(config_file) = &args.config_file {
+        // If a config file was set in the args try read it and exit if we can't.
+        match Config::read_from_file(config_file) {
+            Ok(config) => config,
+            Err(e) => {
+                tracing::error!("Failed to read config from file: {}", e);
```
hinto-janai, line 39:
```suggestion
        // If a config file was set in the args try to read it and exit if we can't.
        match Config::read_from_file(config_file) {
            Ok(config) => config,
            Err(e) => {
                tracing::error!("Failed to read config from file: {e}");
```

binaries/cuprated/src/config.rs
```diff
@@ -1 +1,145 @@
...
+    /// The network we should run on.
+    network: Network,
+
+    /// [`tracing`] config.
+    tracing: TracingConfig,
+
+    /// The P2P network config.
+    p2p: P2PConfig,
+
+    /// The Storage config
```
hinto-janai, line 77:
```suggestion
    /// The storage config.
```

binaries/cuprated/src/config.rs
```diff
@@ -1 +1,145 @@
...
+    storage: StorageConfig,
+}
+
+impl Config {
+    /// Attempts to read a config file in [`toml`] format from the given [`Path`.
```
hinto-janai, line 82:
```suggestion
    /// Attempts to read a config file in [`toml`] format from the given [`Path`].
```

binaries/cuprated/src/config/Cuprate.toml
```diff
@@ -0,0 +1,75 @@
...
+
+## The block downloader config.
+[p2p.block_downloader]
+## The size of the buffer of sequential blocks waiting to be verified and added to the chain (bytes).
+buffer_size = 50_000_000
+## The size of the queue of blocks which are waiting for a parent block to be downloaded (bytes).
+in_progress_queue_size = 50_000_000
+## The target size of a batch of blocks (bytes), must not exceed 100MB.
+target_batch_size = 5_000_000
+## The number of blocks in the first bacth (you probably shouldn't change this).
```
hinto-janai, line 51:
```suggestion
## The number of blocks in the first batch (you probably shouldn't change this).
```

binaries/cuprated/src/config/p2p.rs
```diff
@@ -0,0 +1,110 @@
...
+            "192.99.8.110:28080",
+            "37.187.74.171:28080",
+            "77.172.183.193:28080",
+        ]
+        .as_slice(),
+    };
+
+    seeds
+        .iter()
+        .map(|&s| str::parse(s))
```
hinto-janai, line 107:
```suggestion
        .map(|s| s.parse())
```

hinto-janai (commented):
nits only, will leave other review

helper/src/network.rs
```diff
@@ -5,6 +5,11 @@
 //! into it's own crate.
 //!
 //! `#[no_std]` compatible.
+// TODO: move to types crate.
```
hinto-janai, line 8: Can you move this to `cuprate_types`? `strum` can automatically implement the `FromStr/Display` and lots of other stuff.

Also please add a doc-test asserting the string representations since we rely on it for PATHs now, something like this: https://github.com/Cuprate/cuprate/blob/63216aecaead915550be93236faf6aeeebf0e04f/rpc/types/src/misc/status.rs#L31-L65

binaries/cuprated/src/config/storage.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// The tx-pool config.
+#[derive(Deserialize, Serialize)]
+#[serde(deny_unknown_fields, default)]
+pub struct TxpoolConfig {
+    #[serde(flatten)]
+    pub shared: SharedStorageConfig,
+
+    /// The maximum size of the tx-pool (bytes).
+    pub max_txpool_size: usize,
```
hinto-janai, line 46:
```suggestion
    /// The maximum size of the tx-pool.
    pub max_txpool_byte_size: usize,
```

storage/blockchain/src/config.rs
hinto-janai: I don't think this needs to be changed from `Cow<'static, Path>` -> `PathBuf`.

helper/src/fs.rs
```diff
@@ -178,6 +183,11 @@ impl_path_lazylock! {
     "txpool",
 }
 
+/// Joins the [`Path`] with a folder for the given [`Network`].
+pub fn path_with_network(path: &Path, network: Network) -> PathBuf {
+    path.join(network.to_string())
+}
```
hinto-janai, line 189: The tree should be:
```bash
/home/alice/.cache/cuprate/{mainnet,testnet,stagenet}/*
```
This function and the way it's used will create:
```bash
/home/alice/.cache/cuprate/*/{mainnet,testnet,stagenet}/*
```

Maybe integrate `{mainnet,testnet,stagenet}` into the macro generating the PATHs so we have: `CUPRATE_{MAINNET,TESTNET,STAGENET}_{DATA,BLOCKCHAIN,CONFIG,...}_DIR`?

That's kinda hacky too but this function is a lot worse since it seeps into user space. `monerod` dumps `mainnet` into `.bitmonero/` and creates `.bitmonero/{testnet,stagenet}`. Maybe see how `bitcoin/eth/zcash` do it?

storage/txpool/src/config.rs
```diff
@@ -83,16 +90,23 @@ impl ConfigBuilder {
         }
     }
 
+    /// Change the network this blockchain database is for.
```
hinto-janai, line 93: Bad copy+paste
```suggestion
    /// Change the network this database is for.
```

hinto-janai (changes requested):

binaries/cuprated/src/config.rs
```diff
@@ -1 +1,145 @@
...
+mod p2p;
+mod storage;
+mod tracing_config;
+
+use p2p::P2PConfig;
+use storage::StorageConfig;
+use tracing_config::TracingConfig;
+
+/// The default name of Cuprate's config file.
+const DEFAULT_CONFIG_FILE_NAME: &str = "Cuprate.toml";
```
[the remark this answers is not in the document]
Boog900: > cuprated.toml, we decided on this already no? Cuprate will clash with any future binary

Oops yeah I forgot

binaries/cuprated/src/config.rs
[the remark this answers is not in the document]
Boog900: This is habit from fulfilling the lint requiring comments, I don't think it's enabled here though, I am still going to leave them here though for if that ever happens. FWIW you do this too.

binaries/cuprated/src/config/Cuprate.toml
```diff
@@ -0,0 +1,75 @@
+#     ____                      _
```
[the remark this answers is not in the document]
Boog900: No I didn't explain myself well enough, this exact file isn't valid `toml`, the actual config files we produce/consume are.


Answering https://github.com/Cuprate/cuprate/pull/304#discussion_r1819958586 as well: 
> What are these {{}} values? Shouldn't it be:

This file is passed into the `format` macro so we can swap the paths it produces to the correct ones for different systems : https://github.com/Cuprate/cuprate/blob/06b7429f9e2cf71b404b27bb403d7edc3180df5d/binaries/cuprated/src/config/default.rs#L47-L52.

The output this produces is valid `toml`, the tests check this.


> If it isn't valid TOML, labeling it .toml is pretty bad practice

I agree, however this is an internal file and I prefer having some syntax highlighting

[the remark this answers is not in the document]
Boog900: example output: https://github.com/Cuprate/cuprate/pull/304#pullrequestreview-2397349245

binaries/cuprated/src/config/Cuprate.toml
```diff
@@ -0,0 +1,75 @@
...
+#    / ___|   _ _ __  _ __ __ _| |_ ___
+#   | |  | | | | '_ \| '__/ _` | __/ _ \
+#   | |__| |_| | |_) | | | (_| | ||  __/
+#    \____\__,_| .__/|_|  \__,_|\__\___|
+#              |_|
+#
+
+## The network to run on, valid values: "Mainnet", "Testnet", "Stagenet".
```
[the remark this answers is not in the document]
Boog900: > An alternative would be to strip the config of comments and put all the info within the User Book.

I think a middle ground would be good, some small comments in the config and more in-depth explanations in the book

binaries/cuprated/src/config.rs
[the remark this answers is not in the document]
hinto-janai: > FWIW you do this too

Fair enough, I have done this in the past (in personal/company projects and this codebase) for the exact same reason and I regret it, I try not to do it anymore.

Lots of the criticism I'll give code is because I see a pattern that I've personally done in the past but have since then realized it was a mistake.

> This is habit from fulfilling the lint requiring comments, I don't think it's enabled here though, I am still going to leave them here though for if that ever happens.

I'd still prefer them undocumented since _if_ we enable the lint (I'm less confident we should do this now for the reasons above) then it will show us the fields that are not documented. Duplicating the information is close to doing `#[allow(missing_docs)]` but worse since it adds noise to docs.

Can we compromise and agree that our codebase should not have docs/comments like this in the future?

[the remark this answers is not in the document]
SyntheticBird45: In case you saw it don't mind my prior message I just deleted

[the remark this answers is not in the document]
SyntheticBird45: > Duplicating the information is close to doing #[allow(missing_docs)] but worse since it adds noise to docs.

I think this is something that isn't substantial at all. Most developers will not pay attention to the duplication of info nor does it cost any issue in development.

>  Can we compromise and agree that our codebase should not have docs/comments like this in the future?

My stance is we should stay lenient on that. I don't think this type of review point is even appropriate for other contributors. It rely more of the semantic perfection than actual correctness improvement IMO.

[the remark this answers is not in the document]
hinto-janai: I wrote a long response but it's not worth it. This nit can be ignored/resolved.

helper/src/fs.rs
Boog900: So initially I wanted to do that, however it gets annoying when you allow users to specify custom paths for different things.

I think it might be best to remove the blockchain & txpool paths from the macro and instead have functions that build them on the fly from a `Network` and a `data_dir`, i.e.:

```rust
pub fn blockchain_path(data_dir: &Path, network: Network) -> PathBuf {
    path.join(network.to_string()).join("blockchain")
}
```

(I'll  change it so we match on the network and don't write `mainnet`).

Doing it like that makes it easier IMO rather than needing to match on the network to choose the right constant and potentially using a completely custom dir anyway.

helper/src/network.rs
Boog900: Will do in another PR this is used in a lot of places so don't really want to do it here.

binaries/cuprated/src/config.rs
```diff
@@ -1 +1,145 @@
...
+
+/// Reads the args & config file, returning a [`Config`].
+pub fn read_config_and_args() -> Config {
```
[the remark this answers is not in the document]
Boog900: In my mind leaving them free gives them more weight, this function is not "just" under the "Config" class it is also handling the args for example.

This isn't a function that should be called a lot like `Vec::new` for example, it should only be called once. I associate it more with a general start-up function than the wider use of the `Config` structure.

If you would rather me make this an associated function I can.

storage/blockchain/src/config.rs
Boog900: It doesn't _need_ to but `Cow<'static, Path>` is more annoying to work with for little benefit, it's literally 0 benefit now as we need runtime paths.

binaries/cuprated/src/config.rs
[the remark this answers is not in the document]
hinto-janai: OK nvm, after looking at this again I think the free function is good too.

helper/src/fs.rs
```diff
@@ -150,32 +158,61 @@ impl_path_lazylock! {
...
+/// This is the PATH used for any Cuprate address-book files.
+///
+/// ```rust
+/// use cuprate_helper::{network::Network, fs::{CUPRATE_CACHE_DIR, addressbook_path}};
+///
+/// assert_eq!(addressbook_path(&**CUPRATE_CACHE_DIR, Network::Mainnet).as_path(), CUPRATE_CACHE_DIR.join("addressbook"));
+/// assert_eq!(addressbook_path(&**CUPRATE_CACHE_DIR, Network::Stagenet).as_path(), CUPRATE_CACHE_DIR.join(Network::Stagenet.to_string()).join("addressbook"));
+/// assert_eq!(addressbook_path(&**CUPRATE_CACHE_DIR, Network::Testnet).as_path(), CUPRATE_CACHE_DIR.join(Network::Testnet.to_string()).join("addressbook"));
+/// ```
+pub fn addressbook_path(cache_dir: &Path, network: Network) -> PathBuf {
```
hinto-janai, line 214: nit: this is `AddressBook/address_book` in all other code.
```suggestion
pub fn address_book_path(cache_dir: &Path, network: Network) -> PathBuf {
```

hinto-janai (commented):
1 thing + 2 nits, there's other things too but this needs to be merged and I don't want to hold it up any longer.

binaries/cuprated/src/config/args.rs
```diff
@@ -24,21 +24,26 @@ pub struct Args {
...
-    pub fn apply_args(&self, mut config: Config) -> Config {
-        if let Some(config_folder) = self.generate_config.as_ref() {
-            // This will create the config file and exit.
-            create_default_config_file(config_folder)
+    /// May cause the process to [`exit`].
+    pub fn do_quick_requests(&self) {
+        if self.generate_config {
+            println!("{EXAMPLE_CONFIG}");
+            exit(0);
         };
```
hinto-janai, line 40:
```suggestion
        }
```

hinto-janai (approved):

PR #306, opened 2024-10-07
P2P: move seed nodes to config

Boog900:
### What
Moves the seed nodes to the config + removes a few unused constants 

### Why
Seeds nodes depend on the network.

### DIFF
--- a/p2p/p2p-core/src/network_zones/clear.rs
+++ b/p2p/p2p-core/src/network_zones/clear.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::{IpAddr, Ipv4Addr, SocketAddr},
+    net::{IpAddr, SocketAddr},
     pin::Pin,
     task::{Context, Poll},
 };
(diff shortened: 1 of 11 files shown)


p2p/p2p/src/connection_maintainer.rs
```diff
@@ -104,12 +104,12 @@ where
...
+        let seeds = self
+            .config
+            .seeds
+            .choose_multiple(&mut thread_rng(), MAX_SEED_CONNECTIONS);
 
-        assert!(
-            seeds.len() != 0,
-            "No seed nodes available to get peers from"
-        );
+        assert_ne!(seeds.len(), 0, "No seed nodes available to get peers from");
```
hinto-janai, line 112: nit: https://rust-lang.github.io/rust-clippy/master/index.html#/len_zero

```suggestion
        assert!(!seeds.is_empty(), "No seed nodes available to get peers from");
```

p2p/p2p-core/src/network_zones/clear.rs
```diff
@@ -45,26 +45,10 @@ pub struct ClearNetServerCfg {
...
 
-    const SEEDS: &'static [Self::Addr] = &[
-        ip_v4(176, 9, 0, 187, 18080),
-        ip_v4(88, 198, 163, 90, 18080),
-        ip_v4(66, 85, 74, 134, 18080),
-        ip_v4(51, 79, 173, 165, 18080),
-        ip_v4(192, 99, 8, 110, 18080),
-        ip_v4(37, 187, 74, 171, 18080),
-        ip_v4(77, 172, 183, 193, 18080),
-    ];
```
hinto-janai, line 64: Doesn't removing this mean users will have to manually provide seed nodes?

Will there eventually be default seed nodes added to the `P2PConfig`?

hinto-janai (approved):

p2p/p2p-core/src/network_zones/clear.rs
Boog900: > Doesn't removing this mean users will have to manually provide seed nodes?

Users as in devs using `cuprate-p2p`? yes.

Users as in people using `cuprated`? no.

> Will there eventually be default seed nodes added to the P2PConfig?

in #304 I am creating the config that will be exposed to `cuprated` users in there there will be default seed nodes

p2p/p2p/src/connection_maintainer.rs
Boog900: `seeds` is a type which doesn't have a `is_empty` stable: https://doc.rust-lang.org/nightly/core/iter/traits/exact_size/trait.ExactSizeIterator.html#method.is_empty

PR #308, opened 2024-10-10
cuprated: RPC handlers (json-rpc)

hinto-janai:
### What
Implements the JSON-RPC handlers in `cuprated`; adds various types and changes some misc things as needed.

PR #309, opened 2024-10-11
types: `HardFork` improvements

hinto-janai:
### What
Adds docs, tests, `fn` to the `cuprate_types::HardFork` type.

### Why
Added changes used in #308.

### DIFF
--- a/types/src/hard_fork.rs
+++ b/types/src/hard_fork.rs
@@ -73,13 +73,13 @@ pub enum HardFork {
 }
 
 impl HardFork {
-    /// The current [`HardFork`].
+    /// The latest [`HardFork`].
     ///
     /// ```rust
     /// # use cuprate_types::HardFork;
-    /// assert_eq!(HardFork::CURRENT, HardFork::V16);
+    /// assert_eq!(HardFork::LATEST, HardFork::V16);
     /// ```
-    pub const CURRENT: Self = Self::VARIANTS[Self::COUNT - 1];
+    pub const LATEST: Self = Self::VARIANTS[Self::COUNT - 1];
 
     /// Returns the hard-fork for a blocks [`BlockHeader::hardfork_version`] field.
     ///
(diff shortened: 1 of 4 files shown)


types/src/hard_fork.rs
```diff
@@ -1,6 +1,10 @@
 //! The [`HardFork`] type.
 use std::time::Duration;
 
+use strum::{
+    AsRefStr, Display, EnumCount, EnumIs, EnumString, FromRepr, IntoStaticStr, VariantArray,
+};
```
hinto-janai, line 6: Hopefully pulling in `strum` is justified by replacing the manual "remember to update" code with an automated version.

types/src/hard_fork.rs
```diff
@@ -47,58 +69,87 @@ pub enum HardFork {
...
+            assert!(COUNT <= u8::MAX as usize);
+            COUNT as u8
+        };
+
+        if version == 0 || version > COUNT_AS_U8 {
+            Err(HardForkError::HardForkUnknown)
+        } else {
+            // INVARIANT: we've proved above that `version` is in a valid range.
+            Ok(Self::VARIANTS[(version - 1) as usize])
+        }
```
hinto-janai, line 119: This is also an automatic version of the old `from_version`, this now means updating `HardFork` is as simple as adding another variant.

It's a bit more complex than the `match` though...

types/src/hard_fork.rs
```diff
@@ -109,22 +160,49 @@ impl HardFork {
...
+    ///     assert_eq!(hf.as_u8(), i as u8 + 1);
+    /// }
+    /// ```
+    pub const fn as_u8(self) -> u8 {
+        self as u8
     }
 
     /// Returns the next hard-fork.
-    pub fn next_fork(&self) -> Option<Self> {
-        Self::from_version(*self as u8 + 1).ok()
```
hinto-janai, line 118: Copy is cheaper than `&`.

types/src/hard_fork.rs
Boog900: Couldn't we use [FromRepr](https://docs.rs/strum/0.26.3/strum/derive.FromRepr.html#) here?

types/src/hard_fork.rs
```diff
@@ -47,58 +69,87 @@ pub enum HardFork {
...
 }
 
 impl HardFork {
+    /// The current [`HardFork`].
+    ///
+    /// ```rust
+    /// # use cuprate_types::HardFork;
+    /// assert_eq!(HardFork::CURRENT, HardFork::V16);
+    /// ```
+    pub const CURRENT: Self = Self::VARIANTS[Self::COUNT - 1];
```
Boog900, line 82: Bit of nit but I think this should be called  `LATEST`  as the last HF may have not been activated, therefore it's not always the current HF.

types/src/hard_fork.rs
```diff
@@ -109,22 +160,49 @@ impl HardFork {
...
+    ///
+    /// for hf in HardFork::VARIANTS.iter() {
+    ///     if *hf == HardFork::CURRENT {
+    ///         assert!(hf.is_current());
+    ///     } else {
+    ///         assert!(!hf.is_current());
+    ///     }
+    /// }
+    /// ```
+    pub const fn is_current(self) -> bool {
```
Boog900, line 205: same here: `is_latest`?

Boog900 (approved):

PR #320, opened 2024-10-18
cuprated: internal signatures required for RPC pt. 2

hinto-janai:
### What
Adds more `{function, request, response}` signatures needed by `cuprated`'s RPC system to the `{blockchain, txpool, p2p, blockchain_context, rpc}` APIs.

Continuation of https://github.com/Cuprate/cuprate/pull/297.

This _should_ be all that's needed for JSON-RPC requests, but it may not cover `{binary, other}`.


binaries/cuprated/src/rpc/handler.rs
```diff
@@ -54,6 +56,61 @@ pub enum BlockchainManagerRequest {
...
+    /// Get a visual [`String`] overview of blockchain progress.
+    ///
+    /// This is a highly implementation specific format used by
+    /// `monerod` in the `sync_info` RPC call's `overview` field;
+    /// it is essentially an ASCII visual of blocks.
+    ///
+    /// See also:
+    /// - <https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#sync_info>
+    /// - <https://github.com/monero-project/monero/blob/master/src/cryptonote_protocol/block_queue.cpp#L178>
+    Overview {
```
hinto-janai, line 110: ### TODO
Is it worth re-implementing and supporting these highly specific one-off `monerod` string formats? There's more than this one too.

binaries/cuprated/src/rpc/request/address_book.rs
```diff
@@ -2,44 +2,112 @@
...
 use tower::ServiceExt;
 
 use cuprate_helper::cast::usize_to_u64;
 use cuprate_p2p_core::{
     services::{AddressBookRequest, AddressBookResponse},
+    types::BanState,
     AddressBook, NetworkZone,
 };
 
+// FIXME: use `anyhow::Error` over `tower::BoxError` in address book.
```
hinto-janai, line 17: ### TODO
I tried replacing `tower::BoxError` with `anyhow::Error`, it's pretty deeply connected within `cuprate_consensus` but I think it works, the only problem was that one of the functions used (I forget which, either a `tower` or `tokio` one) for starting a service (I think) depends directly on the `tower::BoxError` type.

For now this uses `.map_err(|e| anyhow!(e))` which means we're boxing the error twice.

p2p/p2p-core/src/types.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub pruning_seed: PruningSeed,
+    pub recv_count: u64,
+    pub recv_idle_time: u64,
+    pub rpc_credits_per_hash: u32,
+    pub rpc_port: u16,
+    pub send_count: u64,
+    pub send_idle_time: u64,
+    pub state: String, // TODO: what type is this?
+    pub support_flags: u32,
+}
```
hinto-janai, line 113: ### TODO
- For `connection_id`, our currently used `u64` should be fine in place of `monerod`'s `boost::uuids::uuid`, right? The RPC output I think just needs a hex version of the ID.
- What can `{ip,port}` be replaced by `std::net::SocketAddr`? What about i2p/tor?
- What type is `{peer_id,state}`?

Context:
- https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/cryptonote_protocol/cryptonote_protocol_defs.h#L47
- https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#get_connections

hinto-janai (commented):
Same as #297, this review is split like such:

| Title       | Purpose |
|-------------|---------|
| Unrelated   | Unrelated changes that can be ignored   
| Signatures  | New requests/responses added to our `tower::Service`s 
| Types       | New types added (for usage in the request/responses)
| TODO        | Things to do after this PR

Same questions that need answering:
- Are any of the requests/responses impossible to fulfill?
- Can any of the requests/responses be done in a better way?
- Do any of the requests/responses belong to a different `Service`?

Boog900 (changes requested):

p2p/p2p-core/src/types.rs
SyntheticBird45: @hinto-janai `boost::uuids::uuid` is 128 bit not 64 bit.
https://www.boost.org/doc/libs/1_86_0/libs/uuid/doc/html/uuid.html

binaries/cuprated/src/rpc/handler.rs
SyntheticBird45: @Boog900 I agree let's do put a surprise there

p2p/p2p-core/src/types.rs
Boog900: > For connection_id, our currently used u64 should be fine in place of monerod's boost::uuids::uuid, right? The RPC output I think just needs a hex version of the ID.

We just the peers address to identify the, we don't have a connection ID. If the address is unknown like an incoming Tor connection we use a u128.

> What can {ip,port} be replaced by std::net::SocketAddr? What about i2p/tor?

Yeah you should be able to pull them from the `std::net::SocketAddr`, if that is what you are asking. I don't know how monerod handles this for Tor.

> What type is {peer_id,state}?

peer_id is a u64 and state is a monerod specific string

rpc/types/src/misc/address_type.rs
```diff
@@ -0,0 +1,107 @@
...
+pub enum AddressType {
+    #[default]
+    Invalid,
+    Ipv4,
+    Ipv6,
+    I2p,
+    Tor,
+}
+
+impl AddressType {
```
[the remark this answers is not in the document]
hinto-janai: Yes, one for `cuprate-p2p` to use and one for `cuprate-rpc-types` to use. The RPC version is the same but it has `serde/epee`. I also added another similar enum for [`state`](https://github.com/Cuprate/cuprate/pull/320#discussion_r1811335020) (`ConnectionState`).

I moved them to `cuprate-types` so they can impl `serde/epee`, although, this does mean `cuprate-p2p-core` now pulls in `cuprate-types`.

p2p/p2p-core/src/types.rs
```diff
@@ -24,100 +25,48 @@ pub struct BanState<A: NetZoneAddress> {
...
+    pub state: ConnectionState,
     pub support_flags: u32,
+
+    // The following fields are slightly different than `monerod`.
+    //
+    /// [`None`] if Tor/i2p or unknown.
+    pub socket_addr: Option<std::net::SocketAddr>,
+    // This field does not exist for Cuprate:
+    // <https://github.com/Cuprate/cuprate/pull/320#discussion_r1811335020>
+    // pub connection_id: u128,
```
hinto-janai, line 61: I guess we'll just use `FIELD_NOT_SUPPORTED` for fields like this?

Thankfully the RPC type is a hex `String` - I guess for types like `u64` we just use `Default::default()`?

consensus/context/src/lib.rs
```diff
@@ -263,6 +264,21 @@ pub enum BlockChainContextRequest {
...
+    CalculatePow {
+        /// The hardfork of the protocol at this block height.
+        hardfork: HardFork,
+        /// The height of the block.
+        height: usize,
+        /// The block data.
+        ///
+        /// This is boxed because [`Block`] causes this enum to be 1200 bytes,
+        /// where the 2nd variant is only 96 bytes.
+        block: Box<Block>,
```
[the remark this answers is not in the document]
Boog900: yeah I am ok with boxing this

p2p/p2p-core/src/types.rs
Boog900: I think it might be best to try mimic the format but just set it to a default value, like `connection_id` is a UUID right? we should just set it to a default UUID and hex encode that.

storage/txpool/src/tx.rs
```diff
@@ -12,3 +12,17 @@ pub struct TxEntry {
...
+/// Used in [`TxpoolReadResponse::BlockTemplateBacklog`](crate::service::interface::TxpoolReadResponse::BlockTemplateBacklog).
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct BlockTemplateTxEntry {
+    /// The transaction's ID (hash).
+    pub id: [u8; 32],
+    /// The transaction's weight.
+    pub weight: u64,
+    /// The transaction's fee.
+    pub fee: u64,
+}
```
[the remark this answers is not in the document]
Boog900: Yeah I think we should combine this request on the database side.

types/src/blockchain.rs
```diff
@@ -128,6 +129,12 @@ pub enum BlockchainReadRequest {
     ///
     /// TODO: document fields after impl.
     CoinbaseTxSum { height: usize, count: u64 },
+
+    /// Get information on all alternative chains.
+    AltChains,
+
+    /// Get the amount of alternative chains that exist.
+    AltChainCount,
```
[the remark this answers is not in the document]
Boog900: I don't think we should - this request just gives you an overview `AltBlocksInChain` returns the full block blobs

p2p/p2p-core/src/types.rs
SyntheticBird45: I agree with @Boog900 . If I had to debug breakage in migration i would prefer it to be still a default id of some sort rather an another type altogether. Just to avoid a big gasp here.

binaries/cuprated/src/rpc/request/address_book.rs
```diff
@@ -2,44 +2,108 @@
...
 use std::convert::Infallible;
 
-use anyhow::Error;
+use anyhow::{anyhow, Error};
+use cuprate_pruning::PruningSeed;
+use cuprate_rpc_types::misc::{ConnectionInfo, Span};
 use tower::ServiceExt;
 
 use cuprate_helper::cast::usize_to_u64;
 use cuprate_p2p_core::{
```
Boog900, line 11:
```suggestion
use tower::ServiceExt;

use cuprate_pruning::PruningSeed;
use cuprate_rpc_types::misc::{ConnectionInfo, Span};
use cuprate_helper::cast::usize_to_u64;
use cuprate_p2p_core::{
```

binaries/cuprated/src/rpc/request/blockchain_context.rs
```diff
@@ -2,7 +2,9 @@
 
 use std::convert::Infallible;
 
-use anyhow::Error;
+use anyhow::{anyhow, Error};
+use cuprate_helper::cast::u64_to_usize;
+use monero_serai::block::Block;
 use tower::{Service, ServiceExt};
 
 use cuprate_consensus_context::{
```
Boog900, line 10:
```suggestion
use monero_serai::block::Block;
use tower::{Service, ServiceExt};

use cuprate_helper::cast::u64_to_usize;
use cuprate_consensus_context::{
```

binaries/cuprated/src/rpc/request/blockchain_manager.rs
```diff
@@ -1,17 +1,22 @@
...
 
 use anyhow::Error;
+use cuprate_p2p_core::NetworkZone;
+use cuprate_rpc_types::misc::Span;
 use monero_serai::block::Block;
 use tower::{Service, ServiceExt};
 
 use cuprate_helper::cast::{u64_to_usize, usize_to_u64};
+use cuprate_pruning::PruningSeed;
+use cuprate_types::{AddAuxPow, AuxPow, HardFork};
```
Boog900, line 11:
```suggestion
use monero_serai::block::Block;
use tower::{Service, ServiceExt};

use cuprate_p2p_core::NetworkZone;
use cuprate_rpc_types::misc::Span;
use cuprate_helper::cast::{u64_to_usize, usize_to_u64};
use cuprate_pruning::PruningSeed;
use cuprate_types::{AddAuxPow, AuxPow, HardFork};
```

Boog900 (changes requested):
These are mostly nits, I also left some comments separately

storage/txpool/src/tx.rs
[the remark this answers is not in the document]
hinto-janai: > combine this request on the database side

What does this mean specifically? 1 type/request/response?:

```rust
struct TxEntry {
    pub id: [u8; 32],
    pub weight: u64,
    pub fee: u64,
    pub time_in_pool: Duration,
}
```

binaries/cuprated/src/rpc/constants.rs
```diff
@@ -0,0 +1,5 @@
+//! Constants used within RPC.
+
+/// The string message used in RPC response fields for when
+/// `cuprated` does not support a field that `monerod` has.
+pub(super) const FIELD_NOT_SUPPORTED: &str = "`cuprated` does not support this field.";
```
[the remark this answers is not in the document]
hinto-janai: @Boog900 @SyntheticBird45 btw I meant this will just be the default value for now, easter egg can replace it whenever.

storage/txpool/src/tx.rs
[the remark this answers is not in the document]
Boog900: Yeah

p2p/p2p-core/src/types.rs
Boog900: Ah ok, I thought this would be returned in RPC requests, if this is just internal and we do replace it with data that decodes into the correct type before sending it over the wire then I am ok with that.

binaries/cuprated/src/rpc/request/blockchain_manager.rs
```diff
@@ -139,3 +146,76 @@ pub(super) async fn target_height(
...
+    //     unreachable!();
+    // };
+
+    let vec: Vec<cuprate_p2p_core::types::Span<Z::Addr>> = todo!();
+
+    // FIXME: impl this map somewhere instead of inline.
+    let vec = vec
+        .into_iter()
+        .map(|span| Span {
+            connection_id: String::from(FIELD_NOT_SUPPORTED),
```
Boog900, line 194: This is putting the wrong string type in the response right? Should it be a `ConnectionId::DEFAULT_STR` ?

Boog900 (approved):

PR #323, opened 2024-10-21
storage: Add common amounts commitment lookup table

### LINKED ISSUE #245: Lookup table for calculating amount commitments
## What

When retrieving pre-RCT outputs from the DB we need to calculate an amount commitment for use in RCT rings, this is currently done manually:

https://github.com/Cuprate/cuprate/blob/27767690ca7d7c5b44172a857581f08c728a740d/storage/blockchain/src/ops/output.rs#L158-L160

Also when adding V2 miner txs:

https://github.com/Cuprate/cuprate/blob/27767690ca7d7c5b44172a857581f08c728a740d/storage/blockchain/src/ops/tx.rs#L123-L129

This issue is for creating a lookup table for common amounts (the validly decomposed amounts).

## Why

Calculating commitments is slower than using a look up table.

## How

I did this in our test sync branch:

https://github.com/Cuprate/cuprate/blob/1eaf32bee956b66591f1537201c16969d5ec8696/helper/src/commitment.rs#L55-L62

However performance tests should be conducted to see if the slow path (line 60) is actually faster than our current method

SyntheticBird45:
### What

Fixes #245

Implements monerod's `zero_commitment` lookup table for common amounts defined here: https://github.com/monero-project/monero/blob/c8214782fb2a769c57382a999eaf099691c836e7/src/ringct/rctOps.cpp#L43

Unlike monerod which uses a simple linear search, this PR implements a binary search over the lookup table, saving a few dozen nanoseconds. 

More micro-optimization could be done like verifying if the amount is over 10 then modulo of amount per 10 should be 0, saving ourselves the lookup cost by directly passing to calculation.


constants/src/crypto.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// This is a lookup table containing the zero commiment of known Pre-RCT value
+/// and is used to avoid computation of the commitment.
+/// 
+/// Defined at:
+/// - <https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/ringct/rctOps.cpp#L44>
+#[rustfmt::skip]
+pub static ZERO_COMMITMENT_LOOKUP_TABLE: [(u64, [u8;32]); 173] = 
+[
+  (0_u64,                       [0x58, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66, 0x66] ),
```
Boog900, line 11: Can we calcualte these at runtime using a LazyLock, like I did here: https://github.com/Cuprate/cuprate/blob/1eaf32bee956b66591f1537201c16969d5ec8696/helper/src/commitment.rs#L55-L62

storage/blockchain/src/ops/output.rs
```diff
@@ -155,9 +158,31 @@ pub fn output_to_output_on_chain(
...
+                    end = mid - 1;
+                }
+                _ => {
+                    start = mid + 1;
+                }
+            }
+        }
+
+        // Compute the zero commitment if not found.
+        c.unwrap_or_else(|| ED25519_BASEPOINT_POINT + *H * Scalar::from(amount))
```
Boog900, line 184: Can we have this function with the commitment table please

storage/blockchain/src/ops/output.rs
```diff
@@ -155,9 +158,31 @@ pub fn output_to_output_on_chain(
...
+        let (mut start, mut end, mut c): (usize, usize, Option<EdwardsPoint>) =
+            (0, (ZC_LT.len() - 1) / 2, None);
+        while start <= end {
+            let mid = (start + end) / 2;
+            let lookup = ZC_LT[mid].0;
+            match amount {
+                a if a == lookup => {
+                    c = CompressedEdwardsY::from_slice(&ZC_LT[mid].1)
+                        .expect("The lookup table cannot contain invalid curve point data")
+                        .decompress();
```
Boog900, line 172: Doing a scalar decomposition here is pretty expensive - the cache could be the decompressed EdwardsPoints

Boog900 (changes requested):

storage/blockchain/src/ops/output.rs
Boog900, line 172: We could keep the decomposed point in the cache instead of doing a decompression on every use.

SyntheticBird45:
Applied requested changes however I wanted to benchmark hashmap for the lookup. The benchmark is jut a lookup over all the decomposed amounts. 
- I kept manual binary search
- Implemented the opportunistic optimization with the modulo. If the value is over 10 and the modulo is different than 0 then it skip the lookup as we now the value isn't in the table.
- Since `cuprate-constants` do not have `std` support I moved back the code to `cuprate-helper` similarly to your test branch.

Benchmark:

benchmark_hashmap

helper/src/crypto.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    // Binary search in lookup table (O(log2(n)) instead of O(n))
+    let (mut start, mut end, mut c): (usize, usize, Option<EdwardsPoint>) =
+        (0, (ZERO_COMMITMENT_LOOKUP_TABLE.len() - 1) / 2, None);
+    while start <= end {
+        let mid = (start + end) / 2;
+        let lookup = ZERO_COMMITMENT_LOOKUP_TABLE[mid].0;
+        match amount {
+            a if a == lookup => {
+                c = Some(ZERO_COMMITMENT_LOOKUP_TABLE[mid].1);
```
Boog900, line 81: I think you are meant to return here, otherwise this will infinite loop.

As a fun exercise I made a "constant time" algorithm:
```rust
    let log = amount.checked_ilog10().unwrap(); // don't actually unwrap here.
    let div = 10_u64.pow(log);

    let most_significant_digit = amount / div;

    if most_significant_digit * div != amount {
        return H_PRECOMP.vartime_multiscalar_mul([Scalar::from(amount), Scalar::ONE]);
    }

    let index = (most_significant_digit + u64::from(log) * 10 - 1 * (u64::from(log) + 1)) as usize;
    
    ZERO_COMMITMENT_LOOKUP_TABLE[index].1
```

note: this is not actually constant time for cryptographic purposes. it would be good to see how that stacks up.


FWIW: I would rather use rusts build in binary search even if it is a little slower.

helper/src/crypto.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Compute it otherwise.
+pub fn compute_zero_commitments(amount: u64) -> EdwardsPoint {
+    // OPTIMIZATION: We first make an arithmetic check that the value lies between valid pre-computed amounts. This permit to avoid the lookup cost if possible
+    if amount > 10 && amount % 10 != 0 {
+        return H_PRECOMP.vartime_multiscalar_mul([Scalar::from(amount), Scalar::from(1_u8)]);
+    }
+
+    // Binary search in lookup table (O(log2(n)) instead of O(n))
+    let (mut start, mut end, mut c): (usize, usize, Option<EdwardsPoint>) =
+        (0, (ZERO_COMMITMENT_LOOKUP_TABLE.len() - 1) / 2, None);
```
Boog900, line 75: Also here we shouldn't be using the mid point as the end.

SyntheticBird45: Thx tests are passing now

helper/src/crypto.rs
SyntheticBird45: I must say you nailed it. Here's the benchmark:

constant_time

```
lookup_search/binary_search_manual
                        time:   [835.15 ns 839.11 ns 843.73 ns]
lookup_search/constant_time
                        time:   [871.62 ns 873.04 ns 874.56 ns]
```

The manual implementation of binary search is faster than constant time yet yours is clever (imo).

SyntheticBird45: Forget to push the end pointer fix on the manual binary search one:

fixed

```
lookup_search/binary_search_std
                        time:   [1.2080 µs 1.2101 µs 1.2124 µs]
lookup_search/binary_search_manual
                        time:   [991.33 ns 993.50 ns 995.41 ns]
lookup_search/hashmap_search_std
                        time:   [1.3606 µs 1.3620 µs 1.3636 µs]
lookup_search/constant_time
                        time:   [864.30 ns 865.61 ns 867.13 ns]
```

[Benchmark code.](https://gist.github.com/SyntheticBird45/e97c045aca71c5a0380dba8cf52bd0c5)

I obviously advocate we use your constant time version with a generous documentation on top.

helper/src/crypto.rs
```diff
@@ -0,0 +1,93 @@
...
+            a if a == lookup => {
+                return ZERO_COMMITMENT_LOOKUP_TABLE[mid].1;
+            }
+            a if a < lookup => {
+                end = mid - 1;
+            }
+            _ => {
+                start = mid + 1;
+            }
+        }
```
hinto-janai, line 88: I think `clippy::comparison_chain` wants this:
```suggestion
        match amount.cmp(&lookup) {
            Ordering::Equal => {
                return ZERO_COMMITMENT_LOOKUP_TABLE[mid].1;
            }
            Ordering::Less => {
                end = mid - 1;
            }
            Ordering::Greater => {
                start = mid + 1;
            }
        }
```

helper/src/crypto.rs
```diff
@@ -0,0 +1,93 @@
...
+use std::sync::LazyLock;
+
+use curve25519_dalek::{
+    constants::ED25519_BASEPOINT_POINT, edwards::VartimeEdwardsPrecomputation,
+    traits::VartimePrecomputedMultiscalarMul, EdwardsPoint, Scalar,
+};
+use monero_serai::generators::H;
+
+/// This is the decomposed amount table containing the mandatory Pre-RCT amounts. It is use to pre-compute 
+/// zero commitments at runtime.
```
hinto-janai, line 13:
```suggestion
/// This is the decomposed amount table containing the mandatory Pre-RCT amounts. It is used to pre-compute 
/// zero commitments at runtime.
```

helper/src/crypto.rs
```diff
@@ -0,0 +1,93 @@
...
+    100000000000000,     200000000000000,     300000000000000,     400000000000000,     500000000000000,     600000000000000,     700000000000000,     800000000000000,     900000000000000,
+    1000000000000000,    2000000000000000,    3000000000000000,    4000000000000000,    5000000000000000,    6000000000000000,    7000000000000000,    8000000000000000,    9000000000000000,
+    10000000000000000,   20000000000000000,   30000000000000000,   40000000000000000,   50000000000000000,   60000000000000000,   70000000000000000,   80000000000000000,   90000000000000000,
+    100000000000000000,  200000000000000000,  300000000000000000,  400000000000000000,  500000000000000000,  600000000000000000,  700000000000000000,  800000000000000000,  900000000000000000,
+    1000000000000000000, 2000000000000000000, 3000000000000000000, 4000000000000000000, 5000000000000000000, 6000000000000000000, 7000000000000000000, 8000000000000000000, 9000000000000000000,
+    10000000000000000000
+];
+
+/// Runtime initialized H generator.
```
hinto-janai, line 41:
```suggestion
/// Runtime initialized [`H`] generator.
```

helper/src/crypto.rs
```diff
@@ -0,0 +1,93 @@
...
+static H_PRECOMP: LazyLock<VartimeEdwardsPrecomputation> =
+    LazyLock::new(|| VartimeEdwardsPrecomputation::new([*H, ED25519_BASEPOINT_POINT]));
+
+/// Runtime initialized zero commitment lookup table
+///
+/// ASSUMPTION: This function assume that the [`ZERO_COMMITMENT_DECOMPOSED_AMOUNT`]
+/// table is sorted.
```
hinto-janai, line 48:
```suggestion
/// # Invariant
/// This function assumes that the [`ZERO_COMMITMENT_DECOMPOSED_AMOUNT`]
/// table is sorted.
```

helper/src/crypto.rs
```diff
@@ -0,0 +1,93 @@
...
+            lookup_table[i] = (
+                amount,
+                (ED25519_BASEPOINT_POINT + *H * Scalar::from(amount)),
+            );
+        }
+
+        lookup_table
+    });
+
+/// This function compute the zero commitment given a specific amount.
```
hinto-janai, line 63:
```suggestion
/// This function computes the zero commitment given a specific amount.
```

helper/Cargo.toml
```diff
@@ -10,12 +10,13 @@ repository = "https://github.com/Cuprate/cuprate/tree/main/consensus"
 
 [features]
 # All features off by default.
-default   = []
+default   = ["std"]
```
hinto-janai, line 13: Either the comment needs to be updated or `"std"` should be removed, if `crypto` needs `std` it should be `crypto = ["std", "..."]`.

helper/src/crypto.rs
SyntheticBird45: constant-time version will be used
https://matrix.to/#/!zPLCnZSsyeFFxUiqUZ:monero.social/$IBmC-hk2AN7XdQwo3T7StkaFDGEeRd-7XBeZenwBUCk?via=monero.social&via=matrix.org&via=kernal.eu

helper/src/crypto.rs
```diff
@@ -0,0 +1,100 @@
...
+    // OPTIMIZATION: Unlike monerod which execute a linear search across its lookup
+    // table (O(n)). Cuprate is making use of an arithmetic based constant time
+    // version (O(1)). It has been benchmarked in both hit and miss scenarios against
+    // a binary search lookup (O(log2(n))). To understand the following algorithm it
+    // is important to observe the pattern that follows the values of
+    // [`ZERO_COMMITMENT_DECOMPOSED_AMOUNT`].
+
+    // First obtain the logarithm base 10 of the amount. and extend it back to obtain
+    // the amount without its most significant digit.
+    let log = amount.checked_ilog10().unwrap();
```
Boog900, line 83:
```suggestion
    // OPTIMIZATION: Unlike monerod which execute a linear search across its lookup
    // table (O(n)). Cuprate is making use of an arithmetic based constant time
    // version (O(1)). It has been benchmarked in both hit and miss scenarios against
    // a binary search lookup (O(log2(n))). To understand the following algorithm it
    // is important to observe the pattern that follows the values of
    // [`ZERO_COMMITMENT_DECOMPOSED_AMOUNT`].

    // First obtain the logarithm base 10 of the amount. and extend it back to obtain
    // the amount without its most significant digit.
    let Some(log) = amount.checked_ilog10() else {
         // amount = 0 so H component is 0. 
         return ED25519_BASEPOINT_POINT;
    }
```

helper/src/crypto.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    // If the *rounded* version is different than the exact amount. Then
+    // there aren't only trailing zeroes behind the most significant digit.
+    // The amount is not part of the table and can calculated apart.
+    if most_significant_digit * div != amount {
+        return H_PRECOMP.vartime_multiscalar_mul([Scalar::from(amount), Scalar::ONE]);
+    }
+
+    // Calculating the index back by progressing within the powers of 10.
+    let index = (most_significant_digit + u64::from(log) * 10 - (u64::from(log) + 1)) as usize;
```
Boog900, line 97: This is probably clearer:

```suggestion
    // Calculating the index back by progressing within the powers of 10.
    // The index of the first value in the cached amount's row.
    let row_start = u64::from(log) * 9;
    // The index of the cached amount
    let index = (most_significant_digit - 1 + row_start) as usize;
```

helper/src/crypto.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Runtime initialized [`H`] generator.
+static H_PRECOMP: LazyLock<VartimeEdwardsPrecomputation> =
+    LazyLock::new(|| VartimeEdwardsPrecomputation::new([*H, ED25519_BASEPOINT_POINT]));
+
+/// Runtime initialized zero commitment lookup table
+///
+/// # Invariant
+/// This function assumes that the [`ZERO_COMMITMENT_DECOMPOSED_AMOUNT`]
+/// table is sorted.
+pub static ZERO_COMMITMENT_LOOKUP_TABLE: LazyLock<[(u64, EdwardsPoint); 172]> =
```
Boog900, line 50: I don't think we need to keep the amount (`u64`) anymore.

Boog900 (changes requested):
It would be good if you could add a test looping trough all the amounts in the table making sure we give the same commitment back from `compute_zero_commitment`

helper/src/crypto.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        lookup_table
+    });
+
+/// This function computes the zero commitment given a specific amount.
+///
+/// It will first attempt to lookup into the table of known Pre-RCT value.
+/// Compute it otherwise.
+#[expect(clippy::cast_possible_truncation)]
+pub fn compute_zero_commitments(amount: u64) -> EdwardsPoint {
```
Boog900, line 69: nit:
```suggestion
pub fn compute_zero_commitment(amount: u64) -> EdwardsPoint {
```

helper/src/crypto.rs
SyntheticBird45: Removed

helper/src/crypto.rs
SyntheticBird45: Fixed

helper/src/crypto.rs
SyntheticBird45: Added

helper/src/crypto.rs
SyntheticBird45: Changed.

Boog900 (approved):

Issue #178, opened 2024-06-19
Improve P2P address book


Boog900:
This is a tracking issue of things that should be improved in our P2P address book:

## Don't replace addresses unless unreachable

We should not replace peers in either peer list (white or gray) unless they are unreachable, this should reduce an attacker ability to fill our peer list.

## Bucket addresses

We could use a bucket system [like Bitcoin](https://bitcoin.stackexchange.com/questions/115539/bucketing-algorithm-in-bitcoin) to store our p2p address, so that we guarantee a more diverse range of addresses.

## Persist banned peers

currently banned peers are forgotten on restarts, we should keep them in the `p2p_store`

willco-1:
i'll take a shot at this!

SyntheticBird45:
Heyo @willco-1 are you still working on persistent ban list ?

PR #329, opened 2024-10-28
p2p: Implement P2P Bucket data structure

SyntheticBird45:
### What

See: https://github.com/Cuprate/cuprate/issues/178

This PR implements a "Bucket" data structure that is a collection of data that discriminates its items into "buckets" (vector of size N) following a defined function.

- Implements Bucket data structure and Bucketable trait
- Implements Bucketable for Ipv4Addr
- Added the crate to the workspace dependencies
- Added `arrayvec` as a dependency

### Why

The end goal is to use this structure to store diverse IPs of different /16 subnets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ members = [
 	"net/wire",
 	"p2p/p2p",
 	"p2p/p2p-core",
+	"p2p/bucket",
 	"p2p/dandelion-tower",
 	"p2p/async-buffer",
 	"p2p/address-book",
@@ -64,6 +65,7 @@ cuprate-levin               = { path = "net/levin"              ,default-feature
 cuprate-wire                = { path = "net/wire"               ,default-features = false}
 cuprate-p2p                 = { path = "p2p/p2p"                ,default-features = false}
 cuprate-p2p-core            = { path = "p2p/p2p-core"           ,default-features = false}
+cuprate-p2p-bucket          = { path = "p2p/p2p-bucket"         ,default-features = false}
 cuprate-dandelion-tower     = { path = "p2p/dandelion-tower"    ,default-features = false}
 cuprate-async-buffer        = { path = "p2p/async-buffer"       ,default-features = false}
 cuprate-address-book        = { path = "p2p/address-book"       ,default-features = false}
@@ -80,6 +82,7 @@ cuprate-rpc-interface       = { path = "rpc/interface"          ,default-feature
 
 # External dependencies
 anyhow                = { version = "1.0.89", default-features = false }
+arrayvec              = { version = "0.7", default-features = false }
 async-trait           = { version = "0.1.82", default-features = false }
 bitflags              = { version = "2.6.0", default-features = false }
 blake3                = { version = "1", default-features = false }
--- a/p2p/bucket/Cargo.toml
+++ b/p2p/bucket/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "cuprate-p2p-bucket"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+authors = ["SyntheticBird"]
+
+[dependencies]
+arrayvec = { workspace = true }
+rand     = { workspace = true, features = ["std", "std_rng"]}
+
+[lints]
+workspace = true
--- a/p2p/bucket/src/lib.rs
+++ b/p2p/bucket/src/lib.rs
@@ -1,0 +1,172 @@
+//! Bucket data structure
+//!
+//! A collection data structure that discriminates its unique items and place them into "buckets".
+//!
+//! The item must implement the [`Bucketable`] trait that defines how to create the discriminant
+//! from the item type. The data structure will internally contain any item into "buckets" or vectors
+//! of sized capacity `N` that regroup all the stored items with this specific discriminant.
+//!
+//! A practical example of this data structure is for storing `N` amount of IP discriminated by their subnets.
+//! You can store in each "buckets" corresponding to a `/16` subnet up to `N` IPs of that subnet.
+//!
+//! # Example
+//!
+//! ```
+//! use cuprate_p2p_bucket::Bucket;
+//! use std::net::Ipv4Addr;
+//!
+//! // Create a new bucket that can store at most 2 IPs in a particular `/16` subnet.
+//! let mut bucket = Bucket::<2,Ipv4Addr>::new();
+//!
+//! // Fulfill the `96.96.0.0/16` bucket.
+//! bucket.push("96.96.0.1".parse().unwrap());
+//! bucket.push("96.96.0.2".parse().unwrap());
+//! assert_eq!(2, bucket.len());
+//! assert_eq!(2, bucket.len_bucket(&[96_u8,96_u8]).unwrap());
+//!
+//! // Push a new IP from another subnet
+//! bucket.push("127.0.0.1".parse().unwrap());
+//! assert_eq!(3, bucket.len());
+//! assert_eq!(2, bucket.len_bucket(&[96_u8,96_u8]).unwrap());
+//! assert_eq!(1, bucket.len_bucket(&[127_u8,0_u8]).unwrap());
+//!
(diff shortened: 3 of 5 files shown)


Cargo.toml
```diff
@@ -51,35 +52,37 @@ opt-level = 3
...
-cuprate-database-service    = { path = "storage/service"        ,default-features = false}
-cuprate-txpool              = { path = "storage/txpool"         ,default-features = false}
-cuprate-pruning             = { path = "pruning"                ,default-features = false}
-cuprate-test-utils          = { path = "test-utils"             ,default-features = false}
-cuprate-types               = { path = "types"                  ,default-features = false}
-cuprate-json-rpc            = { path = "rpc/json-rpc"           ,default-features = false}
-cuprate-rpc-types           = { path = "rpc/types"              ,default-features = false}
-cuprate-rpc-interface       = { path = "rpc/interface"          ,default-features = false}
+cuprate-fast-sync           = { path = "consensus/fast-sync"  ,default-features = false}
+cuprate-consensus-rules     = { path = "consensus/rules"      ,default-features = false}
```
Boog900, line 56: Can you revert the changes to the unrelated lines please

p2p/bucket/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "cuprate-p2p-bucket"
+version = "0.1.0"
+edition = "2021"
+
```
Boog900, line 5: This will need license etc, can you copy from another crate and edit appropriately

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/// A collection data structure discriminating its unique items
+/// with a specified method. Enable a fair distribution of the
+/// items based on their discriminants when popped.
+pub struct Bucket<const N: usize, I: Bucketable> {
+    /// The storage of the bucket
+    storage: BTreeMap<I::Discriminant, ArrayVec<I, N>>,
+    /// Internal round-robin state
+    round_robin: usize,
```
Boog900, line 32: for now can you remove this round_robin pop impl please.

It should be done as a separate struct which impls `Iterator`, however until we need it I would just remove this.

Boog900 (changes requested):

p2p/bucket/src/lib.rs
SyntheticBird45: I'm actually planning on passing to the Address book PR after that. Shouldn't it be preferable to make that iterator structure in this PR instead ?

Boog900: The bucket structure is more for making sure we store a wide range of IPs, not that we select a wide range of IPs, having this round_robin makes it more likely for consecutive subnets to be all the peers we connect to.

Just randomly selecting peers across the whole structure would be better IMO, but this can be considered later.

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn push(&mut self, item: I) {
+        let discriminant = item.discriminant();
+
+        if let Some(vec) = self.storage.get_mut(&discriminant) {
+            // Check if the element already exists
+            let already_exist = vec.iter().any(|v| &item == v);
+            if !already_exist {
+                // Push the new element
+                vec.push(item);
+            }
```
hinto-janai, line 60:
```suggestion
            // Push the item if it doesn't exist.
            if !vec.contains(&item) {
                vec.push(item);
            }
```

p2p/bucket/src/lib.rs
SyntheticBird45: Fair enough.

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+//!
+//! A collection data structure that discriminates its unique items and place them into "buckets"
+//!
+//! The item must implement the [`Bucketable`] trait that defines how to create the discriminant
+//! from the item type. The data structure will internally contain any item into "buckets" or vectors
+//! of sized capacity N that regroup all the stored items with this specific discriminant.
+//!
+//! A practical example of this data structure is for storing N amount of IP discriminated by their subnets.
+//! You can store in each "buckets" corresponding to a /16 subnet up to N IPs of that subnet.
+//!
```
hinto-janai, line 11: An example usage test-doc would be nice.

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Will attempt to remove an item from the bucket
+    pub fn remove(&mut self, item: &I) -> Option<I> {
+        self.storage.get_mut(&item.discriminant()).and_then(|vec| {
+            let find = vec
+                .iter()
+                .enumerate()
+                .filter(|(_, v)| &item == v)
+                .map(|(i, _)| i)
+                .last();
+            find.map(|index| vec.swap_remove(index))
```
hinto-janai, line 78:
```suggestion
            vec.iter()
                .enumerate()
                .filter_map(|(i, v)| (item == v).then_some(i))
                .last()
                .map(|index| vec.swap_remove(index))
```

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl<const N: usize, I: Bucketable> Default for Bucket<N, I> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Bucketable for Ipv4Addr {
+    // We are discriminating by /16 subnets
```
hinto-janai, line 111:
```suggestion
    /// We are discriminating by `/16` subnets.
```

hinto-janai (commented):
Can you add this crate to here?: https://github.com/Cuprate/cuprate/blob/main/books/architecture/src/appendix/crates.md

p2p/bucket/src/lib.rs
SyntheticBird45: If I remove the current pop there is no method to actually extract an IP address from the storage. Do you want me to make the `storage` field public ? or adapt the current `pop` to take a random item (i know this contradict your statement of needing an iterator but I'm going to implement anyway in the next PR since well I'll need it)?

Boog900: Yeah adding a `pop` method that selects  a random peer would be good - although it should leave the peer in structure and just return a reference

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+//! assert_eq!(2, bucket.len_bucket(&[96_u8,96_u8]).unwrap());
+//!
+//! // Push a new IP from another subnet
+//! bucket.push("127.0.0.1".parse().unwrap());
+//! assert_eq!(3, bucket.len());
+//! assert_eq!(2, bucket.len_bucket(&[96_u8,96_u8]).unwrap());
+//! assert_eq!(1, bucket.len_bucket(&[127_u8,0_u8]).unwrap());
+//!
+//! // Attempting to push a new IP within `96.96.0.0/16` bucket will panic.
+//! ```
```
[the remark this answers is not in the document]
SyntheticBird45: Yes, It will. 
Yes, It is a bad API, because it imply you have to use `len_bucket()` before use.
Yes, I would like to replace it with `try_push` or `push` but return a result or option.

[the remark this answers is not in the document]
Boog900: Ah it's an array vec, yes it should return an option, it should not panic

p2p/bucket/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /// Will attempt to remove an item from the bucket.
+    pub fn remove(&mut self, item: &I) -> Option<I> {
+        self.storage.get_mut(&item.discriminant()).and_then(|vec| {
+            vec.iter()
+                .enumerate()
+                .filter_map(|(i, v)| (item == v).then_some(i))
+                .last()
+                .map(|index| vec.swap_remove(index))
```
Boog900, line 122: I think this could probably be made more efficient, maybe by using `find_map` instead of `filter_map`

Boog900 (approved):

PR #334, opened 2024-11-01
rpc/types 32 bit support

Boog900:
### What
Makes changes to some crates to support 32 bit systems for `rpc/types`

### Why
WASM


spirobel:
lgtm 👍 

I managed to construct a request object for get_blocks.bin with this branch.

some observations regarding the types:
Afterwards i passed the request object  to typescript and used it in the body of the fetch request. The response was then passed back to rust and parsed into a GetBlocksResponse object. 

First I used GetBlocksRequest::default() to construct the request which meant the start_height was set to 0.

As a result the finish function gave me:  "Required field was not found!"
 https://github.com/Cuprate/cuprate/blob/b6c4adc83a199886d6f932c1321857fb8a535af5/rpc/types/src/bin.rs#L298

This is because there are no Blocks in this response. 
Afterwards I set the start_height to 100. So the response contained blocks.
The result was still: "Required field was not found!"

This time this was because of 

``` rust
  let pool_info_extent = self.pool_info_extent.ok_or(ELSE)?; 
``` 

After I manually inserted

``` rust
 let pool_info_extent = PoolInfoExtent::None;
```
the Response was parsed successfully.

Suggestions:
1. Empty blocks should be a valid GetBlocksResponse object.
2. pool_info_extent should be set to None instead of throwing an error
3. the error message should indicate which field is missing.

net/epee-encoding/src/varint.rs
```diff
@@ -58,7 +59,12 @@ pub fn read_varint<B: Buf>(r: &mut B) -> Result<u64> {
 ///     assert_eq!(buf.as_slice(), expected_bytes);
 /// }
 /// ```
-pub fn write_varint<B: BufMut>(number: u64, w: &mut B) -> Result<()> {
+pub fn write_varint<B: BufMut, T: TryInto<u64>>(number: T, w: &mut B) -> Result<()> {
+    let number = number
+        .try_into()
+        .map_err(|_| "Tried to write a varint bigger than 64-bits")
+        .unwrap();
```
hinto-janai, line 66: I was worried that this would cause perf regressions but after testing it seems the compiler is smart enough in most (all?) cases to make this a no-op.

.github/workflows/ci.yml
```diff
@@ -46,6 +46,35 @@ jobs:
     - name: Spell Check
       uses: crate-ci/typos@master
 
+  # Run 32-bit support check separately.
+  bit-32-support:
```
hinto-janai, line 50: Maybe makes more sense to specify WASM 32-bit instead of general 32-bit which we probably won't/can't support.
```suggestion
  # Run 32-bit WASM support check separately.
  wasm-32-bit-support:
```

.github/workflows/ci.yml
```diff
@@ -46,6 +46,35 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        # The list of crates to check for 32 bit support
```
hinto-janai, line 55:
```suggestion
        # The list of crates to check for WASM 32 bit support
```

.github/workflows/ci.yml
```diff
@@ -46,6 +46,35 @@ jobs:
...
+        with:
+          submodules: recursive
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: wasm32-unknown-unknown
+
+      - name: Build 32 bit
```
hinto-janai, line 75:
```suggestion
      - name: Build WASM 32-bit
```

hinto-janai (approved):

net/epee-encoding/src/varint.rs
Boog900: yeah even then this would be pretty cheap anyway.

PR #335, opened 2024-11-01
Fix ci and loosen version requirements 

Boog900:
Fixes `deny` by adding an exception, also removes `cargo audit` from CI as it does the same thing as `cargo deny`.

Also loosens the version requirements on dependencies.


SyntheticBird45:
Why making loose version requirements exactly ?

Boog900:
There is just no need for the minimum versions to be this exact, it makes it harder for Cargo to find the best version to use.

SyntheticBird45 (approved):

hinto-janai:
@Boog900 conflict needs to be resolved

hinto-janai (approved):

PR #338, opened 2024-11-10
types: fix pruned `BlockCompleteEntry`

Boog900:
I misnamed this field: https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/cryptonote_protocol/cryptonote_protocol_defs.h#L123


spirobel:
The pruned tx blobs response from the getblocks.bin / get_blocks.bin endpoint are correctly parsed after this change is applied.

hinto-janai (approved):

PR #339, opened 2024-11-10
P2P: Periodically eject unreachable peers from AddressBook + Replace Peer list `IndexMap` by `BucketMap`

SyntheticBird45:
### What

Fixes parts of: https://github.com/Cuprate/cuprate/issues/178

- Implements `PeerPinger` in `cuprate-p2p`. This background task will periodically fetch a random white and gray peers and try to ping them. Replace them in the list if successful, drop it otherwise.
- Added a new `AddWhitePeer` request to `AddressBookRequest`. Needed for the `PeerPinger` to replace white peers.
- Renamed `Bucket` as `BucketMap`.
- Completed `BucketMap` API for usage within `peer_list.rs` in `cuprate-address-book`.
- Added `Bucketable` trait as a dependency for `NetZoneAddress` and implemented discrimination to `SocketAddr`.

### Why

As described in #178, Cuprate will at the moment keep indefinitely any peer contained in its list, even ones that may not be reachable anymore in the long/mid-term.

The use of `BucketMap` instead of `IndexMap` will ensure that we store at most `N` amount of IP addresses from the same subnet. This ensure a form of diversity and could potentially help at mitigating https://github.com/monero-project/research-lab/issues/126


SyntheticBird45:
@Boog900 I appreciate an ahead of time review of the API changes before I pass to test coverage.
Also I defined `N` at 8. So the bucket is storing up to 8 IP addresses of the same subnet. Do you wanna change it?

PR #342, opened 2024-11-16
Wire: fix IPv4 Endianness

Boog900:
Monero uses LE for IP addresses, because it just has to be different :D

https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/contrib/epee/include/net/net_utils_base.h#L97

### DIFF
--- a/net/wire/src/network_address/epee_builder.rs
+++ b/net/wire/src/network_address/epee_builder.rs
@@ -1,5 +1,6 @@
-use bytes::Buf;
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
+
+use bytes::Buf;
 
 use cuprate_epee_encoding::{epee_object, EpeeObjectBuilder};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


Boog900:
The reason I only just spotted this is because `monerod` won't changed IPv6 mapped IPv4 address into canonical IPv4 addresses so we were only making successful connections to address given IPv6 mapped.

hinto-janai (commented):
A couple questions:

- Does this affect how we (de)serialize `u32`-form IPv4s in RPC? e.g. the `ip` field in [`set_bans`](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#set_bans)
- Should this be updated as well? https://github.com/Cuprate/cuprate/blob/0f1ad6db1b1a3c5b3c66086ca1592a28d2012f3e/test-utils/src/test_netzone.rs#L63

Boog900:
> Does this affect how we (de)serialize u32-form IPv4s in RPC? e.g. the ip field

Yes those seem to be in LE as well. 

> Should this be updated as well? 

No that's just test code

PR #344, opened 2024-11-19
cuprated: Init

Boog900:


Boog900 (commented):
I think this PR is ready as an initial binary initialization, there is still a lot needing to b added but with people (mostly) successfully syncing I think it's ready to be reviewed.

binaries/cuprated/src/logging.rs
```diff
@@ -0,0 +1,137 @@
...
+    let appender_config = &config.tracing.file;
+    let (appender, guard) = tracing_appender::non_blocking(
+        tracing_appender::rolling::Builder::new()
+            .rotation(Rotation::DAILY)
+            .max_log_files(appender_config.max_log_files)
+            .build(logs_path(&config.fs.data_directory, config.network()))
+            .unwrap(),
+    );
+
+    // TODO: drop this when we shutdown.
```
SyntheticBird45, line 103: To fix in a later PR ?

binaries/cuprated/src/main.rs
```diff
@@ -29,8 +49,156 @@ fn main() {
...
+            txpool_write_handle,
+            context_svc.clone(),
+            block_verifier,
+            config.block_downloader_config(),
+        )
+        .await;
+
+        // Start the command listener.
+        let (command_tx, command_rx) = mpsc::channel(1);
+        std::thread::spawn(|| commands::command_listener(command_tx));
```
SyntheticBird45, line 131: Is there no other alternative a whole thread here ?

binaries/cuprated/src/main.rs
```diff
@@ -29,8 +49,156 @@ fn main() {
...
+                };
+                let context = blockchain_context.unchecked_blockchain_context();
+
+                let uptime = statics::START_INSTANT.elapsed().unwrap_or_default();
+                let (hours, minutes, second) = secs_to_hms(uptime.as_secs());
+
+                println!("STATUS:\n  uptime: {hours}h {minutes}m {second}s,\n  height: {},\n  top_hash: {}", context.chain_height, hex::encode(context.top_hash));
+            }
+        }
+    }
```
SyntheticBird45, line 201: It might grow later. Wouldn't it be better to put that in a separate file ?

binaries/cuprated/src/main.rs
Boog900: not unless you know one, I think this is the only/best way

binaries/cuprated/src/logging.rs
Boog900: yep, when a proper shutdown is added

binaries/cuprated/src/main.rs
SyntheticBird45: Tokio have async [`stdin`](https://docs.rs/tokio/latest/tokio/io/fn.stdin.html) reading and [`Lines streaming`](https://docs.rs/tokio/latest/tokio/io/struct.Lines.html)
It could just be put into a tokio task imo.

If not, then we should consider something like [`thread_priority`] crate (or reimplement the important part in `cuprate-helper`), to decrease the priority of the command thread.

SyntheticBird45: > We could, although IMO a single thread is not going to be a problem. Surely the thread will only get woken up if there is something to read from stdin?

You are right. I just checked and on Linux and Windows, I/O operations place the thread in blocking(linux)/suspended(windows) state, so the scheduler won't awake them unless the file descriptor is updated in kernel space.
This resolve it.

binaries/cuprated/src/commands.rs
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                    println!("NEW LOG FILTER: {filter}");
+                };
+
+                match output_target {
+                    OutputTarget::File => logging::modify_file_output(modify_output),
+                    OutputTarget::Stdout => logging::modify_stdout_output(modify_output),
+                }
+            }
+            Command::Status => {
```
hinto-janai, line 104:
```suggestion

            Command::Status => {
```

binaries/cuprated/src/commands.rs
```diff
@@ -0,0 +1,126 @@
...
+                    .expect(PANIC_CRITICAL_SERVICE_ERROR)
+                else {
+                    unreachable!();
+                };
+                let context = blockchain_context.unchecked_blockchain_context();
+
+                let uptime = statics::START_INSTANT.elapsed().unwrap_or_default();
+                let (hours, minutes, second) = secs_to_hms(uptime.as_secs());
+
+                println!("STATUS:\n  uptime: {hours}h {minutes}m {second}s,\n  height: {},\n  top_hash: {}", context.chain_height, hex::encode(context.top_hash));
```
hinto-janai, line 120:
```suggestion
                let (h, m, s) = secs_to_hms(uptime.as_secs());
                let height = context.chain_height;
                let top_hash = hex::encode(context.top_hash);

                println!("STATUS:\n  uptime: {h}h {m}m {s}s,\n  height: {height},\n  top_hash: {top_hash}");
```

binaries/cuprated/src/commands.rs
```diff
@@ -0,0 +1,126 @@
...
+        line.clear();
+    }
+}
+
+/// The [`Command`] handler loop.
+pub async fn io_loop(
+    mut incoming_commands: mpsc::Receiver<Command>,
+    mut context_service: BlockChainContextService,
+) -> ! {
+    while let Some(command) = incoming_commands.recv().await {
```
hinto-janai, line 86: 1/2
```suggestion
    loop {
        let Some(command) = incoming_commands.recv().await else {
            unreachable!();
        };
        
```

binaries/cuprated/src/commands.rs
```diff
@@ -0,0 +1,126 @@
...
+
+                let uptime = statics::START_INSTANT.elapsed().unwrap_or_default();
+                let (hours, minutes, second) = secs_to_hms(uptime.as_secs());
+
+                println!("STATUS:\n  uptime: {hours}h {minutes}m {second}s,\n  height: {},\n  top_hash: {}", context.chain_height, hex::encode(context.top_hash));
+            }
+        }
+    }
+
+    unreachable!()
```
hinto-janai, line 125: 2/2
```suggestion
```

binaries/cuprated/src/config/rayon.rs
```diff
@@ -0,0 +1,18 @@
...
+pub struct RayonConfig {
+    /// The number of threads to use for the [`rayon::ThreadPool`].
+    pub threads: usize,
+}
+
+impl Default for RayonConfig {
+    fn default() -> Self {
+        Self {
+            // 75% available threads.
+            threads: (std::thread::available_parallelism().unwrap().get() * 3).div_ceil(4),
```
hinto-janai, line 15: `cuprate_helper::thread` can be used for thread percents here and elsewhere.

binaries/cuprated/src/logging.rs
```diff
@@ -0,0 +1,137 @@
...
+        >,
+    >,
+> = OnceLock::new();
+
+/// The [`Filter`] used to alter cuprated's log output.
+pub struct CupratedTracingFilter {
+    pub level: LevelFilter,
+}
+
+impl Display for CupratedTracingFilter {
```
hinto-janai, line 58:
```suggestion
// Custom display behavior for command output.
impl Display for CupratedTracingFilter {
```

binaries/cuprated/src/logging.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/// Initialize [`tracing`] for logging to stdout and to a file.
+pub fn init_logging(config: &Config) {
+    // initialize the stdout filter, set `STDOUT_FILTER_HANDLE` and create the layer.
+    let (stdout_filter, stdout_handle) = ReloadLayer::new(CupratedTracingFilter {
+        level: config.tracing.stdout.level,
+    });
+
+    drop(STDOUT_FILTER_HANDLE.set(stdout_handle));
```
hinto-janai, line 87: Why not `unwrap` here to assert it was only set once? `CupratedTracingFilter` needs `#[derive(Debug)]` for this.
```suggestion
    STDOUT_FILTER_HANDLE.set(stdout_handle).unwrap();
```

binaries/cuprated/src/logging.rs
```diff
@@ -0,0 +1,137 @@
...
+    );
+
+    // TODO: drop this when we shutdown.
+    forget(guard);
+
+    // initialize the appender filter, set `FILE_WRITER_FILTER_HANDLE` and create the layer.
+    let (appender_filter, appender_handle) = ReloadLayer::new(CupratedTracingFilter {
+        level: appender_config.level,
+    });
+    drop(FILE_WRITER_FILTER_HANDLE.set(appender_handle));
```
hinto-janai, line 110: Same here
```suggestion
    FILE_WRITER_FILTER_HANDLE.set(appender_handle).unwrap();
```

binaries/cuprated/src/main.rs
```diff
@@ -29,8 +46,109 @@ fn main() {
...
+            blockchain_read_handle,
+            txpool_write_handle,
+            context_svc.clone(),
+            block_verifier,
+            config.block_downloader_config(),
+        )
+        .await;
+
+        // Start the command listener.
+        let (command_tx, command_rx) = mpsc::channel(1);
```
hinto-janai, line 127: This applies to any `'static` data but these channels can be leaked such that edge-cases with thread-exit order + channel send/recvs don't panic. Another method is returning on channel errors in `command_listener` and `io_loop`.

Contrived yet possible panic:

- `command_listener` thread receives command
- CTRL+C is received
- `io_loop` thread exits
- `command_listener` thread panics on `blocking_send()`

Another example (more likely):
- CTRL+C is received
- `command_listener` thread exits before `io_loop` thread
- `io_loop` thread panics on `recv()`

hinto-janai (commented):
Could you write a short `//!` on each new file introduced explaining what it is for?

binaries/cuprated/src/blockchain/syncer.rs
```diff
@@ -99,11 +87,18 @@ where
...
+                    tracing::info!("Received stop signal, stopping block downloader");
                     break;
                 }
                 batch = block_batch_stream.next() => {
                     let Some(batch) = batch else {
+                        check_update_blockchain_context(&mut context_svc, &mut blockchain_ctx).await?;
+                        let raw_blockchain_context = blockchain_ctx.unchecked_blockchain_context();
+
+                        if check_behind_peers(raw_blockchain_context, &mut clearnet_interface).await? {
+                            tracing::info!("Synchronised with the network.");
```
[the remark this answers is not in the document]
Boog900: never :gb:

binaries/cuprated/src/main.rs
Boog900: I added error handling to the 2 loops, however shutdown in general needs to be improved which I plan to work on in a future PR.

p2p/address-book/src/store.rs
```diff
@@ -39,10 +39,13 @@ pub(crate) fn save_peers_to_disk<Z: BorshNetworkZone>(
     })
     .unwrap();
 
-    let file = cfg
-        .peer_store_directory
-        .join(format!("{}_p2p_state", Z::NAME));
-    spawn_blocking(move || fs::write(&file, &data))
+    let dir = cfg.peer_store_directory.clone();
+    let file = dir.join(format!("{}_p2p_state", Z::NAME));
```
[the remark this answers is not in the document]
Boog900: I changed this to just:
```
$peer_store_directory/{Z::NAME}
```

binaries/cuprated/src/blockchain/syncer.rs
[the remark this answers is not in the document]
hinto-janai: Mentioned elsewhere but this may actually matter to be output compatible with `monerod`, probably ok for now though

binaries/cuprated/src/config/rayon.rs
hinto-janai: I think this was missed but can be a TODO.

binaries/cuprated/src/blockchain/syncer.rs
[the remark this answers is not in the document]
hinto-janai: Mentioned elsewhere but this may actually matter to be output compatible with `monerod`, probably ok for now though.

hinto-janai (approved):
Last review comment was also missed:

> 

> Could you write a short //! on each new file introduced explaining what it is for?

Can be done later I suppose.

binaries/cuprated/src/config/rayon.rs
Boog900: Ah yeah I didn't do this because `cuprate_helper::thread` rounds down, should I just change them to round up?

hinto-janai: Sure update docs too please

Issue #348, opened 2024-11-25
`get_range` with `heed` does not work.


Boog900:
## Bug
`heed`'s  `get_range` function returns incorrect results when a different ordering is used with the DB/table.

## Steps to reproduce
Add this:
```rust
        for (v, vv) in table.get_range(100..249).unwrap().zip(100..249) {
            assert_eq!(v.unwrap(), vv);
        }
```

After this:
https://github.com/Cuprate/cuprate/blob/f3c1a5c2aa4629bf69b75268de21fc9112f09405/storage/database/src/backend/tests.rs#L377-L386

---

I think this is because heed's range cursor compares raw bytes here: https://docs.rs/heed/latest/src/heed/iterator/range.rs.html#416

Boog900:
Yeah slightly embarrassing but I went too aggressive on trying to simplify the test case, updated the issue with the new test case that actually fails due to this issue.

Boog900:
https://github.com/meilisearch/heed/pull/289

SyntheticBird45:
`heed` has been replaced with Tapes and Fjall

PR #350, opened 2024-11-27
database: Test heed get_range

gzalz:
### What
Test the get_range function in `storage/database/src/backend/heed/database.rs`. 

### Why
This is meant to address issue #348 .

### Where
This is related to the heed database backend and related tests.

### How
This PR is meant to address a perceived misunderstanding in issue #348 . The table insertions in the test have incrementing keys but all have a value equal to `0u64`.

`get_range` is defined to return Values rather than Keys, as per the function signature, yet in the reproduction code all the Values must equal 0. 

```
impl<T: Table> DatabaseIter<T> for HeedTableRo<'_, T> {
    #[inline]
    fn get_range<'a, Range>(
        &'a self,
        range: Range,
    ) -> Result<impl Iterator<Item = Result<T::Value, RuntimeError>> + 'a, RuntimeError>
    where
        Range: RangeBounds<T::Key> + 'a
```

If we wish to test `get_range` we should be checking the Values within the new loop, as this is what get_range returns. I believe the functionality of `get_range` is correct.

### DIFF
--- a/storage/database/src/backend/tests.rs
+++ b/storage/database/src/backend/tests.rs
@@ -360,7 +360,7 @@ fn tables_are_sorted() {
     // Insert range, assert each new
     // number inserted is the minimum `last()` value.
     for key in RANGE {
-        table.put(&key, &(key as u64)).unwrap();
+        table.put(&key, &u64::from(key)).unwrap();
         table.contains(&key).unwrap();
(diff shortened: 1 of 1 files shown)


Boog900:
Thank you for this, but as explained in the comment on the issue my example was bad.

PR #353, opened 2024-11-28
workspace: add/fix 1.83 lints

hinto-janai:
### What
Adds/fixes lints from [Rust 1.83](https://blog.rust-lang.org/2024/11/28/Rust-1.83.0.html).

Added clippy lints:
- [`non_zero_suggestions`](https://rust-lang.github.io/rust-clippy/master/index.html#non_zero_suggestions)
- [`manual_is_power_of_two`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_is_power_of_two)
- [`used_underscore_items`](https://rust-lang.github.io/rust-clippy/master/index.html#used_underscore_items)

Added `rustc` lint group:
- [`rust_2024_compatibility`](https://doc.rust-lang.org/rustc/lints/groups.html)


storage/database/src/error.rs
```diff
@@ -7,6 +7,9 @@ use std::fmt::Debug;
 /// Alias for a thread-safe boxed error.
 type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
+/// [`Result`] with [`RuntimeError`] as the error.
+pub type DbResult<T> = Result<T, RuntimeError>;
+
```
hinto-janai, line 12: `clippy::type_complexity` is more sensitive, there's some in `p2p/` silenced with `#[expect]`.

There's a pattern of doing `type Result<T> = Result<T, $common_crate_error>` [^1] [^2] [^3] so I thought we may as well fix https://github.com/Cuprate/cuprate/actions/runs/12071739897/job/33664158034?pr=353#step:11:51 with this type rather than `#[expect]`.

Lots of diffs in this PR are just moving over to `DbResult`. This could be `DatabaseResult` too but I think it's bit long for something that will appear in a lot of signatures. I think `Result` would clash / be confusing.

[^1]: https://doc.rust-lang.org/1.82.0/std/io/type.Result.html
[^2]: https://doc.rust-lang.org/1.82.0/std/fmt/type.Result.html
[^3]: https://docs.rs/heed/latest/heed/type.Result.html

consensus/context/src/difficulty.rs
```diff
@@ -328,8 +328,7 @@ fn next_difficulty(
         time_span = 1;
     }
 
-    // TODO: do checked operations here and unwrap so we don't silently overflow?
-    (windowed_work * u128::from(hf.block_time().as_secs()) + time_span - 1) / time_span
+    (windowed_work * u128::from(hf.block_time().as_secs())).div_ceil(time_span)
```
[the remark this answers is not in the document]
Boog900: I think we still need the `TODO` , right?

consensus/context/src/difficulty.rs
```diff
@@ -328,8 +328,7 @@ fn next_difficulty(
...
 }
```
hinto-janai, line 332:
```suggestion
    windowed_work
        .checked_mul(u128::from(hf.block_time().as_secs()))
        .unwrap()
        .div_ceil(time_span)
}
```
```suggestion
    // TODO: do `checked_mul` here and unwrap so we don't silently overflow?
    (windowed_work * u128::from(hf.block_time().as_secs())).div_ceil(time_span)
}
```

consensus/context/src/difficulty.rs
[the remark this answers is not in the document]
hinto-janai: https://github.com/Cuprate/cuprate/pull/353#discussion_r1862605054

consensus/context/src/difficulty.rs
Boog900: I'm still sure if unwrapping is the best option, so I am going to keep the TODO

PR #361, opened 2025-01-07
rpc/types: add missing defaults to fields

Boog900:
These fields were missing defaults: https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/rpc/core_rpc_server_commands_defs.h#L265-L266

reported by @spirobel

### DIFF
--- a/rpc/types/src/bin.rs
+++ b/rpc/types/src/bin.rs
@@ -132,8 +132,8 @@ define_request_and_response! {
         start_height: u64,
         current_height: u64,
         output_indices: Vec<BlockOutputIndices>,
-        daemon_time: u64,
-        pool_info: PoolInfo,
+        daemon_time: u64 = default_zero::<u64>(), "default_zero",
+        pool_info: PoolInfo = PoolInfo::None, "PoolInfo::default",
     }
 }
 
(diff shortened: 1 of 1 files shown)


rpc/types/src/bin.rs
```diff
@@ -132,8 +132,8 @@ define_request_and_response! {
         start_height: u64,
         current_height: u64,
         output_indices: Vec<BlockOutputIndices>,
-        daemon_time: u64,
-        pool_info: PoolInfo,
+        daemon_time: u64 = default_zero::<u64>(), "default_zero",
+        pool_info: PoolInfo = PoolInfo::None, "PoolInfo::default",
```
hinto-janai, line 136:
```suggestion
        // FIXME: use `default()` after <https://github.com/Cuprate/cuprate/pull/355>
        pool_info: PoolInfo = PoolInfo::None, "PoolInfo::default",
```

hinto-janai (approved):
Maybe worth noting that this is indirectly fixed by https://github.com/Cuprate/cuprate/pull/355#discussion_r1897459209, although that may be temporary and this change is needed regardless.

Boog900:
Yeah I don't like the idea of making every field have a default, I am ok with it as a temporary solution though.

Monero actually gives every field a default as well, which has lead to problems: https://github.com/monero-project/monero/pull/9570

PR #365, opened 2025-01-17
cuprated: `killswitch` and `version` module

hinto-janai:
### What
- Adds a killswitch mechanism for releases before v1
- Changes `--version` to output more information in JSON

### Why
```
Tue, Jan 14, 2025, 18:28:22 - hinto: I've also considered not releasing binaries in case `cuprated` exposes major network problems, no binaries would a hacky but real limiter to that problem
Tue, Jan 14, 2025, 18:29:02 - hinto: or if `cuprated` itself is the issue
Tue, Jan 14, 2025, 18:30:04 - SyntheticBird: The `what is important / what should be focused on in Cuprate` is going to be interesting discuss (to re discuss in an extent).
Tue, Jan 14, 2025, 18:30:16 - boog900: yeah we will need to stress that alpha means don't trust with anything critical
Tue, Jan 14, 2025, 18:31:05 - Rucknium: Zcash has a different release philosophy. At least with the old C++ zcashd, old releases had an intentional killswitch after their "expiration date" ev
en without a hard fork. Just to put it in context.
Tue, Jan 14, 2025, 18:32:02 - boog900: Oh yeah that was another idea, we could put a killswitch in the alpha binary to force people to update.
```
```
Thu, Jan 16, 2025, 21:28:13 - hinto: After reading https://github.com/monero-project/monero/pull/9668#issuecomment-2594104964 I think it may be a good idea to open up the security discussion
and gather wider consensus before releasing a v1 `cuprated` binary. If the security scope is being set at a malicious upstream or miscompilation then I think `cuprated` could represent a much
 more real and practical tool to be used by a malicious actor if something like a chain split bug were to be found.\

For reference, there were audits for `reth` and `zebrad` before v1 release:
- https://github.com/paradigmxyz/reth/blob/main/audit/sigma_prime_audit_v2.pdf
- https://www.nccgroup.com/media/bw3nq0vm/_ncc_group_zcashfoundation_e005955_report_2023-06-27_v10.pdf

For the alpha build, releasing build instructions / documentation should be okay without further review...? An argument could be made that it's not difficult to use an alpha `cuprated` for th
e same attacks but then again there's probably cheaper and more profitable ways to attack the network?
Thu, Jan 16, 2025, 21:30:21 - Mαlοri: I agree with the comment
Thu, Jan 16, 2025, 21:30:54 - hinto: Regardless I think the killswitch is a good idea, perhaps just a thread that slowly loops on `assert!(current_timestamp() < some_timestamp)` sounds ok?
Thu, Jan 16, 2025, 21:31:51 - SyntheticBird: fwiw I would like to avoid the killswitch to kill the node at runtime but rather block at next start
Thu, Jan 16, 2025, 21:32:12 - boog900:  no it should kill at runtime IMO
```


binaries/cuprated/src/killswitch.rs
```diff
@@ -0,0 +1,82 @@
...
+use cuprate_helper::time::current_unix_timestamp;
+
+/// Assert that this is not a v1 release and an alpha release.
+pub const _: () = {
+    const_format::assertcp_ne!(
+        crate::constants::MAJOR_VERSION,
+        "1",
+        "`cuprated` major version is 1, killswitch module should be deleted."
+    );
+};
```
hinto-janai, line 26: 2/6

This module should be removed after v1 release, this helps remind us.

binaries/cuprated/src/killswitch.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// The killswitch activates if the current timestamp is ahead of this timestamp.
+///
+/// Sat Mar 01 2025 05:00:00 GMT+0000
+pub const KILLSWITCH_ACTIVATION_TIMESTAMP: u64 = 1740805200;
```
hinto-janai, line 31: 3/6

The killswitch timestamp. Upon reaching this timestamp, `cuprated` will `std::process::exit`.

We can decide and change the value before each release.

binaries/cuprated/src/killswitch.rs
```diff
@@ -0,0 +1,82 @@
...
+    if current_ts > KILLSWITCH_ACTIVATION_TIMESTAMP {
+        print_killswitch_msg(format!("The killswitch activation timestamp for alpha builds has passed (current_ts: {current_ts}, killswitch_activation_timestamp: {KILLSWITCH_ACTIVATION_TIMESTAMP})"));
+        exit(88);
+    }
+}
+
+/// Spawn a thread that sleeps until the [`KILLSWITCH_ACTIVATION_TIMESTAMP`] activates.
+pub fn init_killswitch() {
+    // Check if we should exit immediately.
+    killswitch();
```
hinto-janai, line 66: 4/6

The killswitch activates in 2 ways:

1. On initial startup
1. During runtime

This line checks if we're already ahead of the timestamp and will exit if so.

```bash
timedatectl set-ntp false
timedatectl set-time 2025-03-02
date -u
```

> Sun Mar  2 05:00:01 AM UTC 2025

```bash
cargo run --bin cuprated
```

> killswitch: The killswitch activation timestamp for alpha builds has passed (current_ts: 1740891602, killswitch_activation_timestamp: 1740805200). `cuprated` will now exit. For more details on why this exists, see: <https://github.com/Cuprate/cuprate/blob/main/binaries/cuprated/src/killswitch.rs>.

binaries/cuprated/src/killswitch.rs
```diff
@@ -0,0 +1,82 @@
...
+        let sleep_duration = Duration::from_secs(KILLSWITCH_ACTIVATION_TIMESTAMP - current_ts);
+        std::thread::sleep(sleep_duration);
+
+        // To account for any miscalculated or drifted sleep time,
+        // loop until the killswitch activates.
+        loop {
+            killswitch();
+            std::thread::sleep(Duration::from_secs(30));
+        }
+    });
```
hinto-janai, line 81: 5/6

Else we exit at runtime when we reach the timestamp.

binaries/cuprated/src/killswitch.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn print_killswitch_msg(msg: String) {
+        eprintln!("killswitch: {msg}. `cuprated` will now exit. For more details on why this exists, see: <https://github.com/Cuprate/cuprate/blob/main/binaries/cuprated/src/killswitch.rs>.");
+    }
+
+    let current_ts = current_unix_timestamp();
+
+    if current_ts < SYSTEM_CLOCK_SANITY_TIMESTAMP {
+        print_killswitch_msg(format!("The system clock is too far behind (current_ts: {current_ts}) and is not reliable to use"));
+        exit(66);
+    }
```
hinto-janai: 6/6

This guards against a system with a messed up system clock from evading the killswitch.

```bash
timedatectl set-ntp false
timedatectl set-time 2010-03-06
date -u
```

> Sat Mar  6 05:00:01 AM UTC 2010

```bash
cargo run --bin cuprated
```

> killswitch: The system clock is too far behind and is not reliable to use (current_ts: 1267851602, killswitch_activation_timestamp: 1740805200). `cuprated` will now exit. For more details on why this exists, see: <https://github.com/Cuprate/cuprate/blob/main/binaries/cuprated/src/killswitch.rs>.

binaries/cuprated/src/version.rs
```diff
@@ -0,0 +1,73 @@
...
+    blockchain_db_version: u64,
+    /// `cuprated`'s semantic version.
+    semantic_version: &'static str,
+    /// Build type, either `debug` or `release`.
+    build: &'static str,
+    /// Git commit hash of the build.
+    commit: &'static str,
+    /// Timestamp at which `cuprated`'s killswitch activates.
+    killswitch_timestamp: u64,
+}
```
hinto-janai, line 42: This replaces the `--version` output for:
1. More info
2. JSON (for easier interopt)

This is added in this PR because we can add killswitch info.

```bash
cargo run --bin cuprated -- --version
```
```json
{
  "major_version": 0,
  "minor_version": 0,
  "patch_version": 1,
  "rpc_major_version": 3,
  "rpc_minor_version": 14,
  "rpc_version": 196622,
  "hardfork": 16,
  "blockchain_db_version": 0,
  "semantic_version": "0.0.1",
  "build": "debug",
  "commit": "fb6ec1710b64c8a822aea227c0140449a681b94a",
  "killswitch_timestamp": 1740805200
}
```

SyntheticBird45 (changes requested):
Let a comment regarding 3/6.

binaries/cuprated/src/killswitch.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/// Spawn a thread that sleeps until the [`KILLSWITCH_ACTIVATION_TIMESTAMP`] activates.
+pub fn init_killswitch() {
+    // Check if we should exit immediately.
+    killswitch();
+
+    // Else spawn a thread that waits until we should.
+    std::thread::spawn(|| -> ! {
```
Boog900, line 69: could we make this an async task so we don't need another thread? We would have to move the `init_killswitch` after the tokio runtime is initialized

binaries/cuprated/src/killswitch.rs
hinto-janai: I'm ok with editing this before releases if they are every 6-8 weeks. FWIW I don't think making it easier to bypass this is good but here's a patch for `env!()` + `build.rs`, thoughts?:
```diff
diff --git a/binaries/cuprated/build.rs b/binaries/cuprated/build.rs
new file mode 100644
index 00000000..ae7129e7
--- /dev/null
+++ b/binaries/cuprated/build.rs
@@ -0,0 +1,20 @@
+fn main() {
+    set_killswitch_timestamp_env();
+}
+
+/// This sets the git `CUPRATED_KILLSWITCH_TIMESTAMP` environment variable.
+fn set_killswitch_timestamp_env() {
+    // If the env variable is set, use it.
+    let timestamp: u64 = if let Some(timestamp) = option_env!("CUPRATED_KILLSWITCH_TIMESTAMP") {
+        timestamp
+            .parse()
+            .expect("Could not parse CUPRATED_KILLSWITCH_TIMESTAMP as u64.")
+    } else {
+        // Else use a fallback.
+        //
+        // Sat Mar 01 2025 05:00:00 GMT+0000
+        1740805200
+    };
+
+    println!("cargo::rustc-env=CUPRATED_KILLSWITCH_TIMESTAMP={timestamp}");
+}
diff --git a/binaries/cuprated/src/killswitch.rs b/binaries/cuprated/src/killswitch.rs
index 7ce29af1..54e0445a 100644
--- a/binaries/cuprated/src/killswitch.rs
+++ b/binaries/cuprated/src/killswitch.rs
@@ -12,7 +12,7 @@
 //! This behavior is limited to an alpha build;
 //! this module will be removed after a stable v1 release.
 
-use std::{process::exit, time::Duration};
+use std::{process::exit, sync::LazyLock, time::Duration};
 
 use cuprate_helper::time::current_unix_timestamp;
 
@@ -25,28 +25,37 @@ const _: () = {
     );
 };
 
+/// A timestamp known to have been passed.
+///
+/// This is an arbitrary timestamp used for
+/// sanity checking the system's clock to make
+/// sure it is not overly behind.
+///
+/// Fri Jan 17 2025 14:19:10 GMT+0000
+const SYSTEM_CLOCK_SANITY_TIMESTAMP: u64 = 1737123550;
+
 /// The killswitch activates if the current timestamp is ahead of this timestamp.
 ///
-/// Sat Mar 01 2025 05:00:00 GMT+0000
-pub const KILLSWITCH_ACTIVATION_TIMESTAMP: u64 = 1740805200;
+/// This env variable is set in `cuprated`'s `build.rs`.
+pub static KILLSWITCH_TIMESTAMP: LazyLock<u64> = LazyLock::new(|| {
+    let timestamp = env!("CUPRATED_KILLSWITCH_TIMESTAMP")
+        .parse()
+        .expect("Could not parse killswitch timestamp as u64");
+
+    assert!(timestamp > SYSTEM_CLOCK_SANITY_TIMESTAMP);
+
+    timestamp
+});
 
 /// Check if the system clock is past a certain timestamp,
 /// if so, exit the entire program.
 fn killswitch() {
-    /// A timestamp known to have been passed.
-    ///
-    /// This is an arbitrary timestamp used for
-    /// sanity checking the system's clock to make
-    /// sure it is not overly behind.
-    ///
-    /// Fri Jan 17 2025 14:19:10 GMT+0000
-    const SYSTEM_CLOCK_SANITY_TIMESTAMP: u64 = 1737123550;
-
     let current_ts = current_unix_timestamp();
 
     // Prints a generic killswitch message.
     let print_killswitch_msg = |msg| {
-        eprintln!("killswitch: {msg}. (current_ts: {current_ts}, killswitch_activation_timestamp: {KILLSWITCH_ACTIVATION_TIMESTAMP}). `cuprated` will now exit. For more details on why this exists, see: <https://github.com/Cuprate/cuprate/pull/365>.");
+        let timestamp =
+        eprintln!("killswitch: {msg}. (current_ts: {current_ts}, killswitch_activation_timestamp: {}). `cuprated` will now exit. For more details on why this exists, see: <https://github.com/Cuprate/cuprate/pull/365>.", *KILLSWITCH_TIMESTAMP);
     };
 
     if current_ts < SYSTEM_CLOCK_SANITY_TIMESTAMP {
@@ -54,13 +63,13 @@ fn killswitch() {
         exit(66);
     }
 
-    if current_ts > KILLSWITCH_ACTIVATION_TIMESTAMP {
+    if current_ts > *KILLSWITCH_TIMESTAMP {
         print_killswitch_msg("The killswitch activation timestamp for alpha builds has passed.");
         exit(88);
     }
 }
 
-/// Spawn a thread that sleeps until the [`KILLSWITCH_ACTIVATION_TIMESTAMP`] activates.
+/// Spawn a thread that sleeps until the [`KILLSWITCH_TIMESTAMP`] activates.
 pub fn init_killswitch() {
     // Check if we should exit immediately.
     killswitch();
@@ -69,7 +78,7 @@ pub fn init_killswitch() {
     std::thread::spawn(|| -> ! {
         // Sleep until killswitch activation.
         let current_ts = current_unix_timestamp();
-        let sleep_duration = Duration::from_secs(KILLSWITCH_ACTIVATION_TIMESTAMP - current_ts);
+        let sleep_duration = Duration::from_secs(*KILLSWITCH_TIMESTAMP - current_ts);
         std::thread::sleep(sleep_duration);
 
         // To account for any miscalculated or drifted sleep time,
diff --git a/binaries/cuprated/src/version.rs b/binaries/cuprated/src/version.rs
index 84540023..97202ea1 100644
--- a/binaries/cuprated/src/version.rs
+++ b/binaries/cuprated/src/version.rs
@@ -10,7 +10,7 @@ use cuprate_types::HardFork;
 
 use crate::{
     constants::{MAJOR_VERSION, MINOR_VERSION, PATCH_VERSION, VERSION},
-    killswitch::KILLSWITCH_ACTIVATION_TIMESTAMP,
+    killswitch::KILLSWITCH_TIMESTAMP,
 };
 
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
@@ -56,7 +56,7 @@ impl CupratedVersionInfo {
             semantic_version: VERSION,
             build: BUILD,
             commit: COMMIT,
-            killswitch_timestamp: KILLSWITCH_ACTIVATION_TIMESTAMP,
+            killswitch_timestamp: *KILLSWITCH_TIMESTAMP,
         }
     }
 }
```

binaries/cuprated/src/killswitch.rs
hinto-janai: Are you ok with config/CLI/logging stuff running too? The thread is suspended until exit too so it is light.

Boog900: I would be but yeah we may as well just keep the thread.

binaries/cuprated/src/killswitch.rs
Boog900: IMO keeping it as a fixed constant is better, someone could just recompile the same code to get around this right?

hinto-janai: Yes, pull repo, edit const to 0, build.

Boog900: no I mean your patch all they have to do is just run `cargo build` to reset the timer. Yeah with the current code If you really want to you could reset/disable the timer manually but at that point you may as well just pull from the repo again.

SyntheticBird45: > FWIW I don't think making it easier to bypass this is good but here's a patch for env!() + build.rs, thoughts?:

I may be dumb on this one but I think you forgot to add the build timestamp and period to `CUPRATED_KILLSWITCH_TIMESTAMP`. `set_killswitch_timestamp_env()` literally just set manually `1740805200` (since `CUPRATED_KILLSWITCH_TIMESTAMP` env variable will never be set)

The idea was that the `else {` branch return `SystemTime::now().duration_since(UNIX_EPOCH).as_sec() + CUPRATED_TESTING_DURATION`.

> no I mean your patch all they have to do is just run cargo build to reset the timer.

Yes. Literally just have to rebuild it to actualize the timer. But now I see how it's not very useful as the alpha will not distribute binaries of any sort, so people will build manually and have different killswitch timestamp (no killswitch timestamp to an extent). PS: Gonna be very funny to see all the cuprate alpha node simultaneously disconnect from p2p network.

> I'm ok with editing this before releases if they are every 6-8 weeks.

This could be re discuss next-meeting fwiw, but as said prior, I find 6 weeks is too short, but 8 weeks seems like a good balance to me.

SyntheticBird45 (approved):

Boog900 (approved):

PR #366, opened 2025-01-17
Consensus: Improve API

Boog900:
### What

Makes the consensus crates easier to interact with.

### Why

Fast sync is hard to add to the binary with the current interface, also this should be done at some point anyway.

### How
removes the `Service` interface for block/tx verification and adding a separate function for getting the blockchain context.


consensus/context/src/lib.rs
```diff
@@ -403,11 +364,20 @@ pub enum BlockChainContextResponse {
...
     channel: PollSender<task::ContextTaskRequest>,
 }
 
-impl Service<BlockChainContextRequest> for BlockChainContextService {
+impl BlockchainContextService {
+    /// Get the current [`BlockchainContext`] from the cache.
+    pub fn blockchain_context(&mut self) -> &BlockchainContext {
+        self.cached_context.load()
+    }
+}
```
Boog900, line 378: Getting the `BlockchainContext` is now done by this function instead of a `tower::Service` request.

Cargo.toml
```diff
@@ -100,6 +100,7 @@ cuprate-zmq-types         = { path = "zmq/types",                 default-featur
 
 # External dependencies
 anyhow                = { version = "1", default-features = false }
+arc-swap              = { version = "1", default-features = false }
```
Boog900, line 103: I've pulled in `arc-swap`, probably not necessary as a `RwLock` would be fast enough, thoughts?

consensus/src/transactions.rs
```diff
@@ -1,20 +1,35 @@
...
+//! let tx = start_tx_verification()
+//!              .append_txs(vec![tx])
+//!              .prepare()?
+//!              .just_semantic(HardFork::V9)
+//!              .queue(&batch_verifier)?;
+//!
+//! assert!(batch_verifier.verify());
+//! Ok(())
+//! # }
+//! ```
```
Boog900, line 27: The other big change is the block/transaction verification no longer uses a `Service` interface. This also means we no longer need to use an `Arc` or `Mutex` in parts of the tx.

Cargo.toml
SyntheticBird45: Only benchmark would say so and I expect it to be drastically different between cpu architectures. I would still say i better use something that guarantee atomic operations than one who doesn't. It's free improvement imo for almost free maintenance burden. I also bet arc-swap is already included into Cargo.lock anyway.

consensus/src/block/free.rs
```diff
@@ -7,6 +7,35 @@ use cuprate_types::TransactionVerificationData;
...
+    for (i, tx_hash) in block.transactions.iter().enumerate() {
+        if tx_hash != &txs[i].tx_hash {
+            let at_index = txs[i..]
+                .iter()
+                .position(|tx| &tx.tx_hash == tx_hash)
+                .ok_or(ExtendedConsensusError::TxsIncludedWithBlockIncorrect)?;
+
+            txs.swap(i, at_index);
+        }
+    }
```
hinto-janai, line 28: This ordering algo seems to error:
```rust
//! ```Cargo.toml
//! [package]
//! name = "order"
//! version = "0.1.0"
//! edition = "2021"
//!
//! [dependencies]
//! rand = "*"
//! ```

use rand::{seq::SliceRandom, thread_rng};

fn main() {
    // generate 100 transaction hashes
    let v1: Vec<[u8; 32]> = (0..100).map(|_| rand::random()).collect();

    // create a clone, but ordered differently
    let mut v2 = v1.clone();
    v2.shuffle(&mut thread_rng());

    // both vecs contain the same hashes but are in different ordering
    assert_eq!(v1.len(), v2.len());
    assert!(v1 != v2);

    // ordering function errors
    assert!(order_transactions(v1, v2).is_err());
}

fn order_transactions(v1: Vec<[u8; 32]>, mut v2: Vec<[u8; 32]>) -> Result<(), ()> {
    for (i, tx_hash) in v1.iter().enumerate() {
        if tx_hash != &v2[i] {
            let Some(at_index) = v2[i..].iter().position(|tx| tx == tx_hash) else {
                return Err(());
            };

            v2.swap(i, at_index);
        }
    }

    Ok(())
}
```

hinto-janai: Also here are some interesting perf tests:
```
full_clone: 0.12496812
iter_branch_clone: 0.15211749
iter_position_swap: 2.227751
```
```rust
//! ```Cargo.toml
//! [package]
//! name = "order"
//! version = "0.1.0"
//! edition = "2021"
//!
//! [dependencies]
//! rand = "*"
//! ```

use std::{hint::black_box as b, time::Instant};

use rand::{seq::SliceRandom, thread_rng};

fn main() {
    let v1: Vec<[u8; 32]> = (0..100).map(|_| rand::random()).collect();
    let mut v2: Vec<[u8; 32]> = v1.clone();
    v2.shuffle(&mut thread_rng());

    assert_eq!(v1.len(), v2.len());
    assert!(v1 != v2);

    // clone entire vec if `!=`
    test(b(v1.clone()), b(v2.clone()), full_clone, "full_clone");

    // iter, clone hash if `!=`
    test(
        b(v1.clone()),
        b(v2.clone()),
        iter_branch_clone,
        "iter_branch_clone",
    );

    // same algo as PR but ignores errors
    test(
        b(v1.clone()),
        b(v2.clone()),
        iter_position_swap,
        "iter_position_swap",
    );
}

fn test(
    v1: Vec<[u8; 32]>,
    v2: Vec<[u8; 32]>,
    f: fn(v1: &Vec<[u8; 32]>, v2: &mut Vec<[u8; 32]>),
    name: &str,
) {
    let now = Instant::now();
    for _ in 0..1000000 {
        f(&b(v1.clone()), &mut b(v2.clone()));
    }
    println!("{name}: {}", now.elapsed().as_secs_f32());
}

fn full_clone(v1: &Vec<[u8; 32]>, v2: &mut Vec<[u8; 32]>) {
    if v1 != v2 {
        *v2 = v1.clone();
    }
}

fn iter_branch_clone(v1: &Vec<[u8; 32]>, v2: &mut Vec<[u8; 32]>) {
    for (h1, h2) in v1.iter().zip(v2) {
        if h1 != h2 {
            *h2 = *h1;
        }
    }
}

fn iter_position_swap(v1: &Vec<[u8; 32]>, v2: &mut Vec<[u8; 32]>) {
    for (i, tx_hash) in v1.iter().enumerate() {
        if tx_hash != &v2[i] {
            if let Some(at_index) = v2[i..].iter().position(|tx| tx == tx_hash) {
                v2.swap(i, at_index);
            }
        }
    }
}
```

consensus/src/transactions.rs
```diff
@@ -40,183 +55,223 @@ mod free;
...
+    prepped_txs: Vec<TransactionVerificationData>,
+}
+
+impl VerificationWanted {
+    /// Only semantic verification.
+    ///
+    /// Semantic verification is verification that can done without other blockchain data. The [`HardFork`]
+    /// is technically other blockchain data but excluding it reduces the amount of things that can be checked
+    /// significantly, and it is easy to get compared to other blockchain data needed for contextual validation.
+    pub fn just_semantic(self, hf: HardFork) -> SemanticVerification {
```
hinto-janai, line 140: 1/2

nit (I think `only` is a more used term in these contexts)
```suggestion
    /// Only semantic verification.
    ///
    /// Semantic verification is verification that can done without other blockchain data. The [`HardFork`]
    /// is technically other blockchain data but excluding it reduces the amount of things that can be checked
    /// significantly, and it is easy to get compared to other blockchain data needed for contextual validation.
    pub fn only_semantic(self, hf: HardFork) -> SemanticVerification {
```

types/src/transaction_verification_data.rs
```diff
@@ -37,6 +35,8 @@ impl TxVersion {
 pub enum CachedVerificationState {
     /// The transaction has not been validated.
     NotVerified,
+    /// The transaction was only validated semantically.
+    JustSemantic(HardFork),
```
hinto-janai, line 39: 2/2
```suggestion
    /// The transaction was only validated semantically.
    OnlySemantic(HardFork),
```

consensus/src/transactions/contextual_data.rs
```diff
@@ -254,3 +249,6 @@ pub async fn batch_get_decoy_info<'a, D: Database + Clone + Send + 'static>(
         .map_err(ConsensusError::Transaction)
     }))
 }
+
+pub trait Captures<U> {}
+impl<T: ?Sized, U> Captures<U> for T {}
```
hinto-janai, line 254:
```suggestion
mod sealed {
    /// TODO: Remove me when 2024 Rust
    ///
    /// <https://rust-lang.github.io/rfcs/3498-lifetime-capture-rules-2024.html#the-captures-trick>
    pub trait Captures<U> {}
    impl<T: ?Sized, U> Captures<U> for T {}
}
```

binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -83,30 +86,33 @@ impl super::BlockchainManager {
         block: Block,
         prepared_txs: HashMap<[u8; 32], TransactionVerificationData>,
     ) -> Result<IncomingBlockOk, anyhow::Error> {
-        if block.header.previous != self.cached_blockchain_context.top_hash {
+        if block.header.previous
```
hinto-janai, line 89: Should this function be like this to prevent ABA? `verify_main_chain_block` still calls `blockchain_context()` internal too.

```rust
pub async fn handle_incoming_block(
    &mut self,
    block: Block,
    prepared_txs: HashMap<[u8; 32], TransactionVerificationData>,
) -> Result<IncomingBlockOk, anyhow::Error> {
    let blockchain_context = self.blockchain_context_service.blockchain_context();

    if block.header.previous != blockchain_context {
        self.handle_incoming_alt_block(block, prepared_txs).await?;
        return Ok(IncomingBlockOk::AddedToAltChain);
    }

    let verified_block = verify_main_chain_block(
        block,
        prepared_txs,
        &mut self.blockchain_context_service,
        self.blockchain_read_handle.clone(),
    )
    .await?;

    let block_blob = Bytes::copy_from_slice(&verified_block.block_blob);
    self.add_valid_block_to_main_chain(verified_block).await;

    self.broadcast_block(block_blob, blockchain_context.chain_height)
        .await;

    Ok(IncomingBlockOk::AddedToMainChain)
}
```

binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -272,25 +267,18 @@ impl super::BlockchainManager {
...
-            unreachable!();
-        };
+        let alt_block_info =
+            sanity_check_alt_block(block, prepared_txs, self.blockchain_context_service.clone())
+                .await?;
 
         // TODO: check in consensus crate if alt block with this hash already exists.
 
-        // If this alt chain
+        // If this alt chain has more cumulative difficulty reorg.
```
hinto-janai, line 276:
```suggestion
        // If this alt chain has more cumulative difficulty, reorg.
```

binaries/cuprated/src/blockchain/syncer.rs
```diff
@@ -114,7 +97,7 @@ where
 
 /// Returns `true` if we are behind the current connected network peers.
 async fn check_behind_peers(
-    raw_blockchain_context: &RawBlockChainContext,
+    raw_blockchain_context: &BlockchainContext,
```
hinto-janai, line 100:
```suggestion
    blockchain_context: &BlockchainContext,
```
Also at:
```
binaries/cuprated/src/blockchain/syncer.rs
116:    if cumulative_difficulty <= raw_blockchain_context.cumulative_difficulty {
```

consensus/src/block/free.rs
Boog900: Damn nice find, should have added a test. FWIW peers send txs in order which is why this worked fine, obviously we shouldn't rely on peers doing that though.

The error is so subtle:

```suggestion
    for (i, tx_hash) in block.transactions.iter().enumerate() {
        if tx_hash != &txs[i].tx_hash {
            let at_index = txs[i..]
                .iter()
                .position(|tx| &tx.tx_hash == tx_hash)
                .ok_or(ExtendedConsensusError::TxsIncludedWithBlockIncorrect)?;

            // The above `position` will give an index from inside its view of the slice so we need to add the difference.
            txs.swap(i, (i + at_index);
        }
    }
```

Boog900: FWIW the thing we are swapping are `Transaction`s which would be expensive to clone

binaries/cuprated/src/blockchain/manager/handler.rs
Boog900: This is the only place that updates the context so no ABA. Also `add_valid_block_to_main_chain` updates the context so we would need to fetch again.

hinto-janai (approved):

PR #367, opened 2025-01-17
lints: enable `wildcard_enum_match_arm`

hinto-janai:
### What
Enables the ["hot"](https://github.com/Cuprate/cuprate/issues/131) `wildcard_enum_match_arm` lint and applies fixes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -346,7 +346,7 @@ allow_attributes = "deny"
 undocumented_unsafe_blocks = "deny"
 # multiple_unsafe_ops_per_block = "deny"
 # single_char_lifetime_names = "deny"
-# wildcard_enum_match_arm = "deny"
+wildcard_enum_match_arm = "deny"
 
 [workspace.lints.rust]
 # Cold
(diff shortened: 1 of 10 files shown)


hinto-janai:
This lint is less about formality, more so about moving responsibility away from humans. I am not a fan of the verbosity sometimes but I am okay with it if this lint is enforced on all current and future `match`s by all contributors.

In edge cases there is `#[expect]`.

p2p/p2p-transport/src/arti.rs
```diff
@@ -164,7 +164,10 @@ impl Transport<Tor> for Arti {
 
                     Ok(stream.split())
                 }
-                _ => {
+                IncomingStreamRequest::Begin(_)
+                | IncomingStreamRequest::BeginDir(_)
+                | IncomingStreamRequest::Resolve(_)
+                | _ => {
```
Boog900, line 170: this one should be ignored with an `expect` IMO

Boog900 (changes requested):

Boog900 (approved):

PR #384, opened 2025-02-22
Consensus: Fix reorgs

Boog900:
### What
Fixes blockchain re-orgs.


consensus/context/src/alt_chains.rs
```diff
@@ -119,7 +115,7 @@ impl AltChainMap {
             weight_cache: None,
             difficulty_cache: None,
             cached_rx_vm: None,
-            chain_height: top_height,
+            chain_height: top_height + 1,
```
Boog900, line 118: off by 1 here

consensus/src/block/alt_block.rs
```diff
@@ -178,7 +178,6 @@ where
     // Add this alt cache back to the context service.
     context_svc
         .oneshot(BlockChainContextRequest::AddAltChainContextCache {
-            prev_id: block_info.block.header.previous,
```
Boog900, line 181: the `prev_id` should be the current block hash here otherwise the block wont be able to find the cache.

changed to just set this in the alt block cache rather than requiring people to us the correct value

consensus/context/src/alt_chains.rs
```diff
@@ -83,12 +83,8 @@ impl AltChainMap {
...
 
     /// Add an alt chain cache to the map.
-    pub(crate) fn add_alt_cache(
-        &mut self,
-        prev_id: [u8; 32],
-        alt_cache: Box<AltChainContextCache>,
-    ) {
-        self.alt_cache_map.insert(prev_id, alt_cache);
+    pub(crate) fn add_alt_cache(&mut self, alt_cache: Box<AltChainContextCache>) {
+        self.alt_cache_map.insert(alt_cache.top_hash, alt_cache);
```
Boog900, line 87: correct value now used here

Boog900 (commented):
Some small fixes for alt block handling

hinto-janai (approved):

PR #386, opened 2025-02-25
Update to experimental monero oxide api

Boog900:
### What

Updates to the monero-oxide API I have been working on, which keeps all points in the tx compressed.

### Why

(de)compressions are very expensive, this reduces the number of them needed.


binaries/cuprated/config/0.0.1.toml
```diff
@@ -43,9 +43,9 @@ peer_save_period = { secs = 90, nanos = 0 }
 ## The block downloader config.
 [p2p.block_downloader]
 ## The size of the buffer of sequential blocks waiting to be verified and added to the chain (bytes).
-buffer_bytes = 50_000_000
+buffer_bytes = 1_000_000_000
 ## The size of the queue of blocks which are waiting for a parent block to be downloaded (bytes).
-in_progress_queue_bytes = 50_000_000
+in_progress_queue_bytes = 500_000_000
```
Boog900, line 48: during testing I found it was better to keep `buffer_bytes` at least 2x `in_progress_queue_bytes`.

binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -271,12 +271,28 @@ impl super::BlockchainManager {
...
+            .expect(PANIC_CRITICAL_SERVICE_ERROR)
+        else {
+            unreachable!();
+        };
+
+        if chain.is_some() {
+            // The block could also be in the main-chain here under some circumstances.
+            return Ok(AddAltBlock::Cached);
+        }
+
```
Boog900, line 291: duplicate changes from #384, ignore here.

binaries/cuprated/src/blockchain/syncer.rs
```diff
@@ -3,7 +3,7 @@ use std::{sync::Arc, time::Duration};
 
 use futures::StreamExt;
 use tokio::{
-    sync::{mpsc, Notify},
+    sync::{mpsc, oneshot, Notify},
```
Boog900, line 6: oops, will be fixed when the allows are removed.

consensus/context/src/alt_chains.rs
Boog900, line 1: same here, duplicate changes from #384

types/src/output_cache.rs
```diff
@@ -87,8 +87,7 @@ impl OutputCache {
                         OutputOnChain {
                             height,
                             time_lock: tx.prefix().additional_timelock,
-                            // TODO: this needs to check the point is canonical.
-                            key: out.key.decompress(),
+                            key: out.key,
```
Boog900, line 90: this is now checked in the verify functions

binaries/cuprated/config/0.0.1.toml
hinto-janai: Perhaps `in_progress_queue_bytes` shouldn't be exposed to users and/or `buffer_bytes` should be calculated like such?:

```rust
let buffer_bytes = min(in_progress_queue_bytes * 2, buffer_bytes);
```

hinto-janai (approved):

binaries/cuprated/config/0.0.1.toml
Boog900: For now we should keep it as is but will probably be something to look at as well as possibly making it dynamic.

PR #389, opened 2025-03-02
cuprated: Add fast sync

### LINKED ISSUE #373: Unstable sync towards the top of the chain.
## Bug
When you get close to the top of the chain the syncer often exits early, this isn't a problem as it'll be restarted but it does get shown in the logs and probably means we are downloading some of the top blocks multiple times.

## Expected behavior
We should be able to sync without needing to restart the syncer as long as all peers give us valid data.

## Steps to reproduce
Just start `cuprated` and wait to get near the top then look at the logs.

Boog900:
### What

fixes #373 

Adds fast sync


binaries/cuprated/src/blockchain/fast_sync.rs
```diff
@@ -0,0 +1,22 @@
...
+};
+
+/// Set the fast-sync hashes according to the provided values.
+pub fn set_fast_sync_hashes(fast_sync: bool, network: Network) {
+    cuprate_fast_sync::set_fast_sync_hashes(if fast_sync && network == Network::Mainnet {
+        FAST_SYNC_HASHES
+    } else {
+        &[]
+    });
+}
```
Boog900, line 22: This is the function where the hashes for fast sync are chosen - only mainnet uses fast sync (same as monerod)

binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -195,7 +201,32 @@ impl super::BlockchainManager {
...
+        info!(fast_sync = false, "Successfully added block batch");
+    }
+
+    /// Handles an incoming block batch while we are under the fast sync height.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if any internal service returns an unexpected error that we cannot
+    /// recover from.
+    async fn handle_incoming_block_batch_fast_sync(&mut self, batch: BlockBatch) {
```
Boog900, line 213: the new function to handle an incoming fast sync batch, the block downloader guarantees the blocks downloaded where the ones asked for so no checking needs to be done on the blocks here

binaries/cuprated/src/blockchain/syncer.rs
```diff
@@ -51,6 +54,9 @@ where
 
     tracing::debug!("Waiting for new sync info in top sync channel");
 
+    let semaphore = Arc::new(Semaphore::new(1));
+
+    let mut sync_permit = Arc::new(semaphore.clone().acquire_owned().await.unwrap());
```
Boog900, line 59: This permit requires we have fully handled the blocks of a previous download attempt before starting another, hopefully fixing #373

consensus/fast-sync/src/fast_sync.rs
```diff
@@ -1,225 +1,213 @@
...
+///
+/// *once we are passed the fast sync blocks all entries will be returned as valid as
+/// we can not check their validity here.
+///
+/// There may be more entries returned than passed in as entries could be split.
+///
+/// # Panics
+///
+/// This will panic if [`set_fast_sync_hashes`] has not been called.
+pub async fn validate_entries<N: NetworkZone>(
```
Boog900, line 64: This is the main block entries validation function, because we need to keep block hashes in their respective batch (we can split batches but we can't join) it makes this pretty complicated. I have tried to compensate with comments

p2p/p2p/src/block_downloader.rs
```diff
@@ -280,6 +283,28 @@ where
...
+        let Some(biggest_batch) = biggest_batch else {
+            return self.config.initial_batch_len;
+        };
+
+        calculate_next_block_batch_size(
+            biggest_batch.0 .1.byte_size,
+            biggest_batch.0 .1.len,
+            self.config.target_batch_bytes,
+        )
+    }
```
Boog900, line 306: I have made the amount of blocks requested based of the biggest of the 100 most recent batches we have seen this makes the block downloader a lot more stable when blocks get added fast.

Boog900 (commented):
for the fast sync file, I wouldn't look at the diff just look at the new file

binaries/cuprated/src/blockchain/fast_sync/fast_sync_hashes.bin
Boog900, line 1: if you want to recreate this file you will need to use the create hashes binary to create hashes up to `3349504`

SyntheticBird45 (approved):

binaries/cuprated/src/commands.rs
```diff
@@ -47,6 +47,9 @@ pub enum Command {
 
     /// Print status information on `cuprated`.
     Status,
+
+    /// Print the height of last block contained in the fast sync hashes.
+    FastSyncTopHeight,
```
hinto-janai, line 52: Is this required as a 24/7 available command? Putting it in `--help` data seems more appropriate.

binaries/cuprated/src/blockchain/fast_sync/fast_sync_hashes.bin
hinto-janai: TODO: this process is not clear or easily reproducible/verifiable by others, it should be documented. I'll try to do this later.

- What does this file contain
- How to create this file
- Where does it get created

hinto-janai: TODO: also I will verify this before release, current sha256 hash: `c37c12479223c22b6122d857b154a4da1b972417ddf1464e2f358f615dc2aea6`

binaries/cuprated/src/blockchain/fast_sync.rs
```diff
@@ -0,0 +1,24 @@
...
+static FAST_SYNC_HASHES: &[[u8; 32]] = unsafe {
+    let bytes = include_bytes!("./fast_sync/fast_sync_hashes.bin");
+
+    #[expect(clippy::manual_assert, reason = "assert is not const")]
+    if bytes.len() % 32 != 0 {
+        panic!()
+    }
+
+    slice::from_raw_parts(bytes.as_ptr().cast::<[u8; 32]>(), bytes.len() / 32)
+};
```
hinto-janai, line 15: `bytemuck::cast_slice` almost certainly works here although unfortunately the const `bytemuck::must_cast_slice` seems to trip up.

```suggestion
/// The hashes of the compiled in fast sync file.
static FAST_SYNC_HASHES: &[[u8; 32]] = {
    let bytes = include_bytes!("./fast_sync/fast_sync_hashes.bin");

    if bytes.len() % 32 == 0 {
        // SAFETY: The file byte length must be perfectly divisible by 32, checked above.
        unsafe { slice::from_raw_parts(bytes.as_ptr().cast::<[u8; 32]>(), bytes.len() / 32) }
    } else {
        panic!();
    }
};
```

p2p/p2p/src/block_downloader.rs
```diff
@@ -280,6 +283,28 @@ where
         }
     }
 
+    fn amount_of_blocks_to_request(&self) -> usize {
+        let biggest_batch = self
+            .most_recent_batch_sizes
+            .iter()
+            .max_by(|batch_1, batch_2| {
+                let av1 = batch_1.0 .1.byte_size / batch_1.0 .1.len;
+                let av2 = batch_2.0 .1.byte_size / batch_2.0 .1.len;
```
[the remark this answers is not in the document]
Boog900: yes

consensus/fast-sync/src/fast_sync.rs
```diff
@@ -243,68 +232,16 @@ fn validate_block(
     let weight = block.miner_transaction.weight()
         + verified_txs.iter().map(|tx| tx.tx_weight).sum::<usize>();
 
-    Ok(FastSyncResponse::ValidateBlock(VerifiedBlockInformation {
+    VerifiedBlockInformation {
         block_blob,
         txs: verified_txs,
         block_hash,
-        pow_hash: [0_u8; 32],
+        pow_hash: [u8::MAX; 32],
```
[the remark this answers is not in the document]
Boog900: `[0; 32]` is always a valid PoW hash, `[u8::MAX; 32]` is only correct when the difficulty is at the minimum level. IMO it's better to default to the value that will almost never pass the PoW check.

binaries/cuprated/src/commands.rs
Boog900: I was going to change this but we set the fast sync hashes _after_ we read and process the config so I think it has to stay like this.

I don't think it's too bad.

hinto-janai (approved):

PR #393, opened 2025-03-05
`cuprated`: add startup delay if config is missing

Boog900:
### What
Adds a startup delay with warning if a user does not have a config.

### Why
Not customizing a config for the specific hardware can effect sync performance quite a bit/cause OOMs on weak hardware.

### DIFF
--- a/binaries/cuprated/src/constants.rs
+++ b/binaries/cuprated/src/constants.rs
@@ -1,4 +1,5 @@
 //! General constants used throughout `cuprated`.
+use std::time::Duration;
 
 use const_format::formatcp;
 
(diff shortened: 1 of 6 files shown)


SyntheticBird45:
Can there be a cli flag to skip that startup delay ?

Boog900:
@SyntheticBird45 you could just create a config file?

SyntheticBird45:
> @SyntheticBird45 you could just create a config file?

I mean, right now it's ok. But having a config file written somewhere shouldn't be mandatory later on.

binaries/cuprated/src/constants.rs
```diff
@@ -24,7 +24,11 @@ pub const VERSION_BUILD: &str = formatcp!("{VERSION}-{}", cuprate_constants::bui
 pub const PANIC_CRITICAL_SERVICE_ERROR: &str =
     "A service critical to Cuprate's function returned an unexpected error.";
 
-pub const DEFAULT_CONFIG_WARNING: &str = formatcp!( "WARNING: no config file found, using default config.\nThe default config may not be optimal for your hardware, see the user book here: <TODO>.\nPausing startup for {} seconds.",
+pub const DEFAULT_CONFIG_WARNING: &str = formatcp!(
+    "WARNING: no config file found, using default config.\
+    \nThe default config may not be optimal for your setup, see the user book here: https://user.cuprate.org/.\
+    \nPausing startup for {} seconds. \
+    \nUse the `--skip-config-warning` arg to skip this delay if you relly want to use the default.",
```
SyntheticBird45:
```suggestion
    \nUse the `--skip-config-warning` arg to skip this delay if you really want to use the default.",
```

SyntheticBird45 (approved):

PR #394, opened 2025-03-06
cuprated: config versioning + backwards compat test

hinto-janai:
### What
Creates a versioning scheme for `cuprated` config files and adds a test for backwards compatibility.

### Why
To test/support older configs, for now everything could be breaking although we may as well set this up now.


Boog900 (approved):
IMO we would probably want separate config files with every option specifed as the test cases, but this is ok for now

PR #410, opened 2025-03-19
fast-sync: add tests

Boog900:
### What
Adds some tests to fast sync

### DIFF
--- a/consensus/fast-sync/src/fast_sync.rs
+++ b/consensus/fast-sync/src/fast_sync.rs
@@ -98,6 +98,24 @@ pub async fn validate_entries<N: NetworkZone>(
 
     let mut hashes_stop_diff_last_height = last_height - hashes_stop_height;
 
+    // get the hashes we are missing to create the first fast-sync hash.
+    let BlockchainResponse::BlockHashInRange(starting_hashes) = blockchain_read_handle
+        .ready()
+        .await?
+        .call(BlockchainReadRequest::BlockHashInRange(
+            hashes_start_height..start_height,
+            Chain::Main,
+        ))
+        .await?
+    else {
+        unreachable!()
+    };
+
+    // If we don't have enough hashes to make up a batch we can't validate any.
+    if amount_of_hashes + starting_hashes.len() < FAST_SYNC_BATCH_LEN {
+        return Ok((VecDeque::new(), entries));
+    }
+
     let mut unknown = VecDeque::new();
 
     // start moving from the back of the batches taking enough hashes out so we are only left with hashes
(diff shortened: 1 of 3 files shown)


consensus/fast-sync/src/fast_sync.rs
```diff
@@ -98,6 +98,24 @@ pub async fn validate_entries<N: NetworkZone>(
...
+        .ready()
+        .await?
+        .call(BlockchainReadRequest::BlockHashInRange(
+            hashes_start_height..start_height,
+            Chain::Main,
+        ))
+        .await?
+    else {
+        unreachable!()
+    };
```
SyntheticBird45, line 112: I probably misunderstood something, but if we're in the range of fast-sync, how can we have blocks validated and committed within `BlockChain`, that are in the range of a batch of block we're trying to verify? We wouldn't commit it to the database if they weren't verified, but to verify it we need the whole batch, but if have the whole batch then why are we verifying these blocks?

consensus/fast-sync/src/fast_sync.rs
```diff
@@ -245,3 +250,148 @@ pub fn block_to_verified_block_information(
...
+            .collect::<Vec<_>>();
+
+        set_fast_sync_hashes(fast_sync_hashes.leak());
+
+        hashes
+    });
+
+    proptest! {
+        #[test]
+        fn valid_entry(len in 0_usize..1_500_000) {
```
SyntheticBird45, line 298: The `HASHES` lazy lock is a slice of length 1_024_000, we can probably reduce 1_500_000 to 1_100_000

consensus/fast-sync/src/fast_sync.rs
```diff
@@ -245,3 +250,148 @@ pub fn block_to_verified_block_information(
...
+                let len_right = ret.1.iter().map(|e| e.ids.len()).sum::<usize>();
+
+                assert_eq!(len_left + len_right, len);
+                assert!(len_left <= fast_sync_stop_height());
+                assert!(len_right < FAST_SYNC_BATCH_LEN || len > fast_sync_stop_height());
+            });
+        }
+
+        #[test]
+        fn single_hash_entries(len in 0_usize..1_500_000) {
```
SyntheticBird45, line 333: same here

consensus/fast-sync/src/fast_sync.rs
Boog900: The first fast sync hash uses blocks 0..512. If an incoming batch starts at block 200 we need the 199 hashes before it to make up the hash

Boog900: This could happen if we shutdown while syncing, or we start syncing part of the chain without fast sync

SyntheticBird45 (approved):

PR #416, opened 2025-03-21
cuprated: add `exit` command

hinto-janai:
### What
Adds an `exit` command to `cuprated`.


Boog900:
I am unsure if this should be added right now - this is still a hard exit right?

We need to add proper shutdown code first IMO

PR #420, opened 2025-03-25
cuprated: add system paths to `--version`

hinto-janai:
### What
Adds system {data,cache,config} directory info to `./cuprated --version`.

### Why
Useful info for debugging / sanity checking / users.

### DIFF
--- a/binaries/cuprated/src/version.rs
+++ b/binaries/cuprated/src/version.rs
@@ -1,10 +1,11 @@
 //! Misc version information about `cuprated`.
 
-use std::fmt::Display;
+use std::{fmt::Display, path::PathBuf};
 
 use serde::{Deserialize, Serialize};
 
 use cuprate_constants::build::{BUILD, COMMIT};
+use cuprate_helper::fs::{CUPRATE_CACHE_DIR, CUPRATE_CONFIG_DIR, CUPRATE_DATA_DIR};
 use cuprate_rpc_types::{CORE_RPC_VERSION, CORE_RPC_VERSION_MAJOR, CORE_RPC_VERSION_MINOR};
 use cuprate_types::HardFork;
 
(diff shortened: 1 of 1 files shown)


Boog900:
IMO this should take into account the config, currently it just using the default, right?

hinto-janai:
> IMO this should take into account the config

That means this line:

https://github.com/Cuprate/cuprate/blob/57cd96ed6c75cb7d16ec8156bc66f9e72caf9ad9/binaries/cuprated/src/config.rs#L43

has to be before:
https://github.com/Cuprate/cuprate/blob/57cd96ed6c75cb7d16ec8156bc66f9e72caf9ad9/binaries/cuprated/src/config.rs#L45

That means something like `./cuprated --help` may not work depending on if config reading fails. Kind of a hacky workaround:

```rust
pub fn read_config_and_args() -> Config {
    let args = args::Args::parse();

    let config: Config = try_get_config().map(|c| args.apply_args(c));

    args.do_quick_requests(&config); // if Ok(config), use in `*_directory`, else ""

    match config {
        Ok(c) => c,
        Err(e) => {
            tracing::debug!("Failed to read config: {e}");
            if !args.skip_config_warning {
                eprintln_red(DEFAULT_CONFIG_WARNING);
                std::thread::sleep(DEFAULT_CONFIG_STARTUP_DELAY);
            }
        }
    }
}
```

> currently it just using the default, right?

Yes, should the fields be renamed to `*_default_directory`? I agree that displaying the directories `cuprated` will actually use is helpful, as well as the defaults.

Boog900:
Hmm, I don't really want to move that to after reading the config. I also don't know how useful just giving the default out is - it is just a constant. 

the RPC exposes this right? isn't that enough

PR #424, opened 2025-04-02
`cuprated`: relay rules

Boog900:
# What

Adds tx-pool relay rules to `cuprated`.

### DIFF
--- a/binaries/cuprated/src/txpool/relay_rules.rs
+++ b/binaries/cuprated/src/txpool/relay_rules.rs
@@ -9,16 +9,20 @@ use cuprate_helper::cast::usize_to_u64;
 use cuprate_types::TransactionVerificationData;
 
 /// The maximum size of the tx extra field.
+///
+/// <https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_config.h#L217>
 const MAX_TX_EXTRA_SIZE: usize = 1060;
 
+/// <https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_config.h#L75>
 const DYNAMIC_FEE_REFERENCE_TRANSACTION_WEIGHT: u128 = 3_000;
 
+/// <https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_core/blockchain.h#L646>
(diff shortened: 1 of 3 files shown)


binaries/cuprated/src/txpool/relay_rules.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[error("Tx extra field is too large.")]
+    ExtraFieldTooLarge,
+    #[error("Tx fee too low.")]
+    FeeBelowMinimum,
+}
+
+/// Checks the transaction passes the relay rules.
+///
+/// Relay rules are rules that govern the txs we accept to our tx-pool and propagate around the network.
+pub fn check_tx_relay_rules(
```
hinto-janai, line 31: Some questions:

- There seem to be other relay rules not included here, e.g.: https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_core/tx_pool.cpp#L150-L151, should the diff between `monerod` and `cuprated` relay rule behavior be documented somewhere?
- Should `check_fee/dynamic_base_fee` have some tests to make sure it matches `monerod` over time?

Boog900: > There seem to be other relay rules not included here, e.g.: https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_core/tx_pool.cpp#L150-L151, should the diff between monerod and cuprated relay rule behavior be documented somewhere?

That one isn't really a relay rule IMO, it just prevents a previously timed out tx from getting added to the pool again.

> Should check_fee/dynamic_base_fee have some tests to make sure it matches monerod over time?

Probably, I don't know how we would do that without generating txs though so that will have to come later

hinto-janai: There is other code there: https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_core/tx_pool.cpp#L204-L344.

Should that be replicated here?

hinto-janai (approved):

binaries/cuprated/src/txpool/relay_rules.rs
Boog900: monerod's tx-pool also deals with some consensus checks, only the checks that have `m_no_drop_offense = true` are relay checks

PR #426, opened 2025-04-07
Use blank peer list if saved peer list cannot be read.

### LINKED ISSUE #399: `cuprated` peer list write is not atomic, prevents startup
## Bug
Peer list writes are not atomic:

https://github.com/Cuprate/cuprate/blob/21ad35d44a465efbb5414a902cb6c370b8358303/p2p/address-book/src/store.rs#L47

Peer list reads do not have a recovery method so if a un-readable file exists, it will prevent `cuprated` from starting due to:

https://github.com/Cuprate/cuprate/blob/21ad35d44a465efbb5414a902cb6c370b8358303/p2p/address-book/src/lib.rs#L70-L71

## Expected behavior
- Recovery method on the (de)serialization (e.g. skip + warn on read fail)
- Atomic writes (`write(tmp_file)` -> `rename(tmp_file, actual_file)` or database instead of binary file)

## Steps to reproduce
1. Start `cuprated`
2. Run out of disk space (and maybe have an unlucky peer list write at the same time?)
3. See error `ERROR Could not save peer list to disk, got error: No space left on device (os error 28)`
4. Cannot launch `cuprated`, blocked by: `ERROR Failed to open peer list, Unexpected length of input`

NorrinRadd:
address-book: fix panic when the peer list on the file system cannot be read

### What
Writes a blank peer list to a temporary location, then attempts to clean the old location, and save the new file as the old file.

Fixes #399 

Cuprate can still panic if deleting the old file fails. That error will be logged, to allow manual cleanup and the latest peer list will already be saved to disk as a .tmp file. The user can remove the .tmp file extension if they wish to use it.

### DIFF
--- a/p2p/address-book/src/store.rs
+++ b/p2p/address-book/src/store.rs
@@ -50,15 +50,15 @@ pub(crate) fn save_peers_to_disk<Z: BorshNetworkZone>(
             Ok(_) => {
                 let orig_file_path = file.clone();
                 match fs::remove_file(file) {
-                    Ok(_) => {},
+                    Ok(_) => {}
                     Err(x) => {
                         tracing::error!("{x}. Saved peer list to temp file instead.");
                         return Err(x);
                     }
                 }
                 fs::rename(tmp_file, orig_file_path)
-            },
-            Err(x) => Err(x)
+            }
+            Err(x) => Err(x),
         }
     })
 }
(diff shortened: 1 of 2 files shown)


p2p/address-book/src/store.rs
```diff
@@ -41,10 +41,25 @@ pub(crate) fn save_peers_to_disk<Z: BorshNetworkZone>(
...
+        match fs::write(&tmp_file, &data) {
+            Ok(_) => {
+                let orig_file_path = file.clone();
+                match fs::remove_file(file) {
+                    Ok(_) => {}
+                    Err(x) => {
+                        tracing::error!("{x}. Saved peer list to temp file instead.");
+                        return Err(x);
+                    }
+                }
```
Boog900, line 58: I don't think we need to delete the file first, the name change should overwrite it right?

Boog900 (changes requested):

Issue #435, opened 2025-04-09
Panic at `binaries/cuprated/src/blockchain/manager/handler.rs:170:32`


hinto-janai:
## Bug
This line (sometimes) panics when going from cuprated `0.0.1` -> `0.0.2`.

https://github.com/Cuprate/cuprate/blob/95aca1d4a52c56975ab9e859cfb283be8a7d4838/binaries/cuprated/src/blockchain/manager/handler.rs#L170

Restarting enough times seems to fix it.

## Steps to reproduce
1. Sync `cuprated 0.0.1` to chaintip
2. Run `cuprated 0.0.2`
3. Panic

## Log
macOS ARM64:
```
INFO Starting blockchain syncer
INFO block_downloader: Attempting to download blocks from peers, this may take a while.
INFO incoming_block_batch{start_height=3386273 len=1}:pop_blocks_diff_cache{numb_blocks=1}:get_blocks_timestamps{block_heights=3385538..3385539 chain=Main}: Getting blocks timestamps
INFO incoming_block_batch{start_height=3386273 len=1}:pop_blocks_weight_cache{numb_blocks=1}:get_long_term_weights{range=3286273..3286274 chain=Main}: getting block long term weights.
INFO incoming_block_batch{start_height=3386273 len=1}:pop_blocks_weight_cache{numb_blocks=1}:get_block_weights{range=3386173..3386174 chain=Main}: getting block weights.
INFO incoming_block_batch{start_height=3386273 len=1}: Successfully added block batch alt_chain=true
INFO incoming_block_batch{start_height=3386274 len=1}:pop_blocks_diff_cache{numb_blocks=1}:get_blocks_timestamps{block_heights=3385538..3385539 chain=Main}: Getting blocks timestamps
INFO incoming_block_batch{start_height=3386274 len=1}:pop_blocks_weight_cache{numb_blocks=1}:get_long_term_weights{range=3286273..3286274 chain=Main}: getting block long term weights.
INFO incoming_block_batch{start_height=3386274 len=1}:pop_blocks_weight_cache{numb_blocks=1}:get_block_weights{range=3386173..3386174 chain=Main}: getting block weights.

thread 'cuprated-tokio' panicked at binaries/cuprated/src/blockchain/manager/handler.rs:170:32:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Abort trap: 6
```

Linux x64:
```
[...]

PR #438, opened 2025-04-09
`cuprated`: fix switching to main chain after a reorg

Boog900:
fixes: #435

Sad we couldn't get this before release but oh well.


binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -242,7 +242,6 @@ impl super::BlockchainManager {
     /// This function will panic if any internal service returns an unexpected error that we cannot
     /// recover from.
     async fn handle_incoming_block_batch_alt_chain(&mut self, mut batch: BlockBatch) {
-        // TODO: this needs testing (this whole section does but alt-blocks specifically).
```
Boog900, line 245: I think this is sufficiently tested now :)

hinto-janai (approved):

PR #458, opened 2025-05-07
fix epee bin OOM

Boog900:


net/epee-encoding/src/lib.rs
```diff
@@ -372,3 +372,7 @@ fn skip_epee_value<B: Buf>(r: &mut B, skipped_objects: &mut u8) -> Result<()> {
 fn advance<B: Buf>(n: usize, b: &mut B) -> Result<()> {
     checked_read(b, |b: &mut B| b.advance(n), n)
 }
+
+const fn max_upfront_capacity<T>() -> usize {
+    2_000_000 / size_of::<T>()
+}
```
hinto-janai, line 378:
```suggestion
/// Max upfront capacity when deserializing `Vec<T>`.
///
/// <REASONING>
const fn max_upfront_capacity<T>() -> usize {
    2_000_000 / size_of::<T>()
}
```

Is there reasoning (that could be linked to) on where `2_000_000` comes from?

Boog900: this isn't a hard limit on capacity it just prevent allocating too much up front, this can be exceeded if there is actually this much data

hinto-janai (approved):

PR #461, opened 2025-05-07
cuprated --dry-run option

IznaDev:
### What
adding of command line option --dry-run 

### Why
to verify config file parameters : 
   Config deserialization
   PATH permissions
   If ports can be binded to

### Where
the modifications are in **config.rs** where we create:
 a **check_port()** function which check if the port is available or not
 a **dry_run_check()** function which look for the config file .toml or take the default config and check the deserialization, path permissions and availability of ports by calling **check_port()** function

and in **arg.rs** where we add the **dry_run** argument and call the **dry_run_check()** function

### DIFF
--- a/binaries/cuprated/src/config.rs
+++ b/binaries/cuprated/src/config.rs
@@ -2,11 +2,13 @@
 use std::{
     fs::{read_to_string, File},
     io,
+    net::{IpAddr, TcpListener},
     path::Path,
     str::FromStr,
     time::Duration,
 };
 
+use args::Args;
 use clap::Parser;
 use serde::{Deserialize, Serialize};
 
@@ -61,7 +63,7 @@ const HEADER: &str = r"##     ____                      _
 
 /// Reads the args & config file, returning a [`Config`].
 pub fn read_config_and_args() -> Config {
-    let args = args::Args::parse();
+    let args = Args::parse();
     args.do_quick_requests();
 
     let config: Config = if let Some(config_file) = &args.config_file {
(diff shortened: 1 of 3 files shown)


SyntheticBird45 (approved):
No feedbacks. LGTM

binaries/cuprated/src/config.rs
```diff
@@ -260,6 +262,62 @@ impl Config {
...
     }
+
+    fn check_port(ip: IpAddr, port: u16) -> Result<(), String> {
+        match TcpListener::bind((ip, port)) {
+            Ok(_) => Ok(()),
+            Err(e) => Err(format!("Fail to bind {ip}:{port} : {e}")),
+        }
+    }
+
+    pub fn dry_run_check(args: &Args) {
```
Boog900, line 273: I think to reduce duplicated code it would be better to call this function in  `read_config_and_args` after the config is loaded there and not in `do_quick_requests` where it currently is

binaries/cuprated/src/config.rs
```diff
@@ -2,11 +2,13 @@
 use std::{
     fs::{read_to_string, File},
     io,
+    net::{IpAddr, TcpListener},
     path::Path,
     str::FromStr,
     time::Duration,
 };
 
+use args::Args;
```
Boog900, line 11: nit: because the is a mod and not an external crate it should be grouped with the mod imports around line 40

Boog900 (commented):
Thanks for working on this!

PR #464, opened 2025-05-09
Add Fuzzing + fix found issues

Boog900:
### What
Added fuzz tests using `cargo-fuzz`


fuzz/Cargo.toml
```diff
@@ -0,0 +1,63 @@
+[package]
+name = "cuprate-fuzz"
```
Boog900, line 2: the new fuzz crate

net/epee-encoding/src/lib.rs
```diff
@@ -154,7 +154,7 @@ fn read_head_object<T: EpeeObject, B: Buf>(r: &mut B) -> Result<T> {
 }
 
 fn read_field_name_bytes<B: Buf>(r: &mut B) -> Result<Bytes> {
-    let len: usize = r.get_u8().into();
+    let len: usize = checked_read_primitive(r, Buf::get_u8)?.into();
```
Boog900, line 157: this is the only vuln fix

SyntheticBird45 (approved):
No feedback. Looks awesome

Cargo.toml
```diff
@@ -49,6 +49,59 @@ members = [
 	"helper",
 	"pruning",
 	"test-utils",
+
+	# Fuzz
+	"fuzz"
+]
+
+default-members = [
```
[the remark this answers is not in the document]
hinto-janai: Maybe add a `# <EXPLAINATION>` above `default-members` with this info?

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/cache@v4
+        with:
+          path: ${{ runner.tool_cache }}/cargo-fuzz
+          key: cargo-fuzz-bin-${{ env.CARGO_FUZZ_VERSION }}
+      - run: echo "${{ runner.tool_cache }}/cargo-fuzz/bin" >> $GITHUB_PATH
+      - run: cargo install --root "${{ runner.tool_cache }}/cargo-fuzz" --version ${{ env.CARGO_FUZZ_VERSION }} cargo-fuzz --locked
+
+      # Build and then run the fuzz target.
+      - run: cargo fuzz build ${{ matrix.fuzz_target }} -O
+      - run: cargo fuzz run ${{ matrix.fuzz_target }} -O -- -max_total_time=${{ env.FUZZ_TIME }}
```
hinto-janai, line 50: Is this line needed? `run` will build it.

```suggestion
      # Run the fuzz target.
      - run: cargo fuzz run ${{ matrix.fuzz_target }} -O -- -max_total_time=${{ env.FUZZ_TIME }}
```

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      # Upload fuzzing artifacts on failure for post-mortem debugging.
+      - uses: actions/upload-artifact@v4
+        if: failure()
```
hinto-janai, line 54: Is this correct? https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-of-failure
```suggestion
        if: ${{ failure() }}
```

fuzz/README.md
```diff
@@ -0,0 +1,26 @@
+ # Fuzz Tests
```
hinto-janai, line 1: Maybe document that these have to be updated when adding fuzz tests?:
1. fuzz/Cargo.toml
2. fuzz/fuzz_targets/*.rs
3. CI

.github/workflows/fuzz.yml
Boog900: This file was pretty much ripped from: https://rust-fuzz.github.io/book/cargo-fuzz/ci.html I guess it was like this because the timeout might include some build time :shrug:

.github/workflows/fuzz.yml
Boog900: same here as the other comment idk maybe your are correct but this was taken from the fuzz book

Cargo.toml
```diff
@@ -49,6 +49,64 @@ members = [
...
+
+	# Fuzz
+	"fuzz"
+]
+
+# Windows is a pain, so instead of making the fuzz tests work on windows, we just don't build the fuzz target on windows.
+# This is done by defining the `default-members` here to include everything but `fuzz`.
+# You can still build everything, including `fuzz` by using `--workspace`.
+#
+# Remeber to update this if you add any new crates!
```
hinto-janai, line 61:
```suggestion
# Remember to update this if you add any new crates!
```

hinto-janai (approved):

PR #446, opened 2025-04-29
P2P: Abstract TCP methods under the Transport Trait.

SyntheticBird45:
### What

Part of [ongoing Tor support development](https://ccs.getmonero.org/proposals/syntheticbird_cuprate_arti_2_months.html) for Cuprate

### Changelog

In workspace:
- Added a new `cuprate-p2p-transport` crate that will later welcome complex or crate dependent transport implementations. (Arti for example.)

In `cuprate-p2p-core`:
- Implemented a new Transport trait that abstract methods to listen for or connect to other peers over a NetworkZone Z.
- Reduced `NetworkZone` trait accordingly.
- Added a new `BROADCAST_OWN_ADDR` to `NetworkZone` for anonymized networks.
- Implemented `Tcp` transport for `ClearNet`.
- Implemented `Dummy` transport for placeholder/testing purposes
- Adapted definition of P2P components to the new `Transport` trait.
- Changed original `InBoundStream` implementation to support two different but optional listeners (IPv4/IPv6)

In `p2p-core/tests`:
- Implemented `FragTcp` for fragmented message testing

In `cuprated`:
- Added `p2p.enable_inbound_v6` and `listen_on_v6` config field.
- Added `From<&ClearNetConfig>` for `TransportConfig<ClearNet, Tcp>`.
- Abstracted original `start_clearnet_p2p` to `start_zone_p2p` with a given `NetworkZone` and `Transport`.
- Main P2P initialization has been moved to `cuprated::p2p` in `initialize_p2p` fn.
- Defined `NetworkInterfaces` collection (useless at the moment).

PR #481, opened 2025-05-15
Define Tor Zone, add onion addressing and more

SyntheticBird45:
### What

Part of [ongoing Tor support development](https://ccs.getmonero.org/proposals/syntheticbird_cuprate_arti_2_months.html) for Cuprate

Follows: https://github.com/Cuprate/cuprate/pull/446

### Changelog

In `cuprated`:

- Added `Z: NetworkZone` as a generic requirement for `address_book_config`. Now takes the optional node address in argument.
- Added `tor_net_seed_nodes` fn for obtaining tor seed nodes.
- Added `CrossNetworkInternalPeerId::Tor(_)` variant and `From<InternalPeerId<OnionAddr>>`

In `cuprate-wire`:

- Added `src/network_address/onion_addr.rs` implementing `OnionAddr` type used by `Tor` network zone.
- Implemented parsing, formatting, conversion and validation of `OnionAddr`.
- Implemented 2 validation tests and 2 parsing tests for `OnionAddr`.
- Documented and cleaned up `src/network_address/epee_builder.rs`.
- Changed `u8` `type` field of `TaggedNetworkAddress` to `AddressType` enum equivalent to monerod's `address_type`.
- Added additional `host` and `port` fields to `AllFieldedNetworkAddress` collection type.
- Added `NetworkAddress:Tor` variant and added conversion to related functions.

In `cuprate-address-book`:

- Added `our_own_addr: Z::Addr` field to AddressBookConfig. This adds a `Z: NetworkZone` requirement to `AddressBookConfig`.
- Adapted code to the new generic requirement.
- Implemented handling of new `AddressBookRequest::OwnAddress` for querying the node self specified address for the zone.

In `cuprate-p2p`:

- If `Z::BROADCAST_OUR_OWN_ADDR` = `true`, `handle_timed_sync_request` will insert the node's address to the peerlist being sent.

In `cuprate-p2p-core`:

- Removed `#[async_trait::async_trait]` attribute to `impl NetworkZone for *`.
- Added `AddressBookRequest::OwnAddress` and `AddressBookResponse::OwnAddress(Option<Z::Addr>)`.
- Defined new `Tor` `NetworkZone`

### DIFF
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -161,6 +161,7 @@ config_struct! {
         pub address_book_config: AddressBookConfig,
     }
 
+    /// The config values for P2P clear-net.
     #[derive(Debug, Deserialize, Serialize, PartialEq)]
     #[serde(deny_unknown_fields, default)]
     pub struct ClearNetConfig {
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -1,4 +1,5 @@
 use std::{
+    marker::PhantomData,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     path::Path,
     time::Duration,
@@ -10,8 +11,9 @@ use cuprate_helper::{fs::address_book_path, network::Network};
 use cuprate_p2p::config::TransportConfig;
 use cuprate_p2p_core::{
     transports::{Tcp, TcpServerConfig},
-    ClearNet, Transport,
+    ClearNet, NetworkZone, Transport,
 };
+use cuprate_wire::OnionAddr;
 
 use super::macros::config_struct;
 
@@ -228,16 +230,23 @@ config_struct! {
 
 impl SharedNetConfig {
     /// Returns the [`cuprate_address_book::AddressBookConfig`].
-    pub fn address_book_config(
+    pub fn address_book_config<Z: NetworkZone>(
         &self,
         cache_dir: &Path,
         network: Network,
-    ) -> cuprate_address_book::AddressBookConfig {
+        our_own_address: Option<Z::Addr>,
+    ) -> cuprate_address_book::AddressBookConfig<Z> {
+        assert!(
+            !Z::BROADCAST_OWN_ADDR && our_own_address.is_some(),
+            "This network DO NOT take an incoming address."
+        );
+
         cuprate_address_book::AddressBookConfig {
             max_white_list_length: self.address_book_config.max_white_list_length,
             max_gray_list_length: self.address_book_config.max_gray_list_length,
             peer_store_directory: address_book_path(cache_dir, network),
             peer_save_period: self.address_book_config.peer_save_period,
+            our_own_address,
         }
     }
 }
@@ -327,6 +336,28 @@ pub fn clear_net_seed_nodes(network: Network) -> Vec<SocketAddr> {
             "77.172.183.193:28080",
         ]
         .as_slice(),
+    };
+
+    seeds
+        .iter()
+        .map(|s| s.parse())
+        .collect::<Result<_, _>>()
+        .unwrap()
+}
+
+/// Seed nodes for `Tor`.
+pub fn tor_net_seed_nodes(network: Network) -> Vec<OnionAddr> {
+    let seeds = match network {
+        Network::Mainnet => [
+            "zbjkbsxc5munw3qusl7j2hpcmikhqocdf4pqhnhtpzw5nt5jrmofptid.onion:18083",
+            "qz43zul2x56jexzoqgkx2trzwcfnr6l3hbtfcfx54g4r3eahy3bssjyd.onion:18083",
+            "plowsof3t5hogddwabaeiyrno25efmzfxyro2vligremt7sxpsclfaid.onion:18083",
+            "plowsoffjexmxalw73tkjmf422gq6575fc7vicuu4javzn2ynnte6tyd.onion:18083",
+            "plowsofe6cleftfmk2raiw5h2x66atrik3nja4bfd3zrfa2hdlgworad.onion:18083",
+            "aclc4e2jhhtr44guufbnwk5bzwhaecinax4yip4wr4tjn27sjsfg6zqd.onion:18083",
+        ]
+        .as_slice(),
+        Network::Stagenet | Network::Testnet => [].as_slice(),
     };
 
     seeds
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -272,7 +272,7 @@ impl AddressBookConfig {
         cache_dir: &Path,
         network: Network,
         our_own_address: Option<Z::Addr>,
-    ) -> cuprate_address_book::AddressBookConfig::<Z> {
+    ) -> cuprate_address_book::AddressBookConfig<Z> {
         assert!(
             !Z::BROADCAST_OWN_ADDR && our_own_address.is_some(),
             "This network DO NOT take an incoming address."
--- a/net/wire/src/network_address.rs
+++ b/net/wire/src/network_address.rs
@@ -26,6 +26,9 @@ use cuprate_epee_encoding::EpeeObject;
 mod epee_builder;
 use epee_builder::*;
 
+mod onion_addr;
+pub use onion_addr::*;
+
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum NetZone {
     Public,
@@ -38,6 +41,7 @@ pub enum NetZone {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum NetworkAddress {
     Clear(SocketAddr),
+    Tor(OnionAddr),
 }
 
 impl EpeeObject for NetworkAddress {
@@ -56,6 +60,7 @@ impl NetworkAddress {
     pub const fn get_zone(&self) -> NetZone {
         match self {
             Self::Clear(_) => NetZone::Public,
+            Self::Tor(_) => NetZone::Tor,
         }
     }
 
@@ -72,6 +77,7 @@ impl NetworkAddress {
     pub const fn port(&self) -> u16 {
         match self {
             Self::Clear(ip) => ip.port(),
+            Self::Tor(addr) => addr.port(),
         }
     }
 }
@@ -106,7 +112,7 @@ impl TryFrom<NetworkAddress> for SocketAddr {
     fn try_from(value: NetworkAddress) -> Result<Self, Self::Error> {
         match value {
             NetworkAddress::Clear(addr) => Ok(addr),
-            //_ => Err(NetworkAddressIncorrectZone)
+            NetworkAddress::Tor(_) => Err(NetworkAddressIncorrectZone),
         }
     }
 }
--- a/net/wire/src/network_address/epee_builder.rs
+++ b/net/wire/src/network_address/epee_builder.rs
@@ -1,21 +1,87 @@
+//! Address epee serialization
+//!
+//! Addresses needs to be serialized into a specific format before being sent to other peers.
+//! This module is handling this particular construction.
+//!
+
+//---------------------------------------------------------------------------------------------------- Imports
+
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
 
 use bytes::Buf;
 use thiserror::Error;
 
 use cuprate_epee_encoding::{epee_object, EpeeObjectBuilder};
+use cuprate_types::AddressType;
 
 use crate::NetworkAddress;
 
+use super::OnionAddr;
+
+//---------------------------------------------------------------------------------------------------- Network address construction
+
 #[derive(Default)]
+/// There are no ordering guarantees in epee format and as such all potential fields can be collected during deserialization.
+/// The [`AllFieldsNetworkAddress`] is containing, as its name suggest, all optional field describing an address , as if it
+/// could be of any type.
+struct AllFieldsNetworkAddress {
+    /// IPv4 address
+    m_ip: Option<u32>,
+    /// IP port field
+    m_port: Option<u16>,
+
+    /// IPv6 address
+    addr: Option<[u8; 16]>,
+
+    /// Alternative network domain name (<domain>.onion or <domain>.i2p)
+    host: Option<String>,
+    /// Alternative network virtual port
+    port: Option<u16>,
+}
+
+epee_object!(
+    AllFieldsNetworkAddress,
+    m_ip: Option<u32>,
+    m_port: Option<u16>,
+    addr: Option<[u8; 16]>,
+    host: Option<String>,
+    port: Option<u16>,
+);
+
+impl AllFieldsNetworkAddress {
+    fn try_into_network_address(self, ty: AddressType) -> Option<NetworkAddress> {
+        Some(match ty {
+            AddressType::Ipv4 => NetworkAddress::from(SocketAddrV4::new(
+                Ipv4Addr::from(self.m_ip?.to_le_bytes()),
+                self.m_port?,
+            )),
+            AddressType::Ipv6 => NetworkAddress::from(SocketAddrV6::new(
+                Ipv6Addr::from(self.addr?),
+                self.m_port?,
+                0,
+                0,
+            )),
+            AddressType::Tor => {
+                NetworkAddress::from(OnionAddr::new(self.host?.as_str(), self.port?).ok()?)
+            }
+            // Invalid
+            _ => return None,
+        })
+    }
+}
+
+#[derive(Default)]
+/// A serialized network address being communicated to or from a peer.
 pub struct TaggedNetworkAddress {
-    ty: Option<u8>,
+    /// Type of the network address (used later for conversion)
+    ty: Option<AddressType>,
+    /// All possible fields for a network address
     addr: Option<AllFieldsNetworkAddress>,
 }
 
 epee_object!(
     TaggedNetworkAddress,
-    ty("type"): Option<u8>,
+    ty("type"): Option<AddressType>,
     addr: Option<AllFieldsNetworkAddress>,
 );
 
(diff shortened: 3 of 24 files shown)


binaries/cuprated/src/config/p2p.rs
```diff
@@ -335,3 +353,25 @@ pub fn clear_net_seed_nodes(network: Network) -> Vec<SocketAddr> {
...
+pub fn tor_net_seed_nodes(network: Network) -> Vec<OnionAddr> {
+    let seeds = match network {
+        Network::Mainnet => [
+            "zbjkbsxc5munw3qusl7j2hpcmikhqocdf4pqhnhtpzw5nt5jrmofptid.onion:18083",
+            "qz43zul2x56jexzoqgkx2trzwcfnr6l3hbtfcfx54g4r3eahy3bssjyd.onion:18083",
+            "plowsof3t5hogddwabaeiyrno25efmzfxyro2vligremt7sxpsclfaid.onion:18083",
+            "plowsoffjexmxalw73tkjmf422gq6575fc7vicuu4javzn2ynnte6tyd.onion:18083",
+            "plowsofe6cleftfmk2raiw5h2x66atrik3nja4bfd3zrfa2hdlgworad.onion:18083",
+            "aclc4e2jhhtr44guufbnwk5bzwhaecinax4yip4wr4tjn27sjsfg6zqd.onion:18083",
+        ]
```
SyntheticBird45, line 367: https://github.com/monero-project/monero/blob/master/src/p2p/net_node.inl#L888-L895

net/wire/src/network_address.rs
```diff
@@ -106,7 +112,7 @@ impl TryFrom<NetworkAddress> for SocketAddr {
     fn try_from(value: NetworkAddress) -> Result<Self, Self::Error> {
         match value {
             NetworkAddress::Clear(addr) => Ok(addr),
-            //_ => Err(NetworkAddressIncorrectZone)
+            NetworkAddress::Tor(_) => Err(NetworkAddressIncorrectZone),
```
SyntheticBird45, line 115: just in case putting `_ =>` here causes a clippy error because `Tor` is the only variant remaining.

net/wire/src/network_address/epee_builder.rs
```diff
@@ -1,21 +1,116 @@
+//! Address epee serialization
+//!
+//! Addresses needs to be serialized into a specific format before being sent to other peers.
+//! This module is handling this particular construction.
+//!
+
```
SyntheticBird45, line 6: Did some cleanup here on the way. `cuprate-wire` is pretty important but didn't received much love.

net/wire/src/network_address/epee_builder.rs
```diff
@@ -1,21 +1,116 @@
...
+    /// IP port field
+    m_port: Option<u16>,
+
+    /// IPv6 address
+    addr: Option<[u8; 16]>,
+
+    /// Alternative network domain name (<domain>.onion or <domain>.i2p)
+    host: Option<String>,
+    /// Alternative network virtual port
+    port: Option<u16>,
```
SyntheticBird45, line 90: Two new fields collected for alternative networks.

net/wire/src/network_address/epee_builder.rs
```diff
@@ -1,21 +1,116 @@
...
+
+// Tag indicating the address type.
+// <https://github.com/monero-project/monero/blob/341771ac3e65f88a6e78ca805f8dde2ae6ba7924/contrib/epee/include/net/enums.h#L39>
+#[repr(u8)]
+enum NetworkAddressTag {
+    IPv4 = 1,
+    IPv6 = 2,
+    //I2p = 3,
+    Tor = 4,
+}
```
SyntheticBird45, line 33: Defined an enum type instead of direct number for clarity.

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
+//! Onion address
+//!
+//! This module define v3 tor onion addresses
+//!
+//!
```
SyntheticBird45, line 5: This is the brand new `OnionAddr` type that is simply port + 56 character domain name. It needed to implement `Copy` so I ended up using a `[u8; 56]` (a `[u8; 62]` for including .onion would have been useless)

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    pub const fn port(&self) -> u16 {
+        self.port
+    }
+}
+
+/// Display for [`OnionAddr`] only print the **onion address**. Not the port.
+impl Display for OnionAddr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
SyntheticBird45, line 87: `Display` will print the domain only: `<domain>.onion`

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
...
+            .expect("Onion addresses are always containing UTF-8 characters.");
+
+        f.write_str(str.valid())?;
+        f.write_str(".onion")
+    }
+}
+
+/// [`OnionAddr`] parses an onion address **and a port**.
+impl FromStr for OnionAddr {
+    type Err = OnionAddrParsingError;
```
SyntheticBird45, line 101: `FromStr` or `parse()` will attempt to parse domain + port: `<domain>.onion:<port>`

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[test]
+    fn parse_invalid_onion_address_w_port() {
+        for addr in INVALID_ONION_ADDRESSES_WITH_PORT {
+            assert!(
+                addr.parse::<OnionAddr>().is_err(),
+                "Address {addr} has been reported as valid."
+            );
+        }
+    }
+}
```
SyntheticBird45, line 224: Added tests for validation.

p2p/address-book/src/lib.rs
```diff
@@ -33,6 +33,9 @@ pub struct AddressBookConfig {
     pub peer_store_directory: PathBuf,
     /// The amount of time between saving the address book to disk.
     pub peer_save_period: Duration,
+
+    /// Our own address to advertise to peers. (Only set if `Z::BROADCAST_OWN_ADDR` = `true`)
+    pub our_own_address: Option<Z::Addr>,
```
SyntheticBird45, line 38: `AddressBookConfig` now contains our self-designated anonymous address. It is first fetched at config time in `cuprated` and can optionally be added at runtime by say `Arti`.

This makes `AddressBookConfig` now `AddressBookConfig<Z: NetworkZone>`.

p2p/p2p-core/src/client/request_handler.rs
```diff
@@ -101,24 +103,58 @@ where
...
+
+            own_addr
+        } else {
+            None
+        };
+
+        let mut peer_list_req_size = MAX_PEERS_IN_PEER_LIST_MESSAGE;
+        if own_addr.is_some() {
+            peer_list_req_size -= 1;
+        }
```
SyntheticBird45, line 134: similar to monerod we're propagating our anonymous address through the `local_peerlist_new` field of timed sync response.

p2p/p2p-core/src/client/request_handler.rs
```diff
@@ -101,24 +103,58 @@ where
...
+        if let Some(own_addr) = own_addr {
+            // Append our address to the final peer list
+            peers.push(ZoneSpecificPeerListEntryBase {
+                adr: own_addr,
+                id: self.our_basic_node_data.peer_id,
+                last_seen: 0,
+                pruning_seed: PruningSeed::NotPruned,
+                rpc_port: self.our_basic_node_data.rpc_port,
+                rpc_credits_per_hash: self.our_basic_node_data.rpc_credits_per_hash,
+            });
```
SyntheticBird45, line 156: It's appended here.

p2p/p2p-core/src/network_zones/tor.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl NetworkZone for Tor {
+    const NAME: &'static str = "Tor";
+
+    const CHECK_NODE_ID: bool = false;
+
+    const BROADCAST_OWN_ADDR: bool = true;
+
+    type Addr = OnionAddr;
+}
```
SyntheticBird45, line 52: The new `Tor` netzone.

p2p/p2p-core/src/services.rs
```diff
@@ -175,4 +178,10 @@ pub enum AddressBookResponse<Z: NetworkZone> {
 
     /// Response to [`AddressBookRequest::GetBans`].
     GetBans(Vec<BanState<Z::Addr>>),
+
+    /// Response to [`AddressBookRequest::OwnAddress`]
+    ///
+    /// This returns [`None`] if the address book do
+    /// not contain a self designated address.
+    OwnAddress(Option<Z::Addr>),
```
SyntheticBird45, line 186: New request for getting the self designated address in `AddressBookConfig`

binaries/cuprated/src/config/p2p.rs
```diff
@@ -1,16 +1,18 @@
 use std::{
+    marker::PhantomData,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     path::Path,
     time::Duration,
 };
 
+use cuprate_wire::OnionAddr;
 use serde::{Deserialize, Serialize};
```
Boog900, line 9: :(

net/wire/src/network_address/epee_builder.rs
Boog900: IMHO I do not think this is necessary. This is too Rusty.

Just keeping it `u8` is wayy simpler here + we don't need to expose other API from epee-encoding

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        // The domain part must only contain base32 characters.
+        if !domain
+            .as_bytes()
+            .iter()
+            .copied()
+            .all(|c| c.is_ascii_lowercase() || (b'2'..=b'7').contains(&c))
+        {
+            return Err(OnionAddrParsingError::NonBase32Char);
+        }
```
Boog900, line 75: IMO this check is probably unneeded, people can still create invalid addresses even with this check and the only reason this would be invalid is because someone is actively being malicious.Even with this check if someone wanted to be malicious they can create invalid addresses.

net/wire/src/network_address/onion_addr.rs
Boog900: ports do mean something for Tor though, right?

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn from_str(addr: &str) -> Result<Self, Self::Err> {
+        let (addr, port) = addr
+            .split_at_checked(62)
+            .ok_or(OnionAddrParsingError::InvalidLength(addr.len()))?;
+
+        // Port
+        let port: u16 = port
+            .starts_with(':')
+            .then_some(port[1..].parse().ok())
```
Boog900, line 111: this will panic if the string is empty, i.e has nothing after the `.onion`.

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,224 @@
...
+            .flatten()
+            .ok_or(OnionAddrParsingError::InvalidPort)?;
+
+        // Address
+        Self::check_addr(addr)?;
+        let domain = addr.as_bytes()[..56]
+            .try_into()
+            .unwrap_or_else(|e| panic!("We just validated address: {addr} : {e}"));
```
Boog900, line 119: `check_addr` could return the domain

net/wire/src/network_address/onion_addr.rs
Boog900: seems like a good place for a simple `proptest`

p2p/p2p-core/src/client/request_handler.rs
```diff
@@ -1,3 +1,4 @@
+use cuprate_pruning::PruningSeed;
 use futures::TryFutureExt;
```
Boog900, line 2: :(

p2p/p2p-core/src/client/request_handler.rs
Boog900: this should be inserted at a random position, or the vec should be reshuffled.

Boog900 (changes requested):

binaries/cuprated/src/config/p2p.rs
SyntheticBird45: oops.

binaries/cuprated/src/config/p2p.rs
```diff
@@ -228,20 +230,36 @@ config_struct! {
...
+pub trait FetchOwnAddr<Z: NetworkZone> {
+    fn zone_own_addr(&self) -> Option<Z::Addr>;
+}
+
+impl FetchOwnAddr<ClearNet> for SharedNetConfig {
+    fn zone_own_addr(&self) -> Option<<ClearNet as NetworkZone>::Addr> {
+        // [`ClearNet::BROADCAST_OWN_ADDR`] = `false`
+        None
+    }
+}
```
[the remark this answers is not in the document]
SyntheticBird45: This is to avoid `address_book_config_tor`, `address_book_config_clearnet` duplication now that `AddressBookConfig` takes `Z`, the trait bounds forces definition while letting flexibility in the future on how we fetch this address (config now, ControlPort maybe later on for Tor, even some different mechanism for I2P).

net/wire/src/network_address/epee_builder.rs
SyntheticBird45: I would prefer it to be defined in a rusty way still, I didn't like the magic number + comments. @hinto-janai thoughts?

net/wire/src/network_address/onion_addr.rs
SyntheticBird45: https://github.com/monero-project/monero/blob/master/src/net/tor_address.cpp#L63-L67

It is needed as monerod also reject addresses with invalid characters. If we don't reject them Cuprate will end up relaying invalid addresses.

Moreover, I just find it to be common sense to make sure we don't include invalid data and later assumes we're only dealing with API that doesn't care about it ("let's arti/tor resolver sort the bad apples").

It's only returning an error as of now but as you said this is most likely a malicious intent, this could later be considered punishable if repeated.

net/wire/src/network_address/onion_addr.rs
SyntheticBird45: Display is implemented for using `to_string()` in `epee_builder.rs` which really just need the onion address, not the port (as it is put as u16 in port field). I could define a `addr_string()` for OnionAddr and add port to `Display` ?

net/wire/src/network_address/onion_addr.rs
Boog900: I don't think you ran the same test:

```
byte index 1 is out of bounds of ``
thread 'network_address::onion_addr::tests::parse_invalid_onion_address_w_port' panicked at net/wire/src/network_address/onion_addr.rs:111:28:
byte index 1 is out of bounds of ``
```

You have fallen into a pretty common trap, `then_some` is a function that takes in a value `T` and outputs it if the `bool` is `true`. The value `T` is computed before the call and passed. What you want is something that takes in a closure/function and runs that closure/function to produce the value `T` if the bool is `Some`.

```suggestion
            .then(|| port[1..].parse().ok())
```

net/wire/src/network_address/onion_addr.rs
Boog900: > I could define a addr_string() for OnionAddr and add port to Display ?

Yes please

net/wire/src/network_address/epee_builder.rs
SyntheticBird45: As precised in meeting, replaced with already existing `AddressType`.

Cargo.toml
```diff
@@ -110,6 +110,7 @@ arc-swap              = { version = "1", default-features = false }
 arrayvec              = { version = "0.7", default-features = false }
 async-trait           = { version = "0.1", default-features = false }
 bitflags              = { version = "2", default-features = false }
+base32                = { version = "0.5.1", default-features = false }
```
Boog900, line 113: do we need this dep anymore?

Cargo.toml
```diff
@@ -139,6 +140,7 @@ rayon                 = { version = "1", default-features = false }
 serde_bytes           = { version = "0.11", default-features = false }
 serde_json            = { version = "1", default-features = false }
 serde                 = { version = "1", default-features = false }
+sha3                  = { version = "0.10.8", default-features = false }
```
Boog900, line 143: and this one

net/wire/src/network_address/epee_builder.rs
```diff
@@ -1,21 +1,65 @@
...
 
+use super::OnionAddr;
+
+//---------------------------------------------------------------------------------------------------- Network address construction
+
+#[derive(Default)]
+/// There are no ordering guarantees in epee format and as such all potential fields can be collected during deserialization.
+/// The [`AllFieldsNetworkAddress`] is containing, as its name suggest, all optional field describing an address , as if it
+/// could be of any type.
+struct AllFieldsNetworkAddress {
```
Boog900, line 27: Why has this been moved? 

Although here it isn't too bad, doing reshuffles like this complicates the diff so should be avoided.

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,239 @@
...
+    ///
+    /// Returns a `String` containing the onion domain name and ".onion" TLD only, in form of `zbjkbs...ofptid.onion`.
+    pub fn addr_string(&self) -> String {
+        let mut str = self
+            .domain
+            .utf8_chunks()
+            .next()
+            .expect("Onion addresses are always containing UTF-8 characters.")
+            .valid()
+            .to_string();
```
Boog900, line 94: can we not just use https://doc.rust-lang.org/std/str/fn.from_utf8.html here?

net/wire/src/network_address/onion_addr.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/// Display for [`OnionAddr`]. It prints the onion address and port, in the form of `<domain>.onion:<port>`
+impl Display for OnionAddr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let str = self
+            .domain
+            .utf8_chunks()
+            .next()
+            .expect("Onion addresses are always containing UTF-8 characters.");
```
Boog900, line 118: and here

Boog900 (commented):
just some nits

net/wire/src/network_address/epee_builder.rs
Boog900: yeah different style, I normally go the other way round. I don't actually care which way too much but for existing code I wouldn't swap stuff around.

Boog900 (approved):

OBITOONDEADO1MZ (approved):

PR #494, opened 2025-05-30
P2P: Add missing check

YichiZhang0613:
### What

Fix problem mentioned in #497 .
### Why

Check if permit is none as required to prevent potential improper behavior later.
### Where

In p2p/p2p-core/src/handles.rs.
### How

It will make fn build panic when permit is not set for HandleBuilder.


Boog900:
> This PR was written by myself without AI.

Ok nice! do you use a tool to find these errors? if so what tool?

Issue #492, opened 2025-05-29
Workaround `flatten` imcompatability with `deny_unknown_fields`


hinto-janai:
## What
`serde` does not support `flatten` with `deny_unknown_fields`: https://serde.rs/field-attrs.html#flatten.

This causes problems with default values, for example in the config:
- https://github.com/Cuprate/cuprate/pull/423#discussion_r2110581752
- https://github.com/Cuprate/cuprate/actions/runs/15288154950/job/43002509225?pr=423#step:9:743

This could be worked around by manually implementing `flatten` in `config_macro`.

PR #495, opened 2025-05-30
`cuprated`: Fix config flatten

Boog900:
closes: #492

### DIFF
--- a/binaries/cuprated/src/config/macros.rs
+++ b/binaries/cuprated/src/config/macros.rs
@@ -4,7 +4,6 @@ use toml_edit::TableLike;
 /// can insert toml comments created from doc comments on fields.
 ///
 /// # Attributes
-/// - `#[flatten = true]`: lets the writer know that the field is flattened into the parent struct.
 /// - `#[child = true]`: writes the doc comments for all fields in the child struct.
 /// - `#[inline = true]`: inlines the struct into `{}` instead of having a separate `[]` header.
 /// - `#[comment_out = true]`: comments out the field.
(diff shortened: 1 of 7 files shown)


binaries/cuprated/src/config/macros.rs
```diff
@@ -65,12 +64,46 @@ use toml_edit::TableLike;
...
 ///
 /// `----------------------------------------------------------------------`
+///
+/// # Shared Values
+/// Sometimes multiple different configs will overlap in their fields. Manually duplicating the fields
+/// between the structs can lead to maintenance issues as the docs need to be kept in sync. Also using
+/// serde's `flatten` is not supported with `deny_unknown_fields`: <https://serde.rs/field-attrs.html#flatten>.
+///
+/// So this macro provides a way to define shared fields between configs, just add a `Shared` object
+/// with the wanted fields at the top and all structs afterwards will have those fields inserted.
```
Boog900, line 74: `config_struct` is now a [TT Muncher](https://lukaswirth.dev/tlborm/decl-macros/patterns/tt-muncher.html#incremental-tt-munchers) and accepts multiple structs.

Recursively calling itself on every struct used with optional shared values defined first.

binaries/cuprated/src/config/macros.rs
```diff
@@ -87,57 +122,105 @@ macro_rules! config_struct {
...
+               }
+            )?
+            $($tt)*
+        }
+    };
+}
 
-                    'write_field: {
-                        let key_str = &stringify!($field);
+macro_rules! __write_docs {
```
Boog900, line 179: I have defined a new macro to write each field so we don't need to duplicate this code between shared & non-shared fields.

binaries/cuprated/src/rpc/server.rs
```diff
@@ -26,17 +23,29 @@ use crate::{
...
+            false,
+        ),
+        (
+            (
+                config.restricted.enable,
+                config.restricted.address,
+                config.restricted.request_byte_limit,
+            ),
+            true,
+        ),
```
Boog900, line 42: slightly ugly but we no longer have a shared inner struct

hinto-janai (approved):
`--generate-config` diffs with `main` are OK.

PR #499, opened 2025-06-03
storage: add `RuntimeError::ResizedByAnotherProcess`

hinto-janai:
### What
Adds `RuntimeError::ResizedByAnotherProcess` and related handling code to `storage/` crates.

### Why
If another process resizes the database file, the storage code will currently panic.

Discussed here: https://github.com/monero-project/meta/issues/1212.

### DIFF
--- a/storage/service/src/service/write.rs
+++ b/storage/service/src/service/write.rs
@@ -135,7 +135,12 @@ fn database_writer<Req, Res>(
             let response = inner_handler(env, &request);
 
             // If the database needs to resize, do so.
-            if ConcreteEnv::MANUAL_RESIZE && matches!(response, Err(RuntimeError::ResizeNeeded)) {
+            if ConcreteEnv::MANUAL_RESIZE
+                && matches!(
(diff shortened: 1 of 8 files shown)


storage/service/src/service/write.rs
```diff
@@ -135,7 +135,12 @@ fn database_writer<Req, Res>(
             let response = inner_handler(env, &request);
 
             // If the database needs to resize, do so.
-            if ConcreteEnv::MANUAL_RESIZE && matches!(response, Err(RuntimeError::ResizeNeeded)) {
+            if ConcreteEnv::MANUAL_RESIZE
+                && matches!(
+                    response,
+                    Err(RuntimeError::ResizeNeeded | RuntimeError::ResizedByAnotherProcess)
+                )
+            {
```
Boog900, line 143: IMO this should still be a panic for the write interface - no other process should write to cuprated's database.

The read tower interface should handle this though IMO, currently it just returns an error.

PR #509, opened 2025-06-10
Add Tor support to Cuprate (Arti, Tor Daemon, Dandelion router)

SyntheticBird45:
### What

Part of [ongoing Tor support development](https://ccs.getmonero.org/proposals/syntheticbird_cuprate_arti_2_months.html) for Cuprate

Follows: https://github.com/Cuprate/cuprate/pull/446 and https://github.com/Cuprate/cuprate/pull/481

### Changelog

In `workspace`:
- New dependencies: `arti_client`, `Cuprate/tokio-socks.git`, `tor-cell`, `tor-config-path`, `tor-hsservice`, `tor-persist`, `tor-proto`, `tor-rtcompat` (yes nothing was exported).
- In `deny.toml`, whitelisted `Unlicense` license for `arti_client`.

In `cuprate-p2p-transport`:
- Implemented `Transport<ClearNet>` and `Transport<Tor>` for `Arti`. New `ArtiClientConfig`, `ArtiServerConfig` configuration. New `OnionListener` for accepting inbound connections from arti's generated onion service.
- Implemented `Transport<Tor>` for `Daemon`. New `DaemonClientConfig`, `DaemonServerConfig` configuration. New `DaemonInboundStream` listening for incoming TCP  connections from the tor daemon.
- `DisabledListener` as a polyfill for transports with inbound disabled, such as `Transport<ClearNet> for Arti` and in the future `Transport<ClearNet> for Socks5`.

In `cuprate-p2p-core`:
- Removed `Default` and `Debug` bound from `Transport::ClientConfig` and `Transport::ServerConfig`.
- Removed `Clone` bound from `Transport::ServerConfig`.

In `cuprate-p2p`:
- Changed some function visibility to `pub(super)` instead of `pub`.

In `cuprate-wire`:
- Added `borsh` dependency and `BorshSerialize` and `BorshDeserialize` derived implementation to `OnionAddr` for `BorshNetworkZone` requirement in address book.

In `cuprated`:
- New `tor` module containing the initialization of Arti, config helpers and context structure `TorContext` to pass down to p2p initialization function and other helpers.
- New `config/tor` module containing the `[tor]` configuration table. It define tor daemon related variables, as well as arti settings.
- Added `enable_inbound` field to `ClearNetConfig` to disable incoming listener by default.
- Added `proxy` field to `ClearNetConfig` for enabling clearnet over arti and in the future proxy urls.
- Added `TorNetConfig` for setting `Tor` network zone parameters such as listening ports, enabling arti inbound server, or setting an anonymous inbound onion address from an external daemon.
- Modified `initialize_zones_p2p` to now start Tor network zone and use the correct transport depending on user configuration.
- In `txpool/*`, generalized `DiffuseService`, `OutboundPeerStream` and `ConcreteDandelionRouter` for `Z: NetworkZone`. Created a new `MainDandelionRouter` service that will route local txs to a Tor router instead of clearnet if available. Adapted initialization to the changes.


binarybaron:
very cool!

binaries/cuprated/src/tor.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub arti_onion_service: Option<OnionService>,
+}
+
+/// Initialize the Tor network if enabled in configuration
+///
+/// This function will bootstrap Arti if needed by Tor network zone or
+/// clearnet as a proxy.
+pub async fn initialize_tor_if_enabled(config: &Config) -> TorContext {
+    let mode = config.tor.mode;
+    let anonymize_clearnet = config.p2p.clear_net.proxy.to_lowercase() == "Tor";
```
Boog900, line 65: `to_lowercase` `Tor` :)

Do you think it might be good idea to make `proxy` an enum?

binaries/cuprated/src/tor.rs
```diff
@@ -0,0 +1,190 @@
...
+//---------------------------------------------------------------------------------------------------- Transport configuration
+
+pub fn transport_arti_config(config: &Config, ctx: TorContext) -> TransportConfig<Tor, Arti> {
+    // Extracting
+    let Some(bootstrapped_client) = ctx.bootstrapped_client else {
+        panic!("Arti client should be initialized");
+    };
+    let Some(client_config) = ctx.arti_client_config else {
+        panic!("Arti client should be initialized");
+    };
```
Boog900, line 142: bit of nit but these can be combined:
```suggestion
    let (Some(bootstrapped_client), Some(client_config)) = (ctx.bootstrapped_client, ctx.arti_client_config) else {
        panic!("Arti client should be initialized");
    };
```

binaries/cuprated/src/tor.rs
```diff
@@ -0,0 +1,190 @@
...
+        server_config: None,
+    }
+}
+
+pub fn transport_daemon_config(config: &Config) -> TransportConfig<Tor, Daemon> {
+    TransportConfig::<Tor, Daemon> {
+        client_config: DaemonClientConfig {
+            tor_daemon: config.tor.daemon.address,
+        },
+        server_config: (!config.tor.daemon.anonymous_inbound.is_empty()).then_some(
```
Boog900, line 183: should this instead check the bool field?

binaries/cuprated/src/txpool/dandelion.rs
```diff
@@ -1,11 +1,14 @@
...
-    pool::DandelionPoolService, DandelionConfig, DandelionRouter, Graph,
+    pool::DandelionPoolService, DandelionConfig, DandelionRouteReq, DandelionRouter,
+    DandelionRouterError, Graph, State, TxState,
 };
 use cuprate_p2p::NetworkInterface;
-use cuprate_p2p_core::ClearNet;
+use cuprate_p2p_core::{client::InternalPeerID, ClearNet, NetworkZone, Tor};
 use cuprate_txpool::service::{TxpoolReadHandle, TxpoolWriteHandle};
+use futures::future::BoxFuture;
+use tower::{Service, ServiceExt};
```
Boog900, line 11: group

deny.toml
```diff
@@ -101,6 +101,7 @@ ignore = [
 allow = [
     # Nothing required - free to use without permission.
     "CC0-1.0", # https://creativecommons.org/publicdomain/zero/1.0/
+    "Unlicense",
```
Boog900, line 104: is it possible to instead list the crates to allow?

binaries/cuprated/src/config/p2p.rs
```diff
@@ -185,20 +209,85 @@ config_struct! {
...
+        /// Examples     | false
+        pub inbound_onion: bool,
     }
 }
 
 impl Default for ClearNetConfig {
     fn default() -> Self {
         Self {
+            p2p_port: 18080,
+            enable_inbound: false,
```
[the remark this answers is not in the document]
SyntheticBird45: There is an ethical question at hand. We should definitely avoid additional behavior without consent. I would be pretty pissed off if my software suddenly tried to advertise me as being reachable and worse that IPs inbounding towards me without me realizing it. This should be an opt-in thing, this also force people to RTFM before enabling it.

binaries/cuprated/src/config/tor.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        /// Enable isolated circuits for Arti.
+        ///
+        /// If set, Arti will use different tor circuits for each connections. This can
+        /// cause stability issues if the connection count is important.
+        ///
+        /// Type         | boolean
+        /// Valid values | false, true
+        /// Examples     | false
+        pub isolated_circuit: bool,
```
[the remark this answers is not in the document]
SyntheticBird45: Kept in (discussed in DM). Aftermatch of https://cfp.twed.org/mk5/talk/AQBMB3/ presentation. Isolated circuits can yield a throughput/performance benefit

binaries/cuprated/src/tor.rs
SyntheticBird45: It should check bool and throw if anonymous_inbound is empty

deny.toml
SyntheticBird45: There are quite a lot of them. Every subcrates of arti is `Unlicense`

binaries/cuprated/src/config.rs
```diff
@@ -227,6 +236,44 @@ impl Config {
...
 
+    /// The [`Tor`], [`cuprate_p2p::P2PConfig`].
+    pub fn tor_p2p_config(&self, ctx: &TorContext) -> cuprate_p2p::P2PConfig<Tor> {
+        let our_onion_address = match ctx.mode {
+            TorMode::Off => None,
+            TorMode::Daemon => Some(
+                OnionAddr::new(
+                    &self.tor.daemon.anonymous_inbound,
+                    self.p2p.tor_net.p2p_port
+                ).expect("Unable to parse supplied `anonymous_inbound` onion address. Please make sure the address is correct.")),
```
Boog900, line 247: does this require someone to always supply an inbound addr for daemon mode?

binaries/cuprated/src/config/p2p.rs
```diff
@@ -165,6 +177,17 @@ config_struct! {
...
+
+        /// Enable IPv4 inbound server.
+        ///
+        /// The inbound server will listen on port `p2p.clear_net.p2p_port`.
+        /// Setting this to `false` will disable incoming IPv4 P2P connections.
+        ///
+        /// Type         | boolean
+        /// Valid values | false, true
+        /// Examples     | false
+        pub enable_inbound: bool,
```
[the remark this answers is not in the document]
Boog900: why? they are both for changing if we accept inbound connections

binaries/cuprated/src/config/p2p.rs
```diff
@@ -185,20 +210,85 @@ config_struct! {
...
+
+        #[comment_out = true]
+        /// Enable Arti inbound onion service.
+        ///
+        /// Setting this to `true` will enable Arti's onion service for accepting inbound
+        /// Tor P2P connections. The onion service will listen on port `p2p.tor_net.p2p_port`
+        /// The keypair and therefore onion address is generated randomly.
+        ///
+        /// This setting is ignored in `Daemon` mode.
+        ///
```
Boog900, line 256: I think this section needs to be updated as it isn't just for arti anymore

binaries/cuprated/src/config/p2p.rs
```diff
@@ -185,20 +209,85 @@ config_struct! {
...
+
+impl Default for TorNetConfig {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            inbound_onion: false,
             p2p_port: 18080,
+            outbound_connections: 12,
+            extra_outbound_connections: 2,
+            max_inbound_connections: 12,
```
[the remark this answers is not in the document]
Boog900: `max_inbound_connections` should be like 128

binaries/cuprated/src/tor.rs
Boog900: the enum I was thinking would be something like:

```rust
enum {
    Tor,
    Socks(String)
}
```

with the socks being a catch all, this is a nit but does improve the code over string matching :shrug:

binaries/cuprated/src/tor.rs
Boog900: well then the comments should be different and mention the field IMO

binaries/cuprated/src/tor.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    TransportConfig::<Tor, Daemon> {
+        client_config: DaemonClientConfig {
+            tor_daemon: config.tor.daemon.address,
+        },
+        server_config: (!invalid_onion).then_some(DaemonServerConfig {
+            ip: config.tor.daemon.listening_addr.ip(),
+            port: config.tor.daemon.listening_addr.port(),
+        }),
```
Boog900, line 193: if `inbound_onion` is false then a server config will be created here, is that intended?

Boog900 (changes requested):

binaries/cuprated/src/config.rs
SyntheticBird45: I am afraid it is. Will fix that.

binaries/cuprated/src/tor.rs
SyntheticBird45: I agree. thx for clarifying

binaries/cuprated/src/config/p2p.rs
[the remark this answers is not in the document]
SyntheticBird45: `enable_inbound` would have:
```
Enable IPv4 inbound server.

The inbound server will listen on port `p2p.clear_net.p2p_port`.
Settings this to `false` will disable incoming IPv4 P2P connections.
```
`inbound_onion` would have:
```
Enable Tor inbound onion server.

In Arti mode, setting this to `true` will enable Arti's onion serice for accepting inbound Tor P2P connections. They keypair and therefore onion address is generated randomly on first run.

In Daemon mode, setting this to `true` will enable a TCP server listening for inbound connections from your Tor daemon. Refer to the `tor.anonymous_inbound` and `tor.listening_addr` field for onion address and listening configuration.

The server will listen on port `p2p.tor_net.p2p_port`
```

Now if we put it in `Shared`:
```
Enable inbound server for this zone.

The inbound server will listen on the `p2p_port` field of this zone.
Setting this to `false` will disable incoming P2P connections for this zone.
```

We lose so many information for arguably not that much deduplication gain if you consider you have to document it elsewhere. People will immediately see that the comments are duplicated, the beginners will not understand, try to run and realize it don't work. The more knowledgeable will try to seek for documentation elsewhere. "Where is this documentation?" They will be roaming around the configuration file. The best is that the relevant field you are using is redirecting you towards concerned and mandatory fields. The same way that when you read a man page there is always a [`See also`](https://www.man7.org/linux/man-pages/man5/hosts.5.html) section. `enable_inbound` DO NOT work on its own under tor zone, unlike clearnet. You can't use the same documentation for a field that have a different behavior.

I'll just profit from this comment to say that the user book should NEVER, NEVER **exclusively** hold essential documentation for making cuprate work. You do not know if someone is working on an headless environment with no access to a browser elsewhere or if the website is prohibited for some reason.

binaries/cuprated/src/config/p2p.rs
```diff
@@ -185,20 +210,88 @@ config_struct! {
...
+        ///
+        /// Setting this to "" (an empty string) will disable Tor.
+        ///
+        /// When "Daemon" is set, the Tor daemon address to use can be
+        /// specified in `tor.daemon_addr`.
+        ///
+        /// Type         | String
+        /// Valid values | "Arti", "Daemon", "Off"
+        /// Examples     | "Arti"
+        pub enabled: bool,
```
Boog900, line 246: this comment needs updating

binaries/cuprated/src/tor.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    TransportConfig::<Tor, Daemon> {
+        client_config: DaemonClientConfig {
+            tor_daemon: config.tor.daemon.address,
+        },
+        server_config: valid_onion.then_some(DaemonServerConfig {
+            ip: config.tor.daemon.listening_addr.ip(),
+            port: config.tor.daemon.listening_addr.port(),
+        }),
+    }
```
Boog900, line 191: `valid_onion` will still be true if `inbound_onion` is false

Boog900 (changes requested):

Boog900:
Still have some changes I want to make, will put a comment when I am done.

Boog900 (approved):

Boog900:
@SyntheticBird45 If you can review my changes I'll merge

PR #529, opened 2025-08-14
consensus: Remove unwrap() calls from RandomXVm creation

unldenis:
### What

If the creation of `RandomXVm` fails, it would be better to propagate the error than to panic.


Boog900:
thanks for the PR. Do you have any reasons in mind to do this? how can our usage of RX cause an error?

unldenis:
we could propagate the error and pass the task of handling the error to the caller
https://github.com/Cuprate/cuprate/blob/27a97ab106b61225b678cf7376dee35a0c74653f/consensus/src/block.rs#L225

PR #536, opened 2025-08-27
fix tx-pool invalid version message

Boog900:
Fixes the message to update the tx-pool, we were testing with a new table that isn't in the old version.


storage/txpool/src/free.rs
```diff
@@ -69,7 +69,7 @@ pub fn open(config: &Config) -> Result<ConcreteEnv, InitError> {
 
         // Store if this DB has been used before by checking if the metadata table exists.
```
SyntheticBird45, line 70: This comment probably requires an update?

PR #545, opened 2025-09-05
cuprated: add pop blocks command

Boog900:


consensus/context/src/rx_vms.rs
```diff
@@ -214,9 +214,25 @@ impl RandomXVmCache {
...
+
+        if self.seeds.len() < RX_SEEDS_CACHED {
+            let seed_heights = get_last_rx_seed_heights(new_height, RX_SEEDS_CACHED)
+                .split_at(self.seeds.len())
+                .1
+                .to_vec();
+            let seed_hashes = get_block_hashes(seed_heights.clone(), database).await?;
+
+            self.seeds.extend(seed_heights.into_iter().zip(seed_hashes));
+        }
```
Boog900, line 233: The RX cache would panic if too many blocks were popped as it would have no seeds, this adds the seeds

consensus/context/src/weight.rs
```diff
@@ -137,31 +137,33 @@ impl BlockWeightsCache {
...
+        {
+            get_long_term_weight_in_range(
+                new_long_term_start_height..(new_long_term_start_height + numb_blocks),
+                database.clone(),
+                Chain::Main,
+            )
+            .await?
+        } else {
+            vec![]
+        };
```
Boog900, line 151: the old way to do this for the weight cache was completely wrong

Boog900: although it was only a problem for reorgs over the short term window

PR #549, opened 2025-09-10
`cuprated`: Generic SOCKS5 proxy support + ClearNet over Daemon

SyntheticBird45:
### What

Add support to pass P2P traffic to a specific SOCKS5 proxy. Authentication is supported. This is also used to implement `ClearNet` over `Daemon` tor anonymization of clearnet traffic.

Follows: https://github.com/Cuprate/cuprate/pull/509

### Changelog

In `cuprate-p2p-transport`:

- Added new `socks.rs` file containing `Socks` that `impl Transport<ClearNet>`. As well as `SocksClientConfig` (socks5 proxy settings).

In `cuprated`:

- Added `TryFrom<ProxySettings> for SocksClientConfig` to parse socks5 URL in user configuration. The urls are in the form of: `socks5://username:password@host:port`.
- new `socks_transport_config` and `transport_clearnet_daemon_config` helpers.
- Modified `initialize_zones_p2p` accordingly.

### DIFF
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -18,7 +18,7 @@ use cuprate_p2p_core::{
     transports::{Tcp, TcpServerConfig},
     ClearNet, NetworkZone, Tor, Transport,
 };
-use cuprate_p2p_transport::{Arti, ArtiClientConfig, ArtiServerConfig};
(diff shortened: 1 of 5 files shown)


Boog900 (approved):
I don't really like how we don't parse the url in the config and leave it as a string but changing that is harder than it seems

Issue #551, opened 2025-09-27
Private file permissions by default


hinto-janai:
## Feature
`cuprated` creates files with global read/execute permissions (`rwxr-xr-x`), any other process running on the machine can read these files.

Certain libraries that create files (e.g. LMDB) may set their own `umask` although `cuprated` does not. It would be prudent to set a private global `umask` before creating files:

| umask | file permissions |
|-------|------------------|
| `027` | `rwxr-x---`      |
| `007` | `rwx------`      |

## Implementation
Windows will need a separate impl.

```rust
fn main() {
    // SAFETY: calling C.
    unsafe {
        libc::umask(0o027);
    }

    /* ... */
}
```

PR #562, opened 2025-11-04
cuprated: private file permissions (Unix)

hinto-janai:
### What
Implements https://github.com/Cuprate/cuprate/issues/551 for Unix, not yet for Windows.


consensus/fast-sync/proptest-regressions/fast_sync.txt
Boog900: can we remove this I don't think this is related right?

Issue #442, opened 2025-04-11
`cuprated --dry-run`


hinto-janai:
## Feature
Add a `cuprated --dry-run` flag.

```
./cuprated --help

--dry-run        Test the configuration file and exit.
                 `cuprated` will check the configuration for
                 correct syntax and valid values, then print
                 the configuration and exit successfully without
                 starting. If any errors occur, they will be 
                 logged and `cuprated` and exit with an error code.
```

## How
Things that are tested:
- `Config` deserialization
- PATH permissions
- If ports can be binded to

Things that could be tested and aren't hard errors:
- If values are sane (e.g. `max_txpool_byte_size` is unrealistically large, system only has 10 threads but a value like 500 is present, etc) 
- Remaining disk space is too low
- Unrestricted RPC binded to `0.0.0.0`

## Why
https://en.wikipedia.org/wiki/Dry_run_(testing)

Useful for programs that have lots of side-effects after starting, e.g.:

- [`nginx -T`](https://nginx.org/en/docs/switches.html)
- [`cargo add --dry-run`](https://doc.rust-lang.org/cargo/commands/cargo-add.html#dependency-options)
- [`cargo install --dry-run`](https://doc.rust-lang.org/cargo/commands/cargo-update.html#update-options)
- [`cargo upload --dry-run`](https://doc.rust-lang.org/cargo/commands/cargo-publish.html#publish-options)
- [`docker compose --dry-run`](https://docs.docker.com/reference/cli/docker/compose/#options)

PR #563, opened 2025-11-05
cuprated: --dry-run options with check config parse, ports and paths

iskyd:
cuprated: adding --dry-run option

### What
Closes: [#<442>](https://github.com/Cuprate/cuprate/issues/442)

This PR add the --dry-run option to check config deserialization, ports availability and path permissions.
There is another pr in the same scope (https://github.com/Cuprate/cuprate/pull/461). This PR takes care of reviews and added file permissions check.

### DIFF
--- a/binaries/cuprated/src/config.rs
+++ b/binaries/cuprated/src/config.rs
@@ -403,7 +403,7 @@ impl Config {
         }
 
         if self.rpc.restricted.enable {
-            let port = p2p_port(self.rpc.restricted.port, self.network);
+            let port = restricted_rpc_port(self.rpc.restricted.port, self.network);
             let ip = self.rpc.restricted.address;
 
             match Self::check_port(ip, port) {
@@ -416,7 +416,7 @@ impl Config {
         }
 
         if self.rpc.unrestricted.enable {
-            let port = p2p_port(self.rpc.unrestricted.port, self.network);
+            let port = unrestricted_rpc_port(self.rpc.unrestricted.port, self.network);
             let ip = self.rpc.unrestricted.address;
 
             match Self::check_port(ip, port) {


binaries/cuprated/src/config.rs
```diff
@@ -319,6 +326,96 @@ impl Config {
...
+        }
+
+        let test_file = path.join(".cuprate_write_test");
+        if let Err(e) = std::fs::write(&test_file, b"Cuprate") {
+            anyhow::bail!("No write permission for {}", path.display());
+        }
+
+        if let Err(e) = std::fs::remove_file(&test_file) {
+            anyhow::bail!("Cannot remove temporary file from {}", path.display());
+        }
```
hinto-janai: `cuprated` at this point hasn't made these directories so it errors here:

```
cargo r --bin cuprated -- --dry-run --skip-config-warning

Port 0.0.0.0:18080 available.
Error: Path does not exist /home/hinto/.local/share/cuprate
Error Path does not exist /home/hinto/.cache/cuprate
Checks failed.
```

 maybe these checks would be better done with `std::fs::OpenOptions`.

iskyd: @hinto-janai could we consider creating both directories if they don't yet exist during the dry run check?

hinto-janai: Probably fine although it would be nice if `--dry-run` itself didn't have side effects, I think you could check if the parent directory has the right permissions instead of creating a file.

iskyd: @hinto-janai we can just check parent permissions even if that doesn't guarantee that the new dirs will have the proper permissions.
We can also check if the dir already exists than check it otherwise check the parent.
Not sure how to properly check permissions without creating a temporary file, it would be unreliable.

binaries/cuprated/src/config.rs
```diff
@@ -13,6 +14,7 @@ use clap::Parser;
 use safelog::DisplayRedacted;
 use serde::{Deserialize, Serialize};
 
+use anyhow::bail;
```
hinto-janai: nit: should be grouped with other 3rd party crates.

binaries/cuprated/src/config.rs
```diff
@@ -319,6 +327,110 @@ impl Config {
...
+
+            match Self::check_port(IpAddr::V6(ip), port) {
+                Ok(()) => println!("Port {ip}:{port} available."),
+                Err(e) => {
+                    eprintln_red(&format!("Error: {e}"));
+                    errors.push(e);
+                }
+            }
+        }
+
```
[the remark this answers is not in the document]
hinto-janai: > it seems is not something we bind to

Yup, my bad.

binaries/cuprated/src/config.rs
```diff
@@ -394,7 +394,46 @@ impl Config {
...
+            let ip = self.rpc.unrestricted.address;
+
+            match Self::check_port(ip, port) {
+                Ok(()) => println!("Rpc unrestricted {ip}:{port} available."),
+                Err(e) => {
+                    eprintln_red(&format!("Error: {e}"));
+                    errors.push(e);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
iskyd: @hinto-janai we cannot directly use self.rpc.restricted.port since check_port takes an u16 while self.rpc.restricted.port is also a DefaultOrCustom.

PR #577, opened 2026-02-05
cuprated: broadcast newly added alt blocks to peers

### LINKED ISSUE #524: Broadcast new alt blocks
## What
New alt-blocks should be broadcasted to peers as well as new blocks.

## Why
Doing what we currently do can slow down the network as it will cause nodes to initiate a sync if they don't have the block and the alt chain wins.

## Where
https://github.com/Cuprate/cuprate/blob/9c2c942d2fcf26ed8916dc3f9be6db43d8d2ae78/binaries/cuprated/src/blockchain/manager/handler.rs#L95-L96

redsh4de:
## What
Broadcasts newly cached alt blocks to peers

Closes #524

## Why
Prevent nodes from initiating a sync if they don't have a alt block and the alt chain wins

## Where
- `cuprated`
    - `blockchain/manager/handler.rs` - split `Cached` bool in `AddAltBlock` into two variants - `AlreadyCached` and `NewlyCached(bytes)`, which points to the block blob. Used in `handle_incoming_block` to determine whether the block is newly added, and broadcasts the block if it is

## How
Effectively the same pattern as for main blocks

### DIFF
--- a/binaries/cuprated/src/blockchain/manager/handler.rs
+++ b/binaries/cuprated/src/blockchain/manager/handler.rs
@@ -122,12 +122,19 @@ impl super::BlockchainManager {
             let block_hash = block.hash();
             let res = self.handle_incoming_alt_block(block, prepared_txs).await?;
 
-            if matches!(res, AddAltBlock::Cached(true)) {
+            if let AddAltBlock::NewlyCached(block_blob) = res {
                 info!(
                     alt_block = true,
                     hash = hex::encode(block_hash),
                     "Successfully added block"
                 );
+
+                let chain_height = self
+                    .blockchain_context_service
+                    .blockchain_context()
+                    .chain_height;
+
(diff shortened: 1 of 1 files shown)


binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -122,12 +122,19 @@ impl super::BlockchainManager {
...
                 info!(
                     alt_block = true,
                     hash = hex::encode(block_hash),
                     "Successfully added block"
                 );
+
+                let chain_height = self
+                    .blockchain_context_service
+                    .blockchain_context()
+                    .chain_height;
```
Boog900: monerod seems to assume this is the block's height in its logs, which for alt blocks it might not be. This shouldn't be an issue just wanted to note it.

Issue #554, opened 2025-10-01
`cuprated` as a library


hinto-janai:
## What
Initial implementation and type/signature proposal for `cuprated` as a library.

https://github.com/Cuprate/cuprate/issues/516.

## Where
`binaries/cuprated` could be a double binary/library.

Another option is `binaries/cuprated` could be turned into a thin shell around a new library that contains the core node functionality, perhaps this would be cleaner. 

Library name/location: TODO.

## How
Some changes/fixes are necessary as `binaries/cuprated` currently assumes it is a standalone binary e.g.:

- Many types/signatures are `pub`
- Global `rayon`, `tokio`, `tracing` usage; these will panic when initialized elsewhere
- CLI handling is intermixed with the config code, this must be separated

## API (wip)
```rust
//! # `cuprated` library.

/// An active `cuprated` node.
///
/// `Drop` will end the node process.
pub struct Node {/* ... */}

pub struct NodeConfig {/* ... */}
pub struct NodeEventListener {/* ... */}

impl Node {
    /// Launch a new `cuprated` process.
    pub fn launch(conifg: NodeConfig) -> Result<Self, Error>;

    /// `Stream` of events emitted by `cuprated`.
    pub fn events(&self) -> NodeEventListener;

    /// Current config in use.
    pub fn config(&self) -> &NodeConfig;
}

pub mod statics {
    /* `binaries/cuprated/src/statics.rs` could be moved here */
}

pub mod signals {
    /* `binaries/cuprated/src/signals.rs` could be moved here */
}
```

PR #585, opened 2026-02-18
cuprated: Add graceful shutdown

redsh4de:
### What
Ports over the graceful shutdown logic from fjall-wip branch: https://github.com/Cuprate/cuprate/blob/fjall-wip/binaries/cuprated/src/monitor.rs

Depends on the embeddable version #592

### Why
More idiomatic shutdown pattern, previously every exit was equivalent to a crash

### Where
- `cuprated`:
  - `monitor.rs` - `TaskExecutor` handle (cancellation token + task tracker)
  - `config.rs` - `target_max_memory` / `fjall_cache_size` / `block_downloader_config` return Result instead of panicking on unresolved memory
  - `main.rs` - wires up executor, OS signal handler, double-Ctrl+C force exit, shutdown wait loop
  - `commands.rs` - add `Exit` command
  - `blockchain/manager.rs, syncer.rs` - shutdown token in `select` loops, replace `todo!()` with `break`
  - `txpool/manager.rs, incoming_tx.rs` - shutdown token, change ordering by priority due to using `biased`
  - `rpc/server.rs` - `with_graceful_shutdown` on axum, tracked spawns, shutdown token plumbed through
  - `rpc/handlers/other_json.rs` - wiring for `stop_daemon` RPC
  - `lib.rs` - `Node::launch` returns `Result`, creates and distributes `TaskExecutor`, Drop impl
- `storage/blockchain`:
  - `database.rs` - detach drop-log `info!` from caller span

### How
`monitor.rs` introduces `TaskExecutor`, took nomenclature inspo from [reth](https://reth.rs/docs/reth_ethereum/tasks/type.TaskExecutor.html). Instead of how reth does it though with a [static](https://github.com/paradigmxyz/reth/pull/15360/changes), its threaded into the functions to support running multiple `Node` instances in the same process

All `tokio::spawn` replaced with `TaskExecutor::spawn`. Task select loops now use `biased` ordering with the shutdown token as higher priority. 

On shutdown signal, the token is cancelled -> tasks wind down one by one, and `main()` waits on task tracker to finish unwinding before dropping the runtime.

`Node::launch` now returns `Result<Self, anyhow::Error>` instead of panicking on database/service init failures.

Double Ctrl+C force exits.

### DIFF
--- a/binaries/cuprated/src/blockchain/syncer.rs
+++ b/binaries/cuprated/src/blockchain/syncer.rs
@@ -20,6 +20,8 @@ use cuprate_p2p::{
 };
 use cuprate_p2p_core::{client::PeerSyncCallback, ClearNet, CoreSyncData, NetworkZone};
 
+use tokio_util::sync::CancellationToken;
+
 use super::BlockchainInterface;
 
 /// An error returned from the [`Syncer`].
@@ -72,6 +74,7 @@ impl Syncer {
     /// Run the syncer.
     #[instrument(level = "debug", skip_all)]
     #[expect(clippy::significant_drop_tightening)]
+    #[expect(clippy::too_many_arguments)]
     pub async fn run<CN>(
         mut self,
         mut context_svc: BlockchainContextService,
@@ -80,6 +83,7 @@ impl Syncer {
         incoming_block_batch_tx: mpsc::Sender<(BlockBatch, Arc<OwnedSemaphorePermit>)>,
         stop_current_block_downloader: Arc<Notify>,
         block_downloader_config: BlockDownloaderConfig,
+        shutdown_token: CancellationToken,
     ) -> Result<(), SyncerError>
     where
         CN: Service<
@@ -98,7 +102,13 @@ impl Syncer {
         let mut sync_permit = Arc::new(Arc::clone(&semaphore).acquire_owned().await?);
 
         loop {
-            self.notify_syncer.notified().await;
+            tokio::select! {
+                () = shutdown_token.cancelled() => {
+                    tracing::info!("Blockchain syncer shut down.");
+                    return Ok(());
+                }
+                () = self.notify_syncer.notified() => {}
+            }
 
             tracing::trace!("Checking connected peers to see if we are behind",);
 
@@ -126,6 +136,11 @@ impl Syncer {
 
             loop {
                 tokio::select! {
+                    biased;
+                    () = shutdown_token.cancelled() => {
+                        tracing::info!("Blockchain syncer shut down.");
+                        return Ok(());
+                    }
                     () = stop_current_block_downloader.notified() => {
                         tracing::info!("Received stop signal, stopping block downloader");
 
@@ -154,6 +169,9 @@ impl Syncer {
 
                         tracing::debug!("Got batch, len: {}", batch.blocks.len());
                         if incoming_block_batch_tx.send((batch, Arc::clone(&sync_permit))).await.is_err() {
+                            if shutdown_token.is_cancelled() {
+                                return Ok(());
+                            }
                             return Err(SyncerError::IncomingBlockChannelClosed);
                         }
                     }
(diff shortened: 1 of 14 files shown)


redsh4de:
Depends on #592 now.

binaries/cuprated/src/blockchain/syncer.rs
```diff
@@ -98,7 +102,13 @@ impl Syncer {
         let mut sync_permit = Arc::new(Arc::clone(&semaphore).acquire_owned().await?);
 
         loop {
-            self.notify_syncer.notified().await;
+            tokio::select! {
```
SyntheticBird45, line 104: missing a biased ?

binaries/cuprated/src/commands.rs
```diff
@@ -93,11 +96,18 @@ pub fn command_listener(incoming_commands: mpsc::Sender<Command>) -> ! {
...
 pub async fn io_loop(mut incoming_commands: mpsc::Receiver<Command>, mut node: cuprated::Node) {
     let start_instant = Instant::now();
+    let shutdown_token = node.task_executor.cancellation_token();
 
     loop {
-        let Some(command) = incoming_commands.recv().await else {
-            tracing::warn!("Shutting down io_loop command channel closed.");
-            return;
+        let command = tokio::select! {
+            () = shutdown_token.cancelled() => break,
```
SyntheticBird45, line 104: missing biased here too ?

binaries/cuprated/src/config.rs
```diff
@@ -362,28 +362,41 @@ impl Config {
...
-    pub fn target_max_memory(&self) -> u64 {
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `target_max_memory` is unresolved.
+    pub fn target_max_memory(&self) -> anyhow::Result<u64> {
         match self.target_max_memory {
             DefaultOrCustom::Default => {
-                panic!("`target_max_memory` is unresolved; call `resolve_max_memory` first")
+                anyhow::bail!("`target_max_memory` is unresolved; call `resolve_max_memory` first")
```
SyntheticBird45: I think we want to panic there. It's a code assumption if I'm not mistaken

binaries/cuprated/src/lib.rs
```diff
@@ -115,6 +121,15 @@ pub struct Node {
...
     /// The configuration this node was launched with.
     pub config: Arc<Config>,
+
+    /// Task spawning and shutdown executor.
+    pub task_executor: TaskExecutor,
+}
+
+impl Drop for Node {
+    fn drop(&mut self) {
+        self.task_executor.trigger_shutdown();
```
SyntheticBird45, line 131: API wise I think `Node` should not shutdown on drop. It should be manually triggered

SyntheticBird45 (commented):
I let a few comment but overall lgtm

binaries/cuprated/src/blockchain/syncer.rs
redsh4de: oop

binaries/cuprated/src/config.rs
redsh4de: hmm yeah not a fan of the panic in what's now library code but it is a invariant ig

launch time only though so should be fine to keep panic

binaries/cuprated/src/lib.rs
redsh4de: Took shutdown-on-drop from original proposal: #554 

i think it makes sense to keep, because if someone drops the node without having cloned the executor prior, they won't have a way to trigger shutdown anymore and it will just keep running in the runtime. Its only possible while its held, so the Drop's adds defense against that

SyntheticBird45: Fair enough

SyntheticBird45 (approved):

Boog900 (approved):

Issue #191, opened 2024-06-22
In house transaction and block types


Boog900:
## What

Currently we are using `monero-serai` directly across multiple crates, this proposal is to create wrapper types in `types` that we use across crates instead. 

## Why

This will allow us to do things like caching the tx hash, we currently do this in the consensus crates, however we also calculate tx hashes in the block downloader, which isn't cached.  

## Where

`types` and anywhere currently using monero-serai

## How

Right now monero-serai is planning a big change to their transaction type seen as we are going to have to make (a lot of) changes for this it makes sense to do this at the same time

kayabaNerve:
Not to say you shouldn't add a wrapper struct, yet I'm curious the argument against localized hashes.

Hash a TX as it comes in, store said hash with the TX.

Is the argument that the wrapper, despite being a wrapper, would offer better overall UX and prevent cache misses across distinct areas? If you use the Deref types as Box does, it _should_ be quite seamless I think...

Boog900:
yes, mainly for UX. I am also considering only exposing the necessary API for Cuprate on the the types although I don't really think the benefits are there.

PR #586, opened 2026-02-20
cuprated: Add graceful shutdown, pt2: service error propagation

redsh4de:
### What
Depends on pt1, PR #585

Building on part 1, this replaces the panic-on-error patterns in cuprated's services with error propagation that makes use of the graceful shutdown mechanism

### Why
So we dont panic and insta-crash upon a error, shutdown should be graceful. Internal errors still caused panics before this via PANIC_CRITICAL_SERVICE_ERROR

### Where
- `cuprated`:
  - `blockchain/error.rs` (new):
    | Type | Kind | Variants |
    |---|---|---|
    | `BlockValidationError` | peer-fault (ban) | `HardFork(HardForkError)`, `Other(ExtendedConsensusError)` |
    | `BlockManagerError` | manager union | `Validation(BlockValidationError)`, `Internal(#[from] tower::BoxError)` |
    | `IncomingBlockError` | interface union | `Validation(BlockValidationError)`, `Internal(tower::BoxError)`, `Orphan`, `UnknownTransactions(_, _)`, `TooManyTxs`, `ChannelClosed` |

  - `txpool/error.rs` (new):
    | Type | Kind | Variants |
    |---|---|---|
    | `TxValidationError` | peer-fault (ban) | `Parse(io::Error)`, `Consensus(ExtendedConsensusError)`, `DuplicateTransaction`, `RelayRule(RelayRuleError)` |
    | `IncomingTxError` | union | `Validation(#[from] TxValidationError)`, `Internal(#[from] tower::BoxError)` |

  - `monitor.rs` - add `TaskExecutor::spawn_critical`, `panic_message` helper.
  - `logging.rs` - return the log guard, hold it until shutdown completes.
  - `constants.rs` - rename `PANIC_CRITICAL_SERVICE_ERROR` to `CRITICAL_SERVICE_ERROR`
  - `lib.rs` - `Node::launch` returns `Result`; on init failure it cancels partially spawned subsystems before returning the error.
  - `blockchain.rs` - `check_add_genesis` returns `Result`.
  - `blockchain/manager.rs` - run loop returns `Result`, `spawn_critical` for syncer + manager.
  - `blockchain/manager/handler.rs` - `.expect(...)` -> `?` throughout, handlers return `Result<_, BlockManagerError>` or `Result<_, tower::BoxError>`, `handle_command` routes `Internal` via `?` and `Validation` via response channel. Invariant violations kept as panics.
  - `txpool/manager.rs` - uniform DB-error escalation: all 6 handlers propagate `TxPoolError` via `?`. Reorder `promote_tx` and `remove_tx_from_pool` to write to DB first, then apply in memory.
  - `txpool/incoming_tx.rs` - `IncomingTxHandler::init` returns `Result`.
  - `tor.rs`, `p2p.rs` - `transport_*_config` / `initialize_clearnet_p2p` / `initialize_tor_p2p` return `Result`, Tor logs+skips on failure.
  - `rpc/server.rs` - `init_rpc_servers` returns `Result`, per-server task uses `spawn` - If a RPC server fails to start, the error is logged but doesn't initiate shutdown.
- `cuprate-types`
    - `TxConversionError` re-exported.

#### 1. spawn_critical wraps each subsystem's future:

  | Outcome | Action |
  |---|---|
  | `Ok(())` | normal exit if shutdown was requested, else logged as a bug -> trigger shutdown |
  | `Err(_)` | log w/ subsystem name -> trigger shutdown |
  | panic | log via `panic_message` -> trigger shutdown |

#### 2. Added layered errors.
Per subsystem: a typed `ValidationError` for peer-fault paths, plus a union with `Validation` and `Internal(tower::BoxError)` arms. `From` impls route (`ExtendedConsensusError::DBErr` -> Internal; else -> Validation). The manager matches between the two:

  | Variant | Route | Caller's response |
  |---|---|---|
  | `Internal(_)` | `?` -> `spawn_critical` | log + graceful shutdown |
  | `Validation(_)` | back via response channel | ban peer / cancel downloader |

### DIFF
--- a/binaries/cuprated/src/blockchain/manager.rs
+++ b/binaries/cuprated/src/blockchain/manager.rs
@@ -7,11 +7,15 @@ use tokio_util::sync::CancellationToken;
 use tower::{BoxError, Service, ServiceExt};
 use tracing::error;
 
-use cuprate_blockchain::service::{BlockchainReadHandle, BlockchainWriteHandle};
+use cuprate_blockchain::{
+    service::{BlockchainReadHandle, BlockchainWriteHandle},
+    BlockchainError,
+};
 use cuprate_consensus::{
     BlockChainContextRequest, BlockChainContextResponse, BlockchainContextService,
     ExtendedConsensusError,
 };
+use cuprate_consensus_rules::ConsensusError;
 use cuprate_p2p::{
     block_downloader::{self, BlockBatch},
     BroadcastSvc, NetworkInterface,
@@ -20,16 +24,72 @@ use cuprate_p2p_core::ClearNet;
 use cuprate_txpool::service::TxpoolWriteHandle;
 use cuprate_types::{
     blockchain::{BlockchainReadRequest, BlockchainResponse},
-    Chain, TransactionVerificationData,
+    Chain, TransactionVerificationData, TxConversionError,
 };
 
 use crate::{
     blockchain::{
-        chain_service::ChainService, syncer::BlockchainSyncer, types::ConsensusBlockchainReadHandle,
+        chain_service::ChainService, syncer::BlockchainSyncer,
+        types::ConsensusBlockchainReadHandle, BlockValidationError, IncomingBlockError,
     },
+    monitor::FatalError,
     txpool::TxpoolManagerHandle,
     LaunchContext,
 };
+
+/// An error from the blockchain manager's internal handlers.
+#[derive(Debug, thiserror::Error)]
+pub enum BlockManagerError {
+    /// The peer sent us an invalid block; ban them.
+    #[error(transparent)]
+    Validation(BlockValidationError),
+
+    /// We cannot recover; shut the node down.
+    #[error(transparent)]
+    Fatal(#[from] FatalError),
+}
+
+impl From<ExtendedConsensusError> for BlockManagerError {
+    fn from(e: ExtendedConsensusError) -> Self {
+        match e {
+            ExtendedConsensusError::DBErr(e) => Self::Fatal(e),
+            ExtendedConsensusError::ConErr(e) => Self::Validation(e.into()),
+
+            ExtendedConsensusError::TxsIncludedWithBlockIncorrect
+            | ExtendedConsensusError::OneOrMoreBatchVerificationStatementsInvalid
+            | ExtendedConsensusError::NoBlocksToVerify => {
+                Self::Validation(BlockValidationError::Consensus(e))
+            }
+        }
+    }
+}
+
+impl From<ConsensusError> for BlockManagerError {
+    fn from(e: ConsensusError) -> Self {
+        Self::Validation(e.into())
+    }
+}
+
+impl From<BlockManagerError> for IncomingBlockError {
+    fn from(e: BlockManagerError) -> Self {
+        match e {
+            BlockManagerError::Validation(e) => Self::Validation(e),
+            BlockManagerError::Fatal(e) => Self::Fatal(e),
+        }
+    }
+}
+
+impl From<BlockchainError> for BlockManagerError {
+    fn from(e: BlockchainError) -> Self {
+        Self::Fatal(e.into())
+    }
+}
+
+impl From<TxConversionError> for BlockManagerError {
+    fn from(e: TxConversionError) -> Self {
+        Self::Fatal(e.into())
+    }
+}
 
 mod commands;
 mod handler;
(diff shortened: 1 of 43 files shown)


binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -359,32 +386,26 @@ impl super::BlockchainManager {
...
         else {
             unreachable!();
         };
 
         match chain {
             Some((Chain::Alt(_), _)) => return Ok(AddAltBlock::AlreadyCached),
-            Some((Chain::Main, _)) => anyhow::bail!("Alt block already in main chain"),
+            Some((Chain::Main, _)) => return Err(BlockValidationError::AlreadyInMainChain.into()),
             None => (),
         }
```
redsh4de, line 426: monerod doesn't ban here if a block exists on main chain

[`add_new_block`](https://github.com/monero-project/monero/blob/v0.18.5.0/src/cryptonote_core/blockchain.cpp#L4672-L4677) returns `false` when the block already exists, but [`handle_incoming_block`](https://github.com/monero-project/monero/blob/v0.18.5.0/src/cryptonote_core/cryptonote_core.cpp#L1426) does nothing with it -> ends up returning `true`. The handlers [ban gate](https://github.com/monero-project/monero/blob/v0.18.5.0/src/cryptonote_protocol/cryptonote_protocol_handler.inl#L668) only checks `m_verifivation_failed`, which the "already exists"  path doesn't set

redsh4de:
fwiw after these changes i kind of want to rename `ExtendedConsensusError` to `ConsensusServiceError`, as now its just used in service calls to differentiate whether a error is a consensus validation, or a service failure. easier to reason as well, but seperate pr

binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -95,9 +106,9 @@ impl super::BlockchainManager {
     ///
     /// Otherwise, this function will validate and add the block to the main chain.
     ///
-    /// # Panics
+    /// # Errors
     ///
-    /// This function will panic if any internal service returns an unexpected error that we cannot
+    /// This function will return an [`Err`] if any internal service returns an unexpected error that we cannot
     /// recover from.
```
Boog900: This doc is somewhat confusing now, I think we should just remove it or put how the error could be just for the block

binaries/cuprated/src/blockchain/manager.rs
```diff
@@ -20,18 +24,67 @@ use cuprate_p2p_core::ClearNet;
...
+#[derive(Debug, thiserror::Error)]
+pub enum BlockManagerError {
+    /// The peer sent us an invalid block; ban them.
+    #[error(transparent)]
+    Validation(BlockValidationError),
+
+    /// We cannot recover; shut the node down.
+    #[error(transparent)]
+    Fatal(#[from] FatalError),
+}
```
Boog900: IMO this does not need to be a separate type, we could just change the variants on `ExtendedConsensusError`.

binaries/cuprated/src/p2p/request_handler.rs
```diff
@@ -355,6 +354,7 @@ async fn new_fluffy_block<A: NetZoneAddress>(
         return Ok(ProtocolResponse::NA);
     }
 
+    let block_hash = block.hash();
```
Boog900: I don't think this hash is worth it.

binaries/cuprated/src/p2p.rs
```diff
@@ -160,47 +162,50 @@ pub async fn initialize_clearnet_p2p(
...
                     config.clearnet_p2p_config(),
-                    transport_clearnet_arti_config(tor_ctx),
+                    transport_clearnet_arti_config(tor_ctx)?,
+                    Some(peer_sync_callback),
+                )
+                .await
+            }
+            TorMode::Daemon => {
+                start_zone_p2p::<ClearNet, Socks>(
+                    &launch_ctx.blockchain,
```
Boog900, line 171: we have the `blockchain` variable but then do this as well, it is a bit confusing I think. IMO we should just remove the `blockchain` variable

Boog900 (commented):
Just a few nits

binaries/cuprated/src/p2p/request_handler.rs
redsh4de: what about using the block number instead? i think we gotta some identifier for which block failed

redsh4de: ah this is in the fluffy blocks path, yeah not worth it

Boog900: With this: https://github.com/Cuprate/cuprate/issues/191 I think we could do it as it would not add any overhead.

Boog900 (approved):

PR #587, opened 2026-02-21
Tapes db

Boog900:


binaries/cuprated/src/config/p2p.rs
```diff
@@ -102,9 +103,14 @@ config_struct! {
 
 impl From<BlockDownloaderConfig> for cuprate_p2p::block_downloader::BlockDownloaderConfig {
     fn from(value: BlockDownloaderConfig) -> Self {
+        let mut info = sysinfo::System::new();
+        info.refresh_memory();
+
+        let buffer_mem = u64_to_usize(min(info.total_memory() / 5, 1024 * 1024 * 1024));
+
```
Boog900: TODO: change this

storage/blockchain/src/ops/alt_block/block.rs
```diff
@@ -163,34 +172,46 @@ pub fn get_alt_block_hash(
...
+        Chain::Alt(chain_id) => tx_ro
+            .get(
+                &db.alt_blocks_info,
+                bytemuck::bytes_of(&AltBlockHeight {
+                    chain_id: chain_id.into(),
+                    height: *block_height,
+                }),
+            )?
+            .map(|info| {
+                let info: BlockInfo = bytemuck::pod_read_unaligned(info.as_ref());
```
redsh4de: Is L188 correct?

L105 uses `CompactAltBlockInfo` instead of `BlockInfo` for the read

storage/blockchain/src/ops/output.rs
```diff
@@ -1,260 +1,190 @@
...
-    // Delete the output data itself.
-    tables.outputs_mut().delete(pre_rct_output_id)
+    if let Some(e) = err {
+        return Err(e.into());
+    }
+
+    let pre_rct_output_id = PreRctOutputId {
+        amount,
+        // The new `amount_index` is the length of amount of outputs with same amount.
+        amount_index: *num_outputs,
```
redsh4de: shouldn't this be *num_outputs - 1?

if its 0 indexed, then if there is 1 output, it will try to remove from index 1, which is the 2nd item - doesn't exist

storage/blockchain/src/service/write.rs
```diff
@@ -53,127 +125,85 @@ fn handle_blockchain_request(
...
+    // generate a `ChainId` for the popped blocks.
+    let old_main_chain_id = ChainId(rand::random());
+
+    // pop the blocks
+    for _ in 0..numb_blocks {
+        crate::ops::block::pop_block(db, Some(old_main_chain_id), &mut tx_rw, &mut tapes)?;
     }
+
+    tx_rw.commit()?;
+    tapes.commit(Persistence::SyncAll)?;
```
redsh4de: these two should probably be swapped so that tapes commits before fjall as well, so its consistent with how writes do it now

storage/txpool/src/service/write.rs
```diff
@@ -50,20 +82,45 @@ fn handle_txpool_request(
...
+        Input::Gen(_) => unreachable!(),
+    }) {
+        let e = in_progress_key_images.entry(*ki.as_bytes());
 
-    if let Err(e) = ops::add_transaction(tx, state_stem, &mut tables_mut) {
-        drop(tables_mut);
+        match e {
+            Entry::Occupied(o) => return Ok(TxpoolWriteResponse::AddTransaction(Some(*o.get()))),
+            Entry::Vacant(v) => {
+                v.insert(*ki.as_bytes());
```
redsh4de: why are the ki bytes inserted, instead of tx hash? AddTransaction takes in a `Option<TransactionHash>`

storage/blockchain/src/service/read.rs
```diff
@@ -791,93 +685,96 @@ fn next_chain_entry(
...
-    if numb_txs < missing_txs.len() {
+    if block.transactions.len() < missing_txs.len() {
         return Ok(BlockchainResponse::TxsInBlock(None));
     }
 
     let txs = missing_txs
         .into_iter()
-        .map(|index_offset| Ok(tables.tx_blobs().get(&(first_tx_index + index_offset))?.0))
+        .map(|index_offset| {
+            get_tx_blob_from_id(&(block_info.mining_tx_index + index_offset), &tapes, db)
```
redsh4de: is there a reason why this is not something like `mining_index + 1 + index_offset` like in other places that add a +1?

consensus/src/block.rs
redsh4de: why is `hardfork_signal` used for `hf_version`, and `hardfork_version` used for `hf_vote` for alts?

other places use `hardfork_version` for `version`, and hardfork_signal for `vote` (except `alt_block/block.rs` which uses version for vote as well, so this seems intentional but i dont get why)

binaries/cuprated/src/config.rs
```diff
@@ -305,29 +337,40 @@ impl Config {
...
 
-        // We don't set reader threads as we manually make the reader threadpool.
-        cuprate_blockchain::config::ConfigBuilder::default()
-            .network(self.network)
-            .data_directory(self.fs.data_directory.clone())
-            .sync_mode(blockchain.sync_mode)
-            .build()
+        cuprate_blockchain::config::Config {
+            blob_dir: self.fs.fast_data_directory.clone(),
+            index_dir: self.fs.slow_data_directory.clone(),
```
redsh4de: wont not using path_with_network (like for fjall below) cause testnet and mainnet write to the same files? in case someone runs a mainnet node, then decides to run a testnet node on the same machine

storage/blockchain/src/ops/tx.rs
```diff
@@ -1,282 +1,284 @@
...
-///
-/// Thus, after [`remove_tx`], those values (outputs and key images)
-/// will be remove from database tables as well.
-///
-#[doc = doc_error!()]
+/// Removes a transaction from the dynamic tables.
 #[inline]
-pub fn remove_tx(tx_hash: &TxHash, tables: &mut impl TablesMut) -> DbResult<(TxId, Transaction)> {
-    //------------------------------------------------------ Transaction data
-    let tx_id = tables.tx_ids_mut().take(tx_hash)?;
```
redsh4de, line 191: main used to have .take() which deletes the tx_hash, i dont see the `remove_tx_from_dynamic_tables` deleting the tx_hash from tx_ids - only key images and outputs

edit: looks like it impacts `add_alt_transaction_blob` in `src/ops/alt_block/tx.rs` as well - sees tx hash exists, skips adding the blob

consensus/fast-sync/src/create.rs
```diff
@@ -47,28 +46,43 @@ struct Args {
...
+                if let Ok(block_ids) = read_batch(&mut read_handle, height).await {
+                    let hash = hash_of_hashes(block_ids.as_slice());
+                    Ok(Hex(hash))
+                } else {
+                    println!("Failed to read next batch from database");
+                    Err("Failed to read next batch from database")
+                }
+            }
+        })
+        .collect::<futures::stream::FuturesUnordered<_>>();
```
redsh4de: main used a while loop to build the hashes sequentially - should this be `FuturesOrdered` instead?

storage/blockchain/src/ops/alt_block/block.rs
redsh4de: related to previous: why does L198 use `hardfork_version`, instead of `hardfork_signal` like `ops/block.rs`?

binaries/cuprated/src/blockchain/manager/handler.rs
redsh4de, line 334:
```suggestion
                    warn!(
                        "Failed to verify alt block: {}, error {}, banning peer.",
                        hex::encode(hash),
                        e
                    );
                    batch.peer_handle.ban_peer(LONG_BAN);
```
I'd add the warn! log for alt blocks as well to match how main does it now

redsh4de (commented):
some comments

storage/blockchain/src/ops/alt_block/block.rs
Boog900: nice spot, fixed

storage/blockchain/src/ops/output.rs
Boog900: fixed

storage/txpool/src/service/write.rs
Boog900: Because 2 txs can spend the same key image but have different hashes, we can't have double spends in the pool so we use the key images themselves here to prevent this.

storage/blockchain/src/ops/tx.rs
Boog900: Oops forgot to add the explicit removal when I move to a write batch instead of an atomic tx, should be fixed now

consensus/fast-sync/src/create.rs
Boog900: good spot, I remember thinking I need to use the ordered one as well :)

storage/blockchain/src/ops/alt_block/block.rs
Boog900: nice spot, this is incorrect.

storage/blockchain/src/service/read.rs
Boog900: Fixed

consensus/src/block.rs
Boog900: This was a mistake, should now be fixed.

storage/txpool/src/service/write.rs
Boog900: oh wait I see what you mean yeah that should be the tx hash.

storage/blockchain/src/ops/alt_block/tx.rs
```diff
@@ -1,76 +1,79 @@
...
             tx_hash: tx.tx_hash,
-        },
-    )?;
+        }),
+    );
 
-    if tables.tx_ids().get(&tx.tx_hash).is_ok()
-        || tables.alt_transaction_blobs().get(&tx.tx_hash).is_ok()
-    {
+    if db.tx_ids.contains_key(tx.tx_hash)? || db.alt_transaction_blobs.contains_key(tx.tx_hash)? {
```
redsh4de: we do now remove `tx_hash` from `tx_ids` but noticed that changes are commited **after** this function runs at `service/write.rs` - since `contains_key()` reads commited state, it would still would return `true`, and skip adding the blob when `alt_transaction_blobs` check might have returned `false`

Would removing the `tx_ids` check here be fine? Seems like the simplest way to handle it

redsh4de: or by checking the uncommited state

storage/blockchain/src/ops/output.rs
```diff
@@ -1,260 +1,190 @@
...
+            Some(Ok(o)) => u64::from_be_bytes(o[8..].try_into().unwrap()) + 1,
+            Some(Err(e)) => {
+                err = Some(e);
+                0
+            }
+        }
+    });
+
+    if let Some(e) = err {
+        return Err(e.into());
```
redsh4de, line 46: we add amount to cache on L32 but don't remove it on error

edit: alternatively make sure that amount is inserted in cache only on success to avoid needing a remove-on-err, that would be more correct

```suggestion
        pre_rct_numb_outputs_cache.remove(&amount);
        return Err(e.into());
```

storage/blockchain/src/ops/output.rs
```diff
@@ -1,260 +1,190 @@
...
+                0
             }
-        })?;
+        }
+    });
 
-    // Delete the output data itself.
-    tables.outputs_mut().delete(pre_rct_output_id)
+    if let Some(e) = err {
+        return Err(e.into());
```
redsh4de, line 90: same here

```suggestion
        pre_rct_numb_outputs_cache.remove(&amount);
        return Err(e.into());
```

storage/blockchain/src/ops/block.rs
```diff
@@ -214,84 +391,179 @@ pub fn pop_block(
...
-    // Add the miner tx to the blob.
-    let mut miner_tx_blob = tables.tx_blobs().get(&miner_tx_idx)?.0;
-    block.append(&mut miner_tx_blob);
+    tapes.truncate_fixed_sized_tape(&db.tx_infos, block_info.mining_tx_index);
 
-    // Add the blocks tx hashes.
-    monero_oxide::io::VarInt::write(&block_txs.len(), &mut block)
-        .expect("The number of txs per block will not exceed u64::MAX");
+    let cumulative_rct_outs = tapes
+        .read_entry(&db.block_infos, block_height as u64 - 1)?
```
redsh4de, line 414: nit, to match L215
```suggestion
        .read_entry(&db.block_infos, block_height.saturating_sub(1) as u64)?
```

storage/blockchain/src/service/read.rs
```diff
@@ -636,54 +540,50 @@ fn key_images_spent(env: &ConcreteEnv, key_images: HashSet<KeyImage>) -> Respons
...
+        tapes
+            .read_entry(&db.block_infos, height)?
+            .ok_or(BlockchainError::NotFound)
+    };
 
-    let top_block_height = top_block_height(&table_block_heights)?;
+    let top_block_height = tapes
+        .fixed_sized_tape_len(&db.block_infos)
+        .expect("Required tape not open")
+        - 1;
```
redsh4de, line 584: style nit: could port the same defensive check from `chain_height`

unnecessary as genesis block always gets added but good for consistency

```suggestion
    let chain_height = tapes
      .fixed_sized_tape_len(&db.block_infos)
      .expect("Required tape not open");
    
    if chain_height == 0 {
        return Err(BlockchainError::NotFound);
    }
    
    let top_block_height = chain_height - 1;
```

storage/blockchain/src/service/read.rs
```diff
@@ -751,23 +640,28 @@ fn next_chain_entry(
 
     // The returned chain entry must overlap with one of the blocks  we were told about.
     let first_known_block_hash = block_ids[idx];
-    let first_known_height = table_block_heights.get(&first_known_block_hash)?;
+    let first_known_height =
+        block_height(db, &tx_ro, &first_known_block_hash)?.ok_or(BlockchainError::NotFound)?;
 
-    let chain_height = crate::ops::blockchain::chain_height(table_block_heights)?;
+    let chain_height = crate::ops::blockchain::chain_height(db, &tapes)?;
```
redsh4de, line 646: same here

```suggestion
    let chain_height = crate::ops::blockchain::chain_height(db, &tapes)?;
    
    if chain_height == 0 {
        return Err(BlockchainError::NotFound):
    }
    
```

storage/blockchain/src/ops/block.rs
```diff
@@ -345,286 +639,123 @@ pub fn get_block_extended_header_from_height(
...
-    tables: &impl Tables,
+    db: &BlockchainDatabase,
+    tapes: &tapes::TapesReadTransaction,
 ) -> DbResult<(ExtendedBlockHeader, BlockHeight)> {
-    let height = chain_height(tables.block_heights())?.saturating_sub(1);
-    let header = get_block_extended_header_from_height(&height, tables)?;
+    let height = u64_to_usize(
+        tapes
+            .fixed_sized_tape_len(&db.block_infos)
+            .expect("Require tape not found")
```
redsh4de, line 650: its over literally unshippable
```suggestion
            .expect("Required tape not found")
```

storage/blockchain/src/ops/blockchain.rs
```diff
@@ -20,59 +20,37 @@ use crate::{
...
-pub fn chain_height(table_block_heights: &impl DatabaseRo<BlockHeights>) -> DbResult<BlockHeight> {
+pub fn chain_height(
+    db: &BlockchainDatabase,
+    tapes: &tapes::TapesReadTransaction,
+) -> DbResult<BlockHeight> {
     #[expect(clippy::cast_possible_truncation, reason = "we enforce 64-bit")]
-    table_block_heights.len().map(|height| height as usize)
+    Ok(tapes
+        .fixed_sized_tape_len(&db.block_infos)
+        .expect("Required tape must exists") as usize)
```
redsh4de, line 31: here too
```suggestion
        .expect("Required tape must exist") as usize)
```

storage/blockchain/src/service/write.rs
```diff
@@ -53,127 +125,85 @@ fn handle_blockchain_request(
...
+            &block.block,
+            &block.block_hash,
+            block.txs.iter().map(|tx| Cow::Borrowed(&tx.tx)),
+            &mut numb_transactions,
+            &mut tx_rw,
+            &mut pre_rct_numb_outputs_cache,
+        )?;
     }
+
+    tx_rw.commit()?;
```
redsh4de, line 165: note: with graceful shutdown all errors after tapes commit, but before fjall successfully commits should be distinguished due to us being out of sync after that. `BlockchainError::DatabaseOutOfSync`?

`handler.rs` does panic on error now so not a issue rn, but graceful shutdown changes this

idea: recovery function that replays only the out of sync blocks from tapes into fjall when we get that kind of error instead instead of a full chain nuke on desync

storage/blockchain/src/ops/blockchain.rs
```diff
@@ -20,59 +20,37 @@ use crate::{
...
 ///
 /// For example:
-/// - The blockchain has 0 blocks => this returns `Err(RuntimeError::KeyNotFound)`
+/// - The blockchain has 0 blocks => this returns `Err(BlockchainError::KeyNotFound)`
 /// - The blockchain has 1 block (height 0) => this returns `Ok(0)`
 /// - The blockchain has 2 blocks (height 1) => this returns `Ok(1)`
 ///
 /// Note that in cases where no blocks have been written to the
-/// database yet, an error is returned: `Err(RuntimeError::KeyNotFound)`.
+/// database yet, an error is returned: `Err(BlockchainError::KeyNotFound)`.
```
redsh4de, line 44: `KeyNotFound` -> `NotFound`

redsh4de (commented):
mostly nitpicks

storage/blockchain/src/ops/output.rs
Boog900: An error here is unrecoverable so the cache being invalid doesn't matter IMO.

redsh4de (approved):

storage/blockchain/src/service/write.rs
Boog900: Yeah I was going to do that originally, however full fjall nuke is much easier to handle + is not too slow. You can test it by just deleting the fjall DB but keeping the tapes.

storage/blockchain/src/ops/block.rs
hinto-janai, line 1: nit: I don't think you care to change this but in the `storage/` diffs there are many u32/u64/usize `as` ops that could use `cuprate_helper::cast` already in scope.

storage/blockchain/src/ops/block.rs
```diff
@@ -1,138 +1,321 @@
...
+            append_tx.append_bytes(&db.pruned_blobs, &prunable_hash)?;
+        }
+    }
+
+    tracing::trace!("pruned_tape_index: {}", pruned_tape_index);
+
+    // Split the blocks at the point the pruning stripe changes.
+    let start_height = blocks[0].height;
+    let first_block_pruning_seed = cuprate_pruning::DecompressedPruningSeed::new(
+        cuprate_pruning::get_block_pruning_stripe(start_height, usize::MAX, 3).unwrap(),
```
hinto-janai, line 111: nit:

> rg "usize::MAX, 3"
```
storage/blockchain/src/ops/tx.rs
259:            cuprate_pruning::get_block_pruning_stripe(tx_info.height, usize::MAX, 3).unwrap();

storage/blockchain/src/ops/block.rs
111:        cuprate_pruning::get_block_pruning_stripe(start_height, usize::MAX, 3).unwrap(),
143:            cuprate_pruning::get_block_pruning_stripe(blocks[0].height, usize::MAX, 3).unwrap();
405:    let stripe = cuprate_pruning::get_block_pruning_stripe(block_height, usize::MAX, 3).unwrap();
```

```rust
// cuprate_pruning

pub fn get_default_block_pruning_stripe(start_height) -> u32 {
    get_block_pruning_stripe(
        block_height,
        usize::MAX,
        CRYPTONOTE_PRUNING_LOG_STRIPES
    ).unwrap()
}
```

storage/blockchain/src/ops/alt_block/tx.rs
```diff
@@ -1,76 +1,73 @@
...
-    ops::macros::{doc_add_alt_block_inner_invariant, doc_error},
-    tables::{Tables, TablesMut},
+    error::{BlockchainError, DbResult},
     types::{AltTransactionInfo, TxHash},
+    BlockchainDatabase,
 };
 
-/// Adds a [`VerifiedTransactionInformation`] from an alt-block
-/// if it is not already in the DB.
+/// Adds a [`VerifiedTransactionInformation`] from an alt-block if it is not already in the DB.
```
hinto-janai, line 12: nit: there's probably a lot of inaccurate docs at this point; this function replaces it unconditionally.

storage/blockchain/src/ops/block.rs
hinto-janai, line 414: https://github.com/Cuprate/cuprate/pull/587#discussion_r2997388173

Call stack:
- `cuprated::blockchain::manager::pop_blocks`
- `BlockchainWriteRequest::PopBlocks`
- `cuprate_blockchain::service::write::pop_blocks`
- `cuprate_blockchain::ops::block::pop_blocks`

AFAICT input that can pop the genesis is allowed? I think this will underflow.

storage/blockchain/src/ops/alt_block/chain.rs
```diff
@@ -1,85 +1,96 @@
...
-///
 /// # Panics
 ///
-/// This will panic if [`AltBlockHeight::height`] == `0`.
+/// This may panic if the block is invalid.
+///
+/// **THIS IS NOT ATOMIC**
+///
 pub fn update_alt_chain_info(
+    db: &BlockchainDatabase,
```
hinto-janai, line 22: Shouldn't this be given a `tx_rw: &mut fjall::OwnedWriteBatch`? It is only called from `add_alt_block` which has it open already.

storage/blockchain/src/database.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl BlockchainDatabase {
+    /// Open a [`BlockchainDatabase`] with an [`fjall::Database`] for storing data that can't be stored in tapes.
+    pub fn open_with_fjall_database(
+        config: &Config,
+        fjall: fjall::Database,
+    ) -> Result<Self, BlockchainError> {
+        let block_heights = fjall.keyspace("block_heights", KeyspaceCreateOptions::default)?;
+        let key_images = fjall.keyspace("key_images", KeyspaceCreateOptions::default)?;
+        let pre_rct_outputs = fjall.keyspace("pre_rct_output", KeyspaceCreateOptions::default)?;
```
hinto-janai, line 52:
```suggestion
        let pre_rct_outputs = fjall.keyspace("pre_rct_outputs", KeyspaceCreateOptions::default)?;
```

storage/blockchain/src/types.rs
```diff
@@ -150,47 +143,37 @@ pub struct PreRctOutputId {
...
+    }
+
+    /// Deserializes the bytes into a [`PreRctOutputId`]
+    pub fn from_bytes(bytes: &[u8; 16]) -> Self {
+        Self {
+            amount: Amount::from_be_bytes(bytes[..8].try_into().unwrap()),
+            amount_index: AmountIndex::from_be_bytes(bytes[8..].try_into().unwrap()),
+        }
+    }
+}
```
hinto-janai, line 162: Why not use bytemuck for this? Why big endian?

storage/blockchain/src/ops/output.rs
```diff
@@ -1,260 +1,190 @@
...
+    let num_outputs = pre_rct_numb_outputs_cache.entry(amount).or_insert_with(|| {
+        let last_out = db.pre_rct_outputs.prefix(amount.to_be_bytes()).next_back();
+
+        match last_out.map(fjall::Guard::key) {
+            None => 0,
+            Some(Ok(o)) => u64::from_be_bytes(o[8..].try_into().unwrap()) + 1,
+            Some(Err(e)) => {
+                err = Some(e);
+                0
             }
```
hinto-janai, line 85:
```suggestion
            Some(Ok(o)) => u64::from_be_bytes(o[8..].try_into().unwrap()) + 1,
            None | Some(Err(e)) => {
                err = Some(e);
                0
}
```

https://github.com/Cuprate/cuprate/pull/587#discussion_r2984946003

If L80 branch is taken then L96 will be `0 - 1`.

storage/blockchain/src/database.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub(crate) fjall: fjall::Database,
+
+    pub(crate) block_heights: fjall::Keyspace,
+    pub(crate) key_images: fjall::Keyspace,
+    pub(crate) pre_rct_outputs: fjall::Keyspace,
+    pub(crate) tx_ids: fjall::Keyspace,
+    pub(crate) v1_tx_outputs: fjall::Keyspace,
+    pub(crate) alt_chain_infos: fjall::Keyspace,
+    pub(crate) alt_block_heights: fjall::Keyspace,
+    pub(crate) alt_blocks_info: fjall::Keyspace,
```
hinto-janai, line 27:
```suggestion
    pub(crate) alt_block_infos: fjall::Keyspace,
```

storage/blockchain/src/database.rs
```diff
@@ -0,0 +1,260 @@
...
+        let block_heights = fjall.keyspace("block_heights", KeyspaceCreateOptions::default)?;
+        let key_images = fjall.keyspace("key_images", KeyspaceCreateOptions::default)?;
+        let pre_rct_outputs = fjall.keyspace("pre_rct_output", KeyspaceCreateOptions::default)?;
+        let tx_ids = fjall.keyspace("tx_ids", KeyspaceCreateOptions::default)?;
+        let v1_tx_outputs = fjall.keyspace("tx_outputs", KeyspaceCreateOptions::default)?;
+
+        let alt_chain_infos = fjall.keyspace("alt_chain_infos", KeyspaceCreateOptions::default)?;
+        let alt_block_heights =
+            fjall.keyspace("alt_block_heights", KeyspaceCreateOptions::default)?;
+        let alt_blocks_info = fjall.keyspace("alt_blocks_infos", KeyspaceCreateOptions::default)?;
```
hinto-janai, line 59:
```suggestion
        let alt_block_infos = fjall.keyspace("alt_block_infos", KeyspaceCreateOptions::default)?;
```

storage/blockchain/src/service/write.rs
```diff
@@ -53,127 +125,85 @@ fn handle_blockchain_request(
...
+    let old_main_chain_id = ChainId(rand::random());
+
+    // pop the blocks
+    for _ in 0..numb_blocks {
+        crate::ops::block::pop_block(db, Some(old_main_chain_id), &mut tx_rw, &mut tapes)?;
     }
+
+    tx_rw.commit()?;
+    tapes.commit(Persistence::SyncAll)?;
+    Ok(BlockchainResponse::PopBlocks(old_main_chain_id))
```
hinto-janai, line 200: Shouldn't `tx_rw` be passed to `flush_alt_blocks`?

storage/blockchain/src/database.rs
```diff
@@ -0,0 +1,260 @@
...
+    types::{Amount, BlockInfo, RctOutput, TxInfo},
+    BlockchainError,
+};
+
+/// The blockchain database.
+pub struct BlockchainDatabase {
+    pub(crate) linear_tapes: Tapes,
+    pub(crate) fjall: fjall::Database,
+
+    pub(crate) block_heights: fjall::Keyspace,
```
hinto-janai, line 20: Optional but perhaps (K,V) and their encoding should be documented since the types/tables are gone, even monerod's DB is easier to understand at a glance now 😭


```suggestion
    /// K = `[u8; 32]` (block hash), V = [`usize`], encoding = little endian.
    pub(crate) block_heights: fjall::Keyspace,
```

binaries/cuprated/src/config.rs
```diff
@@ -305,29 +337,39 @@ impl Config {
...
 
-        // We don't set reader threads as we manually make the reader threadpool.
-        cuprate_blockchain::config::ConfigBuilder::default()
-            .network(self.network)
-            .data_directory(self.fs.data_directory.clone())
-            .sync_mode(blockchain.sync_mode)
-            .build()
+        cuprate_blockchain::config::Config {
+            blob_dir: path_with_network(&self.fs.fast_data_directory, self.network),
+            index_dir: path_with_network(&self.fs.slow_data_directory.clone(), self.network),
```
hinto-janai, line 342: nit:
```suggestion
            index_dir: path_with_network(&self.fs.slow_data_directory, self.network),
```
`cache_sizes` below can be `Copy` but meh

binaries/cuprated/src/config/fs.rs
```diff
@@ -24,7 +24,22 @@ config_struct! {
         /// | Windows | "C:\Users\Alice\AppData\Roaming\Cuprate\"           |
         /// | macOS   | "/Users/Alice/Library/Application Support/Cuprate/" |
         /// | Linux   | "/home/alice/.local/share/cuprate/"                 |
-        pub data_directory: PathBuf,
+        pub fast_data_directory: PathBuf,
```
hinto-janai, line 27: Is there a reason to split this into two? They're both the same path.

storage/txpool/src/service/write.rs
```diff
@@ -72,101 +129,75 @@ fn add_transaction(
...
+    if !tx_info.flags.contains(TxStateFlags::STATE_STEM) {
+        return Ok(TxpoolWriteResponse::Ok);
+    }
+
+    tx_info.flags.remove(TxStateFlags::STATE_STEM);
+
+    db.tx_infos.insert(tx_hash, bytemuck::bytes_of(&tx_info))?;
+
+    if !db.tx_blobs.contains_key(tx_hash)? {
+        db.tx_infos.remove(tx_hash)?;
```
hinto-janai, line 173: Shouldn't this be through a `OwnedWriteBatch`?

hinto-janai:
I still don't have a good mental model of the atomic-ness between fjall/tapes, there are separate transactions everywhere, sometimes directly into a keyspace, sometimes in different commit orders. It doesn't seem like we can have a deterministic safe sync mode (as much as `monerod` is). Is this just a tradeoff being taken?

Boog900:
The DB is now 2 systems that are for ACID purposes independent. The tapes DB will do 1 update atomically, so will fjall, the 2 updates combined are not atomic. 

This does lead to issues with the fact they could get out of sync. On start-up we solve this by using the tapes DB as the source of truth, if fjall is out of sync, we rebuild it from the tapes (doesn't take that long for me). On some requests where we need an fjall and tapes TX to be in sync we should check if they are then just open another set of txs until they are in sync (this is not done yet).

The only exception is alt blocks, for which fjall is not atomic. This is ok as we clear alt blocks on startup anyway and request should be built to handle only some alt block data for a given block being present.

This all means Cuprated should be able to have a safe sync mode.

storage/blockchain/src/types.rs
Boog900: They need to be ordered by integer value. I'll add a comment

storage/blockchain/src/ops/alt_block/tx.rs
Boog900: I fixed the obvious ones I saw, lmk if you see others/

storage/blockchain/src/ops/block.rs
Boog900: I think I changed all of these, lmk if there are others you see

storage/blockchain/src/ops/block.rs
Boog900: Added an assert in `cuprate_blockchain::service::write::pop_blocks`

storage/blockchain/src/ops/alt_block/chain.rs
Boog900: When updating this we need to see updates immediately so we can't use a batch and must write immediately.

storage/blockchain/src/ops/output.rs
Boog900: fixed

storage/blockchain/src/service/write.rs
Boog900: See the other comments on alt blocks

binaries/cuprated/src/config/fs.rs
Boog900: This is a feature I have seen people wanted in monerod, the option to put the data that takes up the most space on a HDD and have the rest on an SSD.

storage/txpool/src/service/write.rs
Boog900: its a single write, its written like this to allow multiple places to write to the pool at the same time, although that is not done yet.

hinto-janai (approved):

PR #592, opened 2026-03-16
cuprated: embeddable node library with `Node::launch()`

### LINKED ISSUE #516: Make cuprate usable as a library
## Feature
Expose cuprate as a library to allow it to be embedded within other Rust projects.

## Why
This could be useful for a variety of projects:
1. Wallets that might want to integrate `cuprate` as its node. I'm working on [eigenwallet](https://github.com/eigenwallet/core) (previously UnstoppableSwap) and we'd like to bundle cuprate to use as a node for the wallet and for atomic swaps.
2. Someone might want to build a simple GUI for cuprate (e.g with [Tauri](https://v2.tauri.app/))
3. Blockexplorers could be built on top of cuprate and avoid the RPC API all together
4. Light wallet syncing servers (like [monero-lws](https://github.com/vtnerd/monero-lws)) might get re-written in Rust at some point in the future. They could bundle cuprate directly.
5. Maybe for running data science experiments on the Blockchain (like what mrl does all the time). Honestly don't know if this could be useful but naively I'd think it could be useful.
Probably a ton of other use cases I cannot think of at the moment.

## Additional context
1. I'd like to be able to spawn cuprate (e.g `cuprate::launch(...)`. The function should take a config struct to configure how cuprate should run.
2. I'd like to be able to listen for status events: Is the node synced? How many network connections is it doing? How many peers do we know of? Not all of this is necessary but would be cool to be able to display that to a user. I like how `arti_client` is exposing a simple stream where I can listen for events: https://docs.rs/arti-client/latest/arti_client/struct.TorClient.html#method.bootstrap_events

redsh4de:
### What

Extracts `cuprated` into an embeddable library (`lib.rs`/`main.rs` pattern). Takes inspiration from the API proposed in #554

Closes #516

Note to reviewers: This does not implement graceful shutdown / panic removal yet, those have been deferred to #585. For the final state with all changes, skip to #590  

### Why

To enable embedding the node in other applications

### Where
`cuprated`:
- `Cargo.toml` - explicit `[lib]` + `[[bin]]` targets
- `lib.rs` (new) - `Node` struct, `LaunchContext` for initialization items, `Node::launch()`
- `main.rs` - slimmed to a CLI wrapper over `Node::launch()`
- `config.rs` - decoupled from CLI args; `find_config()` returns `Result<Option<Config>>`; `dry_run_check()` returns `Vec<DryRunResult>`
- `blockchain.rs` - new `BlockchainInterface` composite handle
- `blockchain/syncer.rs` - `SyncNotify` -> `Syncer` + `SyncerHandle`
- `blockchain/interface.rs` - `COMMAND_TX` / `BLOCKS_BEING_HANDLED` statics -> `BlockchainManagerHandle`
- `signals.rs` and `statics.rs` deleted; state lives in `NodeContext` for multi-instance use
- `version.rs` - `Default` impl

`consensus/fast-sync`:
- `FAST_SYNC_HASHES` static + `set_fast_sync_hashes` removed; `fast_sync_stop_height` and `validate_entries` take hashes as a parameter from LaunchContext

### How

1. `Node::launch(config)` initializes everything and returns the embedder API:

```rust
pub struct Node {
    pub blockchain: BlockchainInterface,
    pub txpool: TxpoolReadHandle,
    pub clearnet: NetworkInterface<ClearNet>,
    pub tor: Option<oneshot::Receiver<NetworkInterface<Tor>>>,
    pub syncer: SyncerHandle,
    pub config: Arc<Config>
}
```

Internal state previously held in mutable statics lives in LaunchContext - the node is now safe to run multiple times in a single process.

2. The new `BlockchainInterface` composes three handles into one type for ergonomics:
```rust
pub struct BlockchainInterface {
    read: BlockchainReadHandle,
    context_svc: BlockchainContextService,
    manager: BlockchainManagerHandle,
}
```
Embedders use the public impl functions `node.blockchain.read()` and `node.blockchain.context()`

3. The Tor interface arrives via oneshot after sync, similar to the `dandelion.rs` pattern.


binaries/cuprated/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// The caller should set up the following before calling this:
+    /// - Tracing/logging (the node emits tracing events during initialization)
+    /// - Global rayon thread pool (optional, uses rayon defaults if not set)
+    ///
+    /// # Panics
+    ///
+    /// Panics if the database is corrupt or critical services fail to start.
+    pub async fn launch(config: Config) -> Self {
+        // Initialize global static `LazyLock` data.
+        statics::init_lazylock_statics();
```
sneurlax: statics::init_lazylock_statics() twice--once in main.rs, once here.  What should actually be responsible for this?

binaries/cuprated/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+        );
+
+        // Start the blockchain & tx-pool databases.
+        let (mut blockchain_read_handle, mut blockchain_write_handle, _) =
+            cuprate_blockchain::service::init_with_pool(
+                config.blockchain_config(),
+                Arc::clone(&db_thread_pool),
+            )
+            .inspect_err(|e| error!("Blockchain database error: {e}"))
+            .expect(DATABASE_CORRUPT_MSG);
```
sneurlax: won't this panic if the db is corrupt, misconfigured, etc? maybe locked?

inspect_err logs it then panics anyways?

binaries/cuprated/src/commands.rs
```diff
@@ -1,25 +1,66 @@
...
+/// The command handler.
+#[derive(Clone)]
+pub struct CommandHandle {
+    tx: mpsc::Sender<CommandRequest>,
+}
+
+impl CommandHandle {
+    /// Initialize the command handler and return a handle.
+    pub fn init(node_ctx: NodeContext) -> Self {
+        let (tx, mut rx) = mpsc::channel::<CommandRequest>(8);
```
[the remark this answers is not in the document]
redsh4de: Ty for the catch, this should be changed to 1 now

The 8 carried over from when i had a seperate buffered output channel, doesnt apply now when a dedicated response channel is made by the request

SyntheticBird45 (changes requested):

SyntheticBird45:
cancelling workflow because this is some doc nits, let's think a little about our carbon footprint.

redsh4de:
Think this is the last change i want to make here, didn't like how `SyncNotify/State` effectively functions as a handle without it being named such, reduces the mental load imo

Also added `target_height` to the handle so that embedders can create their own progress tracking items and whatnots

Boog900 (approved):

SyntheticBird45 (approved):

PR #596, opened 2026-04-07
cuprated: parse SOCKS5 proxy URL

### LINKED ISSUE #579: parse the proxy url in the config
I don't really like how we don't parse the url in the config and leave it as a string but changing that is harder than it seems

_Originally posted by @Boog900 in https://github.com/Cuprate/cuprate/pull/549#pullrequestreview-3779294805_

redsh4de:
### What
Closes #579 

### Why
no unparsed string :)

### Where
cuprated, p2p

### How
Add a `TryFrom<String>` impl that parses "", "Tor", falls through to attempting to parse anything else as a SOCKS5 proxy url, checking if each part is correct
Also `From<ProxySettings>` for `String` for test compatibility

### DIFF
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -224,16 +224,15 @@ config_struct! {
         #[comment_out = true]
         /// The proxy to use for outgoing P2P connections
         ///
-        /// This setting can only take "Tor" at the moment.
-        /// This will anonymise clearnet connections through Tor.
+        /// Setting this to "Tor" will anonymise clearnet connections through Tor.
         ///
         /// Setting this to "" (an empty string) will disable the proxy.
         ///
         /// Enabling this setting will disable inbound connections.
         ///
         /// Type         | String
-        /// Valid values | "Tor"
-        /// Examples     | "Tor"
+        /// Valid values | "Tor", "socks5://[user:pass@]host:port"
+        /// Examples     | "Tor", "socks5://127.0.0.1:9050"
         pub proxy: ProxySettings,
     }
 
@@ -309,7 +308,7 @@ impl Default for ClearNetConfig {
             listen_on: Ipv4Addr::UNSPECIFIED,
             enable_inbound_v6: false,
             listen_on_v6: Ipv6Addr::UNSPECIFIED,
-            proxy: ProxySettings::Socks(String::new()),
+            proxy: ProxySettings::Disabled,
             outbound_connections: 32,
             extra_outbound_connections: 8,
             max_inbound_connections: 128,
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -231,7 +231,7 @@ config_struct! {
         /// Enabling this setting will disable inbound connections.
         ///
         /// Type         | String
-        /// Valid values | "Tor", "socks5://[user:pass@]host:port"
+        /// Valid values | "Tor", "socks5://host:port", "socks5://user:pass@host:port"
         /// Examples     | "Tor", "socks5://127.0.0.1:9050"
         pub proxy: ProxySettings,
     }
--- a/binaries/cuprated/src/config/p2p.rs
+++ b/binaries/cuprated/src/config/p2p.rs
@@ -231,7 +231,7 @@ config_struct! {
         /// Enabling this setting will disable inbound connections.
         ///
         /// Type         | String
-        /// Valid values | "Tor", "socks5://host:port", "socks5://user:pass@host:port"
+        /// Valid values | "Tor", "socks5://ip:port", "socks5://user:pass@ip:port"
         /// Examples     | "Tor", "socks5://127.0.0.1:9050"
         pub proxy: ProxySettings,
     }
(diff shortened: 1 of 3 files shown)


binaries/cuprated/src/p2p.rs
```diff
@@ -48,33 +48,60 @@ pub use network_address::CrossNetworkInternalPeerId;
...
-                    .map(|(a, b)| (a.to_string(), b.to_string())),
-                ad,
-            )
-        })
-        .unwrap_or((None, url));
+    fn try_from(s: String) -> Result<Self, Self::Error> {
+        match s.as_str() {
+            "" => Ok(Self::Disabled),
+            "Tor" => Ok(Self::Tor),
+            url => {
```
SyntheticBird45: Could you do something like `url if s.starts_with(''socks5://')` instead ? So we can have a `_` matching and return an error on unknown scheme.

binaries/cuprated/src/config/p2p.rs
```diff
@@ -218,16 +218,15 @@ config_struct! {
...
+        /// Setting this to "Tor" will anonymise clearnet connections through Tor.
         ///
         /// Setting this to "" (an empty string) will disable the proxy.
         ///
         /// Enabling this setting will disable inbound connections.
         ///
         /// Type         | String
-        /// Valid values | "Tor"
-        /// Examples     | "Tor"
+        /// Valid values | "Tor", "socks5://host:port", "socks5://user:pass@host:port"
```
SyntheticBird45: We do not have host resolution. Replace `host` with `ip` instead.

binaries/cuprated/src/p2p.rs
```diff
@@ -48,33 +48,60 @@ pub use network_address::CrossNetworkInternalPeerId;
...
-        .map(|(up, ad)| {
-            (
-                up.split_once(':')
-                    .map(|(a, b)| (a.to_string(), b.to_string())),
-                ad,
-            )
-        })
-        .unwrap_or((None, url));
+    fn try_from(s: String) -> Result<Self, Self::Error> {
+        match s.as_str() {
```
SyntheticBird45, line 62: Should use `trim` here I think.

SyntheticBird45 (changes requested):

binaries/cuprated/src/p2p.rs
redsh4de: not necessary imo - since the values need to be in quotes, it would be a issue if the user types in " Tor " with the spaces in the config, user error

binaries/cuprated/src/p2p.rs
redsh4de: Pushed with a TODO, after we are on Rust >1.95 would use the if let guard for a clean one-liner:
```rust
url if let Some(url) = url.strip_prefix("socks5://") => {
    ...
}
_ => Err(anyhow!("Unsupported proxy: '{s}'")),
```

binaries/cuprated/src/p2p.rs
SyntheticBird45: typos can happen fast. I'm more encouraging for small details that make user's life easier. If they accidentally left leading and trailing whitespaces i wouldn't want it to cause any confusion. If we're capable of understanding through mistakes then we should do it.

I can imagine many users frustrated with the following:
- `" socks5://user:pass@ip:port"` => `Error: Unsupported proxy:  socks5://user:pass@ip:port`
- `"socks5://user:pass@ip:port "` => `Error: invalid IPv4 socket address syntax`

I would even tend to say we should lowercase the whole input to support `tor` instead of `Tor`.
cc @Boog900

SyntheticBird45 (commented):
~~waiting on @Boog900 for one of the requested change. Will approve once it is resolved.~~

Edit: resolved

binaries/cuprated/src/p2p.rs
redsh4de: I agree with making the Tor matching case-insensitive 👍

My only nit with trimming here, despite it's benefits, is that it's not done anywhere else. ex: `network = " Mainnet"` with a leading whitespace fails as well

Everything in the config is loaded as is atm - so while it would be more user friendly, i believe trimming would be best applied for all config paths as a seperate PR, if needed

SyntheticBird45: So after discussion we both agree with you. So an issue will be open regarding trimming and *casing. marking as resolved. sorry for the wait

SyntheticBird45 (approved):

binaries/cuprated/src/p2p.rs
```diff
@@ -48,33 +48,60 @@ pub use network_address::CrossNetworkInternalPeerId;
...
+            "" => Ok(Self::Disabled),
+            "Tor" => Ok(Self::Tor),
+            // TODO: use `if let` guard when on >=1.95
```
SyntheticBird45, line 65: Image

https://blog.rust-lang.org/2026/04/16/Rust-1.95.0/

>  **if-let guards in matches**
>
> Rust 1.88 stabilized [let chains](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains). Rust 1.95 brings that capability into match expressions, allowing for conditionals based on pattern matching.

redsh4de: wen `workspace: Rust 1.93..=1.95` PR

Boog900 (approved):

PR #597, opened 2026-04-09
cuprated: broadcast block before notifying txpool

### LINKED ISSUE #541: Broadcast new blocks before removing txs from the pool
## What
Broadcast blocks to the network before we tell the tx-pool about the new block.

## Why
To speed up block propagation 

## Where
cuprated

redsh4de:
### What
Closes #541 

### Why
Faster block propagation

### Where
cuprated

### How
Move the `block_blob` extraction and `broadcast_block` call inside `add_valid_block_to_main_chain`, between the DB write and the txpool notification. Add a `BlockSource` parameter naming each of the three call paths (Incoming, BatchSync, Reorg) - only `Incoming` triggers the broadcast.

### DIFF
--- a/binaries/cuprated/src/blockchain/manager/handler.rs
+++ b/binaries/cuprated/src/blockchain/manager/handler.rs
@@ -148,15 +148,8 @@ impl super::BlockchainManager {
         )
         .await?;
 
-        let block_blob = Bytes::copy_from_slice(&verified_block.block_blob);
-        self.add_valid_block_to_main_chain(verified_block).await;
-
-        let chain_height = self
-            .blockchain_context_service
-            .blockchain_context()
-            .chain_height;
-
-        self.broadcast_block(block_blob, chain_height).await;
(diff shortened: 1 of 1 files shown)


binaries/cuprated/src/blockchain/manager/handler.rs
```diff
@@ -693,6 +698,25 @@ impl super::BlockchainManager {
...
+    ) {
+        self.blockchain_write_handle
+            .ready()
+            .await
+            .expect(PANIC_CRITICAL_SERVICE_ERROR)
+            .call(BlockchainWriteRequest::WriteBlock(verified_block))
+            .await
+            .expect(PANIC_CRITICAL_SERVICE_ERROR);
+    }
+
```
Boog900, line 719: Why was this moved to a separate function?

redsh4de: purely for symmetry's sake, as there already was a "add to cache" function and a "batch-add to db" function

Boog900 (approved):

PR #599, opened 2026-04-11
storage: check alt blocks when handling missing tx request in DB

### LINKED ISSUE #580: Check alt blocks when handling missing tx request in DB

redsh4de:
### What
Support alt blocks in `TxsInBlock`

Closes #580

### Why
`txs_in_block` only searches the main chain at the moment. A peer can request missing transactions for a block that is on an alt chain - either from a reorg or because Cuprate broadcasts alt blocks as of #577

### Where
`storage/blockchain`

### How
- Rename existing `get_alt_block` to `get_alt_block_information` so it's more honest about it's return type (`AltBlockInformation`)
- New `get_alt_block` mirrors the main-chain equivalent (returns `Block`, is used in `txs_in_block`)
- Extend `txs_in_block` to try the main chain first, then fall back to alt-chain tables on `KeyNotFound`
- add and use a `alt_block_height` helper to DRY

### DIFF
--- a/storage/blockchain/src/ops/alt_block/block.rs
+++ b/storage/blockchain/src/ops/alt_block/block.rs
@@ -94,7 +94,7 @@ pub fn add_alt_block(
 ///
 /// This function will look at only the blocks with the given [`AltBlockHeight::chain_id`], no others
 /// even if they are technically part of this chain.
-pub fn get_alt_block(
+pub fn get_alt_block_information(
     db: &BlockchainDatabase,
     alt_block_height: &AltBlockHeight,
     tx_ro: &fjall::Snapshot,
@@ -132,6 +132,22 @@ pub fn get_alt_block(
         ),
         chain_id: alt_block_height.chain_id.into(),
     })
+}
+
+/// Retrieve an alt [`Block`] via its [`AltBlockHeight`].
+///
+/// This function will look at only the blocks with the given [`AltBlockHeight::chain_id`], no others
+/// even if they are technically part of this chain.
+pub fn get_alt_block(
+    db: &BlockchainDatabase,
+    alt_block_height: &AltBlockHeight,
+    tx_ro: &fjall::Snapshot,
+) -> DbResult<Block> {
+    let block_blob = tx_ro
+        .get(&db.alt_block_blobs, bytemuck::bytes_of(alt_block_height))?
+        .ok_or(BlockchainError::NotFound)?;
+
+    Ok(Block::read(&mut block_blob.as_ref()).unwrap())
 }
 
 /// Retrieves the hash of the block at the given `block_height` on the alt chain with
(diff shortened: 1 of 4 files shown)


storage/blockchain/src/ops/alt_block/block.rs
```diff
@@ -134,6 +135,23 @@ pub fn get_alt_block(
...
+pub fn get_alt_block_blob_with_tx_hashes(
+    db: &BlockchainDatabase,
+    alt_block_height: &AltBlockHeight,
+    tx_ro: &fjall::Snapshot,
+) -> DbResult<(Vec<u8>, Vec<TxHash>)> {
+    let block_blob = tx_ro
+        .get(&db.alt_block_blobs, bytemuck::bytes_of(alt_block_height))?
+        .ok_or(BlockchainError::NotFound)?;
+
+    let block = Block::read(&mut block_blob.as_ref()).unwrap();
```
vijaydasmp: I might be mistaken here, but using unwrap() could potentially panic if the block data is malformed

redsh4de: Mirrored the precedent set in https://github.com/Cuprate/cuprate/blob/main/storage/blockchain/src/ops/alt_block/block.rs#L112

makes sense to me though, if something is corrupted - somethings gone really wrong, panic's appropriate

redsh4de:
this was made before the tapes db upgrade, aligned it just now

- renamed the prior `get_alt_block` to `get_alt_block_information` so it's more honest about its return type (returns `AltBlockInformation`)
- `get_alt_block` now is basically a mirror of `get_block`: also returns `Block`, and is used when looking up the missing txes

storage/blockchain/src/service/read.rs
```diff
@@ -908,36 +901,63 @@ fn find_first_unknown(db: &BlockchainDatabase, block_ids: &[BlockHash]) -> Respo
...
 
-    let block = get_block(&block_height, None, &tapes, db)?;
-    let first_tx_index = block_info.mining_tx_index + 1;
+        if block.transactions.len() < missing_txs.len()
+            || missing_txs
+                .iter()
+                .any(|&i| i >= usize_to_u64(block.transactions.len()))
+        {
+            return Ok(BlockchainResponse::TxsInBlock(None));
+        }
```
Boog900, line 925: This check is duplicated can we move it up to de-duplicate it?

redsh4de: Something like this? Nested helper function:

```rust
fn is_tx_range_valid<T>(transactions: &[T], missing_txs: &[u64]) -> bool {
    let tx_len = transactions.len();
    tx_len >= missing_txs.len()
        && !missing_txs
            .iter()
            .any(|&i| i >= usize_to_u64(tx_len))
}
```

Boog900: ah no, I am silly, we need the `block`

Boog900 (approved):

PR #605, opened 2026-05-01
consensus: pass spent KIs to cuprated

redsh4de:
### What
Addresses the FIXME in `add_valid_block_to_main_chain`.

### Why
Block verification in consensus walks every transaction's inputs to dedupe and check key images. After verification, cuprated walked them a second time to notify the txpool.

This makes it so that the KI vec is returned from consensus, eliminating the second walk.

### Where
consensus, helper, cuprated

### How
In `helper`:
- New `tx_key_images()` helper for extracting a transactions key images

In `consensus`:
- Batch preparation now returns `TxsWithKis` per block (transactions + validated key images).
- The verify functions now return `VerifiedBlock` (wrapper for verified block info + the spent key images).
- `verify_prepped_main_chain_block` takes a `PreparedBlockTxs::{Checked, Unchecked}` enum so the caller can say whether key images still need checking:
  -  `Unchecked(Vec<TransactionVerificationData>)` - single-block path. Extracts KIs and checks.
  - `Checked(TxsWithKis)` - batch-sync path. `TxsWithKis` is only produced by `batch_prepare_main_chain_blocks` where validation was already done, so we skip checking here and use the passed KIs as is.
- Split `FullVerification` into two methods, so the txpool path doesn't allocate a key-image Vec it would discard:
  - `verify` (txpool): returns just the transactions.
  - `verify_with_kis` (block, internal): returns transactions and key images.
- Key-image checking seperated into three:
  - `check_kis`: used by the batch & txpool paths. Takes pre-computed key images.
  - `extract_and_check_kis`: single-block path. Walks inputs once and builds both the KI dedup set and Vec in one pass.
  - `check_kis_unspent`: shared chain-spent db query, skipped on a empty set.
- `key_images_spent_checked` bool removed from `BatchPrepareCache`. The existence of `TxsWithKis` means that the key images were checked, so this becomes redundant.
- Capacity hint change: `check_kis_unique`'s old `with_capacity(size_hint().1.unwrap_or(0) * 2)` always resolved to zero for the batch caller, because [when `flat_map` is used on a slice iter, the upper-bound size hint is always `None`](https://users.rust-lang.org/t/how-can-i-make-flatmap-size-hint-use-my-iterators-size-hint/112891/2). To avoid rehashing, we count the KIs once up front and pass capacity as a param.
- New `alt_block_to_verified_block` replaces cuprated's `alt_block_to_verified_block_information`. Computes `total_fees` and `spent_key_images` in one pass, then constructs a `VerifiedBlock`.

In `cuprated`:
- `add_valid_block_to_main_chain` splits the new `VerifiedBlock` into block information and key images, used like before.
- The reorg-reverse path now calls the new `alt_block_to_verified_block` in consensus; the local helper is removed.

### DIFF
--- a/binaries/cuprated/src/blockchain/manager/handler.rs
+++ b/binaries/cuprated/src/blockchain/manager/handler.rs
@@ -3,10 +3,7 @@ use std::{collections::HashMap, sync::Arc};
 
 use bytes::Bytes;
 use futures::{TryFutureExt, TryStreamExt};
-use monero_oxide::{
-    block::Block,
-    transaction::{Input, Transaction},
-};
+use monero_oxide::block::Block;
 use rayon::prelude::*;
 use tower::{Service, ServiceExt};
 use tracing::{info, instrument, warn, Span};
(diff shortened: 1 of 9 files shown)


redsh4de:
Diagram of how key images flow and where checks happen on each path in this PR.
It can be seen that with this, we have a guarantee that we will only extract key images once at the earliest possible step for each path, and get them passed down to the block handler

image

PR #606, opened 2026-05-03
cuprated: add private file permissions for Windows

### LINKED ISSUE #551: Private file permissions by default

redsh4de:
### What
Closes #551 by adding the Windows equivalent.

### Why
In cases where the data paths were outside of the current user's profile directory, files in cuprate's data directories could be read by other users on the same Windows machine.

### Where
`helper`
  - `fs/windows_perms.rs` - Looks up the current user's ID, builds the ACL, makes it the default for files cuprate creates, and makes each folder it creates with it.
  - `fs.rs` - `set_private_global_file_permisisons` is now Unix-only (Windows doesn't really have a 1-to-1 global equivalent). A new Windows-only `set_private_directory_permissions(roots: &[&Path])` passes the folder paths to `windows_perms::apply`. Each function only exists on the OS that needs it
  - `Cargo.toml` - added new create features from `windows` that are needed for this to work

`cuprated`
  - `config.rs` - new `Config::writable_directories()` returns the directories cuprate writes to. On Windows only, `read_config_and_args` calls `set_private_directory_permissions` once the paths are resolved. This also gave a opportunity to rewrite `check_dir_permissions` to loop over the new helper.
  - `main.rs` - `set_private_global_file_permissions` call is now Unix-only

### How
The SDDL string is `O:{user}D:P(A;OICI;FA;;;{user})(A;OICI;FA;;;SY)(A;OICI;FA;;;BA)` ([syntax reference](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-string-format)):
| Part | Meaning |
|------|---------|
| `O:{user}` | Owner is the user 
| `D:P` | Start of the access list, with the "protected" flag so it does not inherit from the parent folder |
| `(A;OICI;FA;;;{user})` | Allow the user full access; sub-folders and files inherit it |
| `(A;OICI;FA;;;SY)` | Same for `SYSTEM` |
| `(A;OICI;FA;;;BA)` | Same for `BUILTIN\Administrators` |

Precedent: [Win32-OpenSSH](https://github.com/PowerShell/Win32-OpenSSH/wiki/Security-protection-of-various-files-in-Win32-OpenSSH)

---

Example with a path like `D:\Cuprate`:
image

### DIFF
--- a/helper/src/fs/windows_perms.rs
+++ b/helper/src/fs/windows_perms.rs
@@ -1,0 +1,158 @@
+//! Windows ACL implementation for [`super::set_private_directory_permissions`].
+
+//---------------------------------------------------------------------------------------------------- Use
+use std::ffi::OsStr;
+use std::os::windows::ffi::OsStrExt;
+use std::path::Path;
+use std::ptr;
+
+use target_os_lib::core::{Error, Owned, Result, PCWSTR, PWSTR};
+use target_os_lib::Win32::Foundation::{
+    ERROR_ALREADY_EXISTS, ERROR_INSUFFICIENT_BUFFER, E_UNEXPECTED, HANDLE, HLOCAL,
+};
+use target_os_lib::Win32::Security::Authorization::{
+    ConvertSidToStringSidW, ConvertStringSecurityDescriptorToSecurityDescriptorW, SDDL_REVISION_1,
+};
+use target_os_lib::Win32::Security::{
+    GetTokenInformation, TokenUser, PSECURITY_DESCRIPTOR, SECURITY_ATTRIBUTES, TOKEN_ACCESS_MASK,
+    TOKEN_QUERY, TOKEN_USER,
+};
+use target_os_lib::Win32::Storage::FileSystem::CreateDirectoryW;
+use target_os_lib::Win32::System::Threading::{GetCurrentProcess, OpenProcessToken};
+
+//---------------------------------------------------------------------------------------------------- SecurityDescriptor
+/// Security descriptor parsed from SDDL, with its `LocalAlloc` buffer freed on drop.
+struct SecurityDescriptor {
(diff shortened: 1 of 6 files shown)


helper/src/fs/windows_perms.rs
```diff
@@ -0,0 +1,141 @@
...
+            _buf: unsafe { Owned::new(HLOCAL(psd.0)) },
+        })
+    }
+}
+
+//---------------------------------------------------------------------------------------------------- Apply
+/// Apply a private ACL to each path in `roots`.
+///
+/// SYSTEM and Administrators are granted access alongside the user so
+/// Windows backup, antivirus, and indexer services keep working.
```
SyntheticBird45, line 56: nit: Not sure what the meaning of these last words are. It's not like a user can revoke SYSTEM access, prevent Administrator from regaining ownership, or antivirus not giving a damn f*ck about permissions anyway.

SyntheticBird45 (commented):
I'll review the windows API usage more in details tomorrow. Nothing else to comment outside of it.

helper/src/fs/windows_perms.rs
redsh4de: its moreso it explain why the SDDL string contains SY and BA as it isn't obvious at first glance without context

PR #607, opened 2026-05-05
add regtest/fakechain

Boog900:
Adds a regtest mode that matches monerod's regtest.

Useful for testing.

### DIFF
--- a/consensus/rules/src/hard_forks.rs
+++ b/consensus/rules/src/hard_forks.rs
@@ -34,15 +34,61 @@ pub fn check_block_version_vote(
     Ok(())
 }
 
+/// The amount of time in seconds after the last fork's scheduled timestamp before we
+/// consider this node likely forked from the network.
+///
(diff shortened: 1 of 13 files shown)


consensus/rules/src/hard_forks.rs
```diff
@@ -59,27 +105,46 @@ impl HFsInfo {
...
+        let last_fork_time = self.0[NUMB_OF_HARD_FORKS - 1].time;
+        if current_time >= last_fork_time.saturating_add(FORKED_TIME) {
+            HardForkState::LikelyForked
+        } else if current_time >= last_fork_time.saturating_add(UPDATE_TIME) {
+            HardForkState::UpdateNeeded
+        } else {
+            HardForkState::Ready
+        }
+    }
+
```
SyntheticBird45, line 126: In preparation of FCMP++ I presume ?

Boog900: No this is an old system for when monerod had HFs every 6 months to combat ASICs. We need it as monerod still exposes this in the RPC.

SyntheticBird45 (approved):

PR #613, opened 2026-05-11
fix get_hashes + make it faster with 1 lookup

Boog900:


binaries/cuprated/src/rpc/service/blockchain.rs
```diff
@@ -136,10 +135,7 @@ pub async fn next_chain_entry(
     } = blockchain_read
         .ready()
         .await?
-        .call(BlockchainReadRequest::NextChainEntry(
-            block_hashes,
-            u64_to_usize(start_height),
-        ))
+        .call(BlockchainReadRequest::NextChainEntry(block_hashes, 10_000))
```
redsh4de, line 138: nit: the 10_000 could be put in cuprate-constants to mirror monerod: https://github.com/monero-project/monero/blob/master/src/cryptonote_config.h#L96

p2p/request_handler.rs could too make use of it then for consistency

redsh4de (approved):

binaries/cuprated/src/rpc/service/blockchain.rs
Boog900: I don't think I want to do this as I do want to look into changing 10,000 for RPC

PR #619, opened 2026-05-24
fix: flush non-blocking log appender on shutdown instead of mem::forget

orrinfrazier:
## Summary
- `init_logging` leaked the `tracing_appender` `WorkerGuard` via `mem::forget`, so buffered log lines were never guaranteed to flush on shutdown — losing the final lines that matter for crash/shutdown post-mortems.
- Return the `WorkerGuard` from `init_logging` (now `#[must_use]`) and hold it in `main()` for the program lifetime, so it drops — and flushes the file appender — after the final shutdown log line.

## Changes
- **`logging.rs`**: extract the appender construction into a private `create_file_appender(config) -> (NonBlocking, WorkerGuard)`; change `init_logging` to return `WorkerGuard`; remove `mem::forget(guard)` and its import.
- **`main.rs`**: bind `let _log_flush_guard = init_logging(&config);` before the tokio runtime so it drops last on scope exit (after `drop(rt)` and `info!("Shutdown complete.")`).
- **`lib.rs`**: update the doc example to hold the guard.
- Add a unit test (`file_appender_flushes_on_guard_drop`) that writes a line, drops the guard, and asserts the line was flushed to a temp log file.

## Testing
- `cargo test -p cuprated --lib` — 12 passed, 0 failed (includes the new flush test)
- `cargo clippy -p cuprated --all-targets -- -D warnings` — clean
- `cargo fmt -p cuprated -- --check` — clean
- `cargo build --bin cuprated` — builds

## Notes
- No new dependencies (`tempfile` was already a dev-dependency, `tracing-appender` already a dependency).
- The guard is intentionally held in `main` rather than `Node`: `init_logging` runs before `Node::launch`, and the terminal `Shutdown complete.` line is emitted in `main` after the node/runtime drop, so the guard must outlive `Node`.

Refs: `issues/S10-1-log-flush-shutdown.md`

### DIFF
--- a/binaries/cuprated/src/lib.rs
+++ b/binaries/cuprated/src/lib.rs
@@ -9,7 +9,8 @@
 //! use cuprated::{config::Config, Node};
 //!
 //! let config = Config::read_from_path("cuprated.toml")?;
-//! cuprated::logging::init_logging(&config);
+//! // Hold the guard for the lifetime of the program; dropping it flushes the log file on shutdown.
+//! let _log_flush_guard = cuprated::logging::init_logging(&config);
 //!
 //! let mut node = Node::launch(config).await;
 //! let height = node.blockchain.context().chain_height;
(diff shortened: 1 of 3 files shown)


orrinfrazier:
Opened against upstream by mistake; intended for my fork. Closing.

PR #625, opened 2026-06-02
RPC: add missing DB request for get_info

Boog900:


storage/blockchain/src/service/read.rs
```diff
@@ -804,7 +804,31 @@ fn block_by_hash(db: &BlockchainDatabase, block_hash: BlockHash) -> ResponseResu
...
+    entries
+        .filter_map(Result::ok)
+        .map(|e| {
+            let p = e.path();
+            if p.is_dir() {
+                dir_size(&p)
+            } else {
+                e.metadata().map_or(0, |m| m.len())
+            }
+        })
```
redsh4de: duplicate function of the below one, CI cryin about it

sidenote: this will recurse into symlink dirs - best case the size will be wrong, worst case a recursion loop if the user makes a symlink that points to a parent dir in the same drive

could match on the file_type and only take action on real directories and files, while letting the rest resolve to 0

```suggestion
        .map(|e| match e.file_type() {
            Ok(ft) if ft.is_dir() => dir_size(&e.path()),
            Ok(ft) if ft.is_file() => e.metadata().map_or(0, |m| m.len()),
            _ => 0,
        })
```

redsh4de (changes requested):

storage/txpool/src/service/read.rs
```diff
@@ -263,7 +263,19 @@ fn size(
...
     include_sensitive_txs: bool,
 ) -> Result<TxpoolReadResponse, TxPoolError> {
-    Ok(TxpoolReadResponse::Size(todo!()))
+    let count = if include_sensitive_txs {
+        db.tx_infos.len()?
+    } else {
+        let mut n = 0_usize;
+        for guard in db.tx_infos.iter() {
+            let info: TransactionInfo = bytemuck::pod_read_unaligned(guard.value()?.as_ref());
+            if !info.flags.contains(TxStateFlags::STATE_STEM) {
```
redsh4de: can use the existing `.private()` helper on flags & remove the TxStateFlags import

```suggestion
            if !info.flags.private() {
```

redsh4de (approved):
one nit rest lgtm

PR #626, opened 2026-06-02
cuprated: rename, move SyncerHandle to BlockchainInterface

redsh4de:
### What
moves SyncerHandle to be accessable from BlockchainInterfac

Renamed Syncer to BlockchainSyncer, to match how other items are prefixed with `Blockchain`

Have BlockchainSyncer be constructed inside of the blockchain manager spawn function

### Why
Symmetry. Syncer is spawned by the blockchain module, so it makes sense for it's handle to join the rest of the blockchain handle crew. Same way how `syncer.rs` is located in `blockchain/`

### Where
`cuprated`

### How
clone SyncerHandle into BlockchainInterface at construction and remove the field from Node. Embedders now should call `node.blockchain.syncer()` to get a clone of the handle. Longer but reads nicer too imo

### DIFF
--- a/binaries/cuprated/src/blockchain.rs
+++ b/binaries/cuprated/src/blockchain.rs
@@ -44,6 +44,8 @@ pub struct BlockchainInterface {
     context_svc: BlockchainContextService,
     /// A handle to the blockchain manager.
     manager: BlockchainManagerHandle,
+    /// A handle to the blockchain syncer.
+    syncer: SyncerHandle,
 }
 
 impl BlockchainInterface {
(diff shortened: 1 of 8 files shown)


SyntheticBird45 (approved):

binaries/cuprated/src/blockchain.rs
```diff
@@ -74,10 +78,21 @@ impl BlockchainInterface {
         self.manager.clone()
     }
 
+    /// Returns a handle to the blockchain syncer.
+    pub fn syncer(&self) -> BlockchainSyncerHandle {
+        self.syncer.clone()
+    }
+
```
Boog900, line 85: does this need to return an owned handle, could it return a ref instead?

redsh4de: Think so, but then wed have to also specify lifetimes and the blast radius would be kinda messy

& for symmetry would be nicer if the rest are made to be refs too, otherwise it'd be the odd one out

Boog900: Fair ig it's not too bad

Boog900 (approved):

PR #631, opened 2026-06-08
RPC: Enable more requests

Boog900:


binaries/cuprated/src/rpc/handlers/json_rpc.rs
```diff
@@ -878,10 +898,10 @@ async fn get_version(
     mut state: CupratedRpcHandler,
     _: GetVersionRequest,
 ) -> Result<GetVersionResponse, Error> {
-    let current_height = helper::top_height(&mut state).await?.0;
-    let target_height = blockchain_manager::target_height(todo!()).await?;
+    let current_height = helper::top_height(&mut state).await?.0 + 1;
```
redsh4de: could use `blockchain::chain_height` directly if we wanna avoid the `+ 1`, helper uses it under the hood anyway

```suggestion
    let current_height = blockchain::chain_height(&mut state.blockchain_read).await?.0;
```

binaries/cuprated/src/rpc/handlers/json_rpc.rs
```diff
@@ -859,8 +879,8 @@ async fn get_coinbase_tx_sum(
         .await?;
 
     // Formats `u128` as hexadecimal strings.
-    let wide_emission_amount = fee_amount.hex_prefix();
-    let wide_fee_amount = emission_amount.hex_prefix();
+    let wide_emission_amount = emission_amount.hex_prefix();
+    let wide_fee_amount = fee_amount.hex_prefix();
```
redsh4de:
```suggestion
    let wide_emission_amount = (emission_amount, emission_amount_top64).hex_prefix();
    let wide_fee_amount = (fee_amount, fee_amount_top64).hex_prefix();
```

redsh4de (changes requested):

binaries/cuprated/src/rpc/handlers/json_rpc.rs
Boog900: good catch

binaries/cuprated/src/rpc/handlers/json_rpc.rs
redsh4de, line 872: monerod has a range guard on the equivalent fn

```suggestion
) -> Result<GetCoinbaseTxSumResponse, Error> {
    let chain_height = blockchain::chain_height(&mut state.blockchain_read).await?.0;
    if request.height >= chain_height || request.count > chain_height - request.height {
        return Err(anyhow!("requested range exceeds blockchain height"));
    }
    
```

redsh4de (approved):
one thing rest lgtm

PR #633, opened 2026-06-09
consensus: parity patches

redsh4de:
### What
consensus behaviour fixes, more info about each in each commit's message, updated the book accordingly

### Why
parity w/ core

### Where
`consensus`

### DIFF
--- a/books/protocol/src/consensus_rules/blocks.md
+++ b/books/protocol/src/consensus_rules/blocks.md
@@ -82,7 +82,7 @@ hard-fork[^version-vote].
 The block's timestamp must not be more than the current UNIX time + 2 hours[^timestamp-upper-limit] and the timestamp
 must not be less than
 the median timestamp over the last 60 blocks[^timestamp-lower-limit], if there are less than 60 blocks in the chain then
-the timestamp is always valid.
+this lower limit is ignored.
 
 ---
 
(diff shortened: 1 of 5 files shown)


Boog900:
Thanks for these. Luckily none of these should cause issues on mainnet as they are to do with past consensus rules.

For alt blocks does it matter that we do a full timestamp check? It means we could catch a bad alt block earlier than monerod which should be a good thing?

redsh4de:
I dont think it matters beyond us just not banning an honest peer over an alt block monerod would accept - divergence here cant lead to a chain split, since the alt chain would be reverified on reorg in both cases

worth keeping the patch for equivalent behaviour & doc accuracy imo (since iirc the book is meant to be a reference for both monerod and cuprated behaviour)

Boog900:
monerod wont relay alt blocks so the banning should not be an issue.

As for the book, in an ideal world alt blocks should be checked to the same standard as main chain blocks. Any divergence from that is an implementation choice rather than consensus.

redsh4de:
yeah ig if the final outcome is equivalent it is a implementation detail

in this case what about folding `_median` back into `check_timestamp` and just documenting the difference?

Boog900:
yeah that sounds good

Boog900 (approved):

PR #634, opened 2026-06-10
RPC: add get_output_distribution

Boog900:


redsh4de (approved):
all lgtm 🚀

we prob need a cache for `output_distribution()` so we dont re-iterate over all `block_infos` from HFV4 for each request, but thats for a different PR

Boog900:
I agree.

PR #638, opened 2026-06-11
consensus: context cache for cumulative RCT outputs

### LINKED ISSUE #635: Cache cumulative_rct_outs for OutputDistribution
> we prob need a cache for `output_distribution()` so we dont re-iterate over all `block_infos` from HFV4 for each request

_Originally posted by @redsh4de in https://github.com/Cuprate/cuprate/pull/634#pullrequestreview-4470600155_

redsh4de:
### What
Adds an in-memory rolling cache for cumulative RCT output counts, used by get_output_distribution for amount 0 requests.

Closes #635 

### Why
Public nodes that get frequent transaction requests benefit from this, as they don't have to rescan `block_infos` every time. Also curbs a potential DoS vector 

### Where
`consensus/context`:
  - `src/distribution.rs` - new module: the cache, the request handler, the slicing/delta math, and the per-block RCT count.
  - `src/lib.rs` - new `RctOutputDistribution` request/response. `numb_rct_outputs` field on `NewBlockData`.
  - `src/task.rs` - build the cache at startup, push on update, truncate on pop, serve the request.

`types`:
  - `src/types.rs` - `OutputDistributionInput` is replaced by `PreRctOutputDistributionInput` using non-zero amounts.
  - `src/blockchain.rs` - renamed the `OutputDistribution` read request to `PreRctOutputDistribution` (now pre-RCT only), added `CumulativeRctOutsInRange`.

`storage/blockchain`:
  - `src/service/read.rs` - renamed the `output_distribution` handler to `pre_rct_output_distribution` and narrowed it to pre-RCT amounts, added `cumulative_rct_outs_in_range`.

`binaries/cuprated`:
  - `src/rpc/handlers/shared.rs` - split RCT and pre-RCT amounts, route RCT to the context and pre-RCT to the DB.
  - `src/rpc/service/blockchain.rs` - renamed the old DB wrapper to `pre_rct_output_distribution` and added `chain_height`.
  - `src/rpc/service/blockchain_context.rs` - new `rct_output_distribution` wrapper.
  - `src/blockchain/manager/handler.rs` - set `numb_rct_outputs` on the new block.

### How
The cache keeps the cumulative RCT output counts per block, built once at startup from the database and kept up to date by the context task - each new block pushes its total, a pop truncates from the end (same as the difficulty and weight caches).

Requests are answered by slicing the cached prefix: the amount 0 (RCT) distribution comes straight from the cache for any `(from, to)` range. Pre-RCT amounts are forwarded to the database and merged back in request order.


Boog900:
Is it possible to change this so the context cache only handles the RCT case and the rpc forwards to the DB itself rather than the cache doing it?

redsh4de:
Yeah, will update with that shape in a bit

redsh4de:
Updated

Boog900 (approved):

storage/blockchain/src/service/read.rs
Boog900:
```suggestion
    Chain, ChainId, ExtendedBlockHeader, PreRctOutputDistributionInput, TransactionBlobs, TxInBlockchain,
```

PR #642, opened 2026-06-13
cuprated: trim whitespace and ignore ASCII case in config strings

### LINKED ISSUE #598: Trimming from config file and lowercase matching
## What
1. Strings defined in config files should be trimmed to avoid confusion on errors.
2. We shouldn't depend on a particular casing for strings.

## Why

Regarding 1. As explained in https://github.com/Cuprate/cuprate/pull/596#discussion_r3050155097

Whitespace typos can happen unexpectedly, Most of the time it's because of the user, but sometimes the software/hardware is at fault. On linux, laptop keyboards can register multiple time a key if battery is short.

I believe this is something Cuprate can entirely ignore. In case, this isn't ignored, whitespace typos can cause immense confusion (which i have experienced twice).
Providing the example I gave from the comment above:
 
`proxy = " socks5://user:pass@ip:port"` => `Error: Unsupported proxy:  socks5://user:pass@ip:port`
`proxy = "socks5://user:pass@ip:port "` => `Error: invalid IPv4 socket address syntax`

If this is 3 AM and you are tired, you will read this and assume the software you are using is complete shit. Most will just go to sleep deceived and feeling betrayed. Some will smash their keyboard against their desk. This is dangerous. But we can avoid these injuries.

Regarding 2.
I don't have a lot of argument. From my experience, this is very uncommon to have configuration or string argument requiring an uppercase character. The way we are requiring this is through the use of serde where we need to define the enum variant name. Which in Rust is by convention (and annoyingly warned) as upper camel case. Conceptually users are not there to write struct or enum names.

I'll add that shift typo can also happen tho much more rare than whitespace.

## How
Describe how the proposal could be implemented.

MavenRain:
### What
Closes #598.

Two changes to how `cuprated` reads string values from the config file:

1. Leading/trailing whitespace in string values is trimmed before parsing.  A warning naming the exact key is printed for each trimmed value, e.g.:
   ```
   Warning: ignoring leading/trailing whitespace in config value `p2p.clear_net.proxy`
   ```
   This is option 1 from the issue discussion (warn, trim, proceed).  The trim pass returns the list of trimmed key paths, so switching to option 2 (error and exit) is a one-line change in `trim_config_text` if that is preferred.

   One deliberate exception: a value that would become *empty* when trimmed is left untouched, so the field's own parser rejects it loudly.  Without this, an all-whitespace `proxy = " "` would silently collapse to `""` (proxy disabled) where it previously hard-errored, which seemed unacceptable for a privacy-relevant field.

2. String values that map to enum-like types now match case-insensitively:
   - `network`: "mainnet", "MAINNET", "FakeChain", ... all work
   - `tor.mode`: "auto", "Daemon", "ARTI", ...
   - `p2p.clear_net.proxy`: "tor" and the "SOCKS5://" scheme (schemes are case-insensitive per RFC 3986; the user:pass and address parts are untouched)
   - the "Default" keyword on every `DefaultOrCustom` field ("default", "DEFAULT", ...)
   - log levels were already case-insensitive via `LevelFilter`'s `FromStr`, so nothing changed there

### Why
Whitespace typos in config files produce confusing errors (see the discussion in #596 that prompted this issue), and users should not need to know Rust's UpperCamelCase enum variant convention to write a config file.

### Where
- `binaries/cuprated/src/config.rs`: trim pass + warnings in `Config::read_from_path`, shared `deserialize_from_str` helper, tests
- `binaries/cuprated/src/config/default.rs`: manual `Deserialize` for `DefaultOrCustom`
- `binaries/cuprated/src/config/{tor,p2p}.rs`: field attributes + doc updates
- `binaries/cuprated/src/tor.rs`: case-insensitive `FromStr` for `TorMode` (without the `arti` feature, "arti" input now gets an actionable "requires the `arti` feature" error instead of an opaque unknown-variant error)
- `binaries/cuprated/src/p2p.rs`: case-insensitive matching in `ProxySettings::try_from`
- `binaries/cuprated/src/logging.rs`: `eprintln_yellow` (shared impl with `eprintln_red`)
- `helper/src/network.rs`: `Network::FromStr` widened to ASCII-case-insensitive, `Display` + `core::error::Error` for `ParseNetworkError`, tests

### How
- `Config::read_from_path` first parses the file with `toml_edit` and walks every string value (nested tables, inline tables, arrays, arrays-of-tables), trimming whitespace and recording the dotted key path of every change, then hands the cleaned text to the existing `toml::from_str` path.  If the file is not parseable TOML the text is passed through unchanged, so error reporting for invalid files is identical to before (covered by a test).
- Case-insensitivity is implemented where each type is defined: `Network` and `TorMode` get case-insensitive `FromStr` impls wired into serde by a shared `deserialize_from_str` helper (same pattern as the existing `level_filter_serde`), `ProxySettings::try_from` compares with `eq_ignore_ascii_case`, and `DefaultOrCustom` implements `Deserialize` manually (the derived `untagged` setup cannot case-fold the "Default" keyword).  As a bonus, `DefaultOrCustom` parse failures now produce concrete "invalid type" messages instead of the derive's "data did not match any variant".
- Serialization is untouched everywhere, so `--generate-config` output is unchanged apart from the new doc lines, and `Network`'s derived serde impls (used outside the config) are unchanged; only `cuprated`'s config fields opt into the case-insensitive path.  The CLI `--network` flag (clap `PossibleValuesParser`, lowercase-only) is also unchanged; happy to unify it with `.ignore_case(true)` here or in a follow-up if wanted.

Known limitation: `toml_edit` 0.23 targets TOML spec 1.0 while `toml` 0.9 targets 1.1, so a config using 1.1-only syntax would skip the trim pass (falling back to today's behavior, with the regular parse still succeeding).  Both crates sit on the same `toml_parser`, so this is a narrow edge that disappears as `toml_edit` adopts 1.1.

The trim walker is exercised against synthetic TOML (arrays, inline tables, arrays-of-tables) rather than `Config` because `Config` currently has no string arrays; the warning emission itself is a plain `eprintln` and is not unit-tested, but the path-collection seam it wraps is.

Disclosure: portions of this change were drafted with AI assistance.  I have personally reviewed, tested, and verified every line, and I take full responsibility for its correctness.

### DIFF
--- a/binaries/cuprated/src/p2p.rs
+++ b/binaries/cuprated/src/p2p.rs
@@ -60,10 +60,12 @@ impl TryFrom<String> for ProxySettings {
     fn try_from(s: String) -> Result<Self, Self::Error> {
         match s.as_str() {
             "" => Ok(Self::Disabled),
-            "Tor" => Ok(Self::Tor),
-            // TODO: use `if let` guard when on >=1.95
-            url if url.starts_with("socks5://") => {
-                let url = url.strip_prefix("socks5://").unwrap();
+            s if s.eq_ignore_ascii_case("tor") => Ok(Self::Tor),
+            url if url
+                .get(..9)
+                .is_some_and(|scheme| scheme.eq_ignore_ascii_case("socks5://")) =>
+            {
+                let url = url.get(9..).unwrap_or_default();
 
                 let (authentication, addr) = match url.rsplit_once('@') {
                     Some((userpass, addr)) => {
--- a/binaries/cuprated/src/config.rs
+++ b/binaries/cuprated/src/config.rs
@@ -23,7 +23,7 @@ use cuprate_p2p_core::{ClearNet, Tor};
 use cuprate_wire::OnionAddr;
 
 use crate::{
-    logging::eprintln_red,
+    logging::{eprintln_red, eprintln_yellow},
     tor::{TorContext, TorMode},
 };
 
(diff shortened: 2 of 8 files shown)


redsh4de:
imo this is a bit more spread out than it needs to be 

I'd probably centralize the related logic into a module like `src/config/trim.rs` and cut down on the tests as some of them bloat the diff imo

redsh4de:
hm, now that there are two variants of colored `eprintln`, i think at this point could also export `Color` from logging.rs and have callsites be just `eprintln_color(Color, format!(...)`, although thats extra churn

thoughts? @Boog900

binaries/cuprated/src/p2p.rs
```diff
@@ -60,10 +60,12 @@ impl TryFrom<String> for ProxySettings {
...
-            "Tor" => Ok(Self::Tor),
-            // TODO: use `if let` guard when on >=1.95
-            url if url.starts_with("socks5://") => {
-                let url = url.strip_prefix("socks5://").unwrap();
+            s if s.eq_ignore_ascii_case("tor") => Ok(Self::Tor),
+            url if url
+                .get(..9)
+                .is_some_and(|scheme| scheme.eq_ignore_ascii_case("socks5://")) =>
+            {
+                let url = url.get(9..).unwrap_or_default();
```
redsh4de, line 68: atp this would be cleaner:

```rust
        // ...
        // Keyword proxies.
        match s.as_str() {
            "" => return Ok(Self::Disabled),
            s if s.eq_ignore_ascii_case("tor") => return Ok(Self::Tor),
            _ => {}
        }

        // URL proxies.
        let (scheme, url) = s
            .split_once("://")
            .ok_or_else(|| anyhow!("Unsupported proxy: '{s}'"))?;

        match scheme.to_ascii_lowercase().as_str() {
            "socks5" => {
        // ...
```

Boog900:
We don't want raw AI code

Issue #540, opened 2025-08-29
Hide certain data in logs


Boog900:
## Feature
Hide certain data from the log file, like stem tx hashes or onion addresses, with an option to show these details. 

## Why
To prevent the log file from being a target.

## Additional context
arti has a crate for this: https://docs.rs/safelog/latest/safelog/index.html

and some guideline for arti: https://tpo.pages.torproject.net/core/arti/contributing/for-developers/logging/

SyntheticBird45:
Interesting. May this behavior be gated behind a debug cfg? Or will this be opt-in + flag/config line to disable

Boog900:
making it opt in would be bad as most wont, it would be opt-out by config/command line yes

MavenRain:
I took a first pass at this using `safelog` (already a workspace dep).  PR adds the opt-out mechanism (a `tracing.redact` config option + `--no-redact` flag, redacting by default per the thread above) and applies it to the transaction-hash log sites in `cuprated`'s txpool.  The peer-address / onion-address sites live in the `p2p-core`, `p2p`, and `address-book` library crates, which don't currently depend on `safelog`;  I'd be happy to extend there in a follow-up if you're OK adding the dep to those crates.  I left the startup `info!("{config}")` dump and block-hash logs out of scope on purpose (notes in the PR).

redsh4de:
This is a bit radical, but thoughts on additionally refusing to start the node if:
1. log redaction has been turned off
2. a node has any RPC enabled on a non-local address (publically available)

Effectively by default, only allowing to disable redaction if the RPC is not available to the outside - i.e. local networks, testing, development, etc. The reason is to make the log file be a target as difficult as possible and an impossibility under normal circumstances unless the user really wants to

This behaviour could be overriden with a `i_know_what_im_doing_allow_unredacted_public_logs` flag, similar to how we gate a unrestricted RPC on a non-local IP.

Boog900:
RPC does not expose the log file though so I am unsure of the gain there?

redsh4de:
it's moreso making sure that in case the RPC is exposed to the public - meaning users other than the node operator are able to connect to it - redaction should be a especially strong default in that case, since the logs can now contain third party data that can be targeted by potential adversaries

ex: if Eve knows that Alice uses Bob's node, Eve can - subpoena for data, compromise Bob's infra, etc like it has happened with Tor node operators

Boog900:
[...]

PR #643, opened 2026-06-13
cuprated: redact transaction hashes from logs

MavenRain:
### What
Adds opt-out redaction of sensitive data from logs, and applies it to the transaction-hash log sites in `cuprated`'s txpool.  This is a first slice of #540, not the whole thing (see "Scope" below).

- New config option `tracing.redact` (default `true`) and a `--no-redact` CLI flag.  When redaction is on (the default), sensitive values are shown as `[scrubbed]`; `--no-redact` (or `redact = false`) shows the raw values.
- Uses [`safelog`](https://docs.rs/safelog) (already a workspace dependency, pulled in by `cuprated` via arti), which is the crate the issue links to.  `safelog` redacts by default; at startup `cuprated` holds a process-lifetime `disable_safe_logging()` guard only when the user opts out.
- Wraps the 8 transaction-hash log/span sites in `txpool/{incoming_tx,manager}.rs` with `safelog::Sensitive`.

Example, default vs `--no-redact`:
```
# default
DEBUG handle_incoming_tx{tx_id=[scrubbed]}: passing tx to tx-pool manager
# --no-redact
DEBUG handle_incoming_tx{tx_id=a1b2...f9}: passing tx to tx-pool manager
```

This matters because the file log defaults to `debug` level, so these transaction hashes were being written to the log file by default.

### Why
Per the issue: to stop the log file from being a target.  A stem-phase transaction hash in a log links a transaction to the node that originated it, so redacting it by default is a privacy win, with an opt-out for debugging (matching the "opt-out by config/command line" decision in the thread).

### Where
- `binaries/cuprated/src/config/tracing_config.rs`: `tracing.redact` field (+ manual `Default`, tests)
- `binaries/cuprated/src/args.rs`: `--no-redact` flag, applied in `apply_args`
- `binaries/cuprated/src/logging.rs`: hold the `safelog` guard for the process when opted out
- `binaries/cuprated/src/txpool/{incoming_tx,manager}.rs`: wrap tx-hash log/span values in `Sensitive`

### How
`safelog::Sensitive<T>` Displays as `[scrubbed]` while safe-logging is enabled (the default) and as the inner value otherwise.  The log sites use the `%` (Display) sigil so the redaction decision is read from `safelog`'s global flag at record time.  The global flag is set once at startup in `init_logging` (before any tx flows), so the behavior is stable for the process.  Both stdout and file logs are redacted (the value is scrubbed at the `Display` layer, before reaching either).

### Scope (what this PR deliberately does NOT do)
To keep the diff contained and avoid unilaterally adding a dependency to library crates, this PR covers only the transaction-hash sites in the `cuprated` binary.  Left for follow-up / discussion:

- **Peer / onion addresses in library crates** (`p2p-core`: handshaker, connector, timeout_monitor spans; `p2p`: seed-node logging; `address-book`: ban/peer-list logging).  These need `safelog` added as a dependency to those crates.  `safelog` already implements `Redactable` for `IpAddr`/`SocketAddr` (partial redaction like `1.2.x.x`), so the extension is mechanical once the dependency is acceptable.  Happy to do it here or separately, your call.
- **The startup `info!("{config}")` dump** (`main.rs`), which prints the node's own config including its onion address and bind/proxy addresses via `Debug`.  That is the node's own configuration rather than per-tx/per-peer data, and redacting it needs a different approach (the `Config` `Debug`/`Display` impl, not per-call-site wrapping); flagging it rather than bundling it.
- **Block hashes** (`blockchain/manager/handler.rs`) and the **RPC bind-address** logs (`rpc/server.rs`) are left unredacted on purpose: block hashes are public chain data, and the RPC bind address is operator-local config, so scrubbing them would cost readability without a privacy benefit.

Open questions for maintainers: (1) OK to add `safelog` to `p2p-core`/`p2p`/`address-book` for the address sites?  (2) redact at all levels (current behavior) or only `info`+ as the arti guideline suggests?

Disclosure: portions of this change were drafted with AI assistance.  I have personally reviewed, tested, and verified every line, and I take full responsibility for its correctness.

### DIFF
--- a/binaries/cuprated/src/config/tracing_config.rs
+++ b/binaries/cuprated/src/config/tracing_config.rs
@@ -5,9 +5,21 @@ use super::macros::config_struct;
 
 config_struct! {
     /// [`tracing`] config.
-    #[derive(Debug, Default, Deserialize, Serialize, Eq, PartialEq)]
+    #[derive(Debug, Deserialize, Serialize, Eq, PartialEq)]
     #[serde(deny_unknown_fields, default)]
     pub struct TracingConfig {
+        /// Redact sensitive data from logs.
+        ///
+        /// When enabled, sensitive values such as transaction hashes are
+        /// replaced with `[scrubbed]` in logs, so a shared log file does not
+        /// leak data that could deanonymise you or your peers.
+        ///
+        /// Disable this only if you need the raw values, e.g. for debugging.
+        ///
+        /// Type         | boolean
+        /// Valid values | true, false
+        pub redact: bool,
+
         #[child = true]
         /// Configuration for cuprated's stdout logging system.
         pub stdout: StdoutTracingConfig,
@@ -15,6 +27,16 @@ config_struct! {
         #[child = true]
         /// Configuration for cuprated's file logging system.
         pub file: FileTracingConfig,
+    }
+}
+
+impl Default for TracingConfig {
+    fn default() -> Self {
+        Self {
+            redact: true,
+            stdout: StdoutTracingConfig::default(),
+            file: FileTracingConfig::default(),
+        }
     }
 }
 
@@ -75,6 +97,22 @@ impl Default for FileTracingConfig {
             level: LevelFilter::DEBUG,
             max_log_files: 7,
         }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn redact_defaults_to_true() {
+        assert!(TracingConfig::default().redact);
+    }
+
+    #[test]
+    fn redact_can_be_disabled() {
+        let config: TracingConfig = toml::from_str("redact = false").unwrap();
+        assert!(!config.redact);
     }
 }
 
--- a/binaries/cuprated/src/logging.rs
+++ b/binaries/cuprated/src/logging.rs
@@ -29,6 +29,12 @@ use cuprate_helper::fs::logs_path;
 
 use crate::config::Config;
 
+/// A [`OnceLock`] which holds the [`safelog::Guard`] that disables log redaction.
+///
+/// Only set when the user opts out of redaction; while it is held, sensitive
+/// values are shown in cleartext. Initialized in [`init_logging`].
+static SAFE_LOGGING_GUARD: OnceLock<safelog::Guard> = OnceLock::new();
+
 /// A [`OnceLock`] which holds the [`Handle`] to update the file logging output.
 ///
 /// Initialized in [`init_logging`].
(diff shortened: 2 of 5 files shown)


binaries/cuprated/src/logging.rs
```diff
@@ -85,6 +91,16 @@ impl<S> Filter<S> for CupratedTracingFilter {
...
 /// Initialize [`tracing`] for logging to stdout and to a file.
 pub fn init_logging(config: &Config) {
+    // `safelog` redacts sensitive values (e.g. transaction hashes) from logs by
+    // default. If the user has opted out, disable redaction for the lifetime of
+    // the process.
+    if !config.tracing.redact {
+        let guard = safelog::disable_safe_logging().expect("nothing else enforces safe logging");
+        SAFE_LOGGING_GUARD
+            .set(guard)
+            .expect("init_logging must only be called once");
```
redsh4de, line 101:
```suggestion
        match safelog::disable_safe_logging() {
            Ok(guard) => SAFE_LOGGING_GUARD
                .set(guard)
                .expect("init_logging must only be called once"),
            Err(e) => tracing::warn!(%e, "could not disable log redaction; keeping it on"),
        }
```

binaries/cuprated/src/config/tracing_config.rs
```diff
@@ -78,6 +100,22 @@ impl Default for FileTracingConfig {
...
+    #[test]
+    fn redact_defaults_to_true() {
+        assert!(TracingConfig::default().redact);
+    }
+
+    #[test]
+    fn redact_can_be_disabled() {
+        let config: TracingConfig = toml::from_str("redact = false").unwrap();
+        assert!(!config.redact);
+    }
```
redsh4de, line 116: unnecessary imo

binaries/cuprated/src/config/tracing_config.rs
```diff
@@ -78,6 +100,22 @@ impl Default for FileTracingConfig {
...
 }
 
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn redact_defaults_to_true() {
+        assert!(TracingConfig::default().redact);
+    }
```
redsh4de, line 110: would move to the existing `config.rs` mod test and get rid of the new mod test block here:

```rust
fn redact_defaults_to_true() {
    assert!(Config::default().tracing.redact);
}
```

redsh4de (changes requested):

Boog900:
We don't want raw AI code

PR #647, opened 2026-07-06
RPC: Add basic get_transactions support

Boog900:
We don't support JSON encoded txs as wallets don't need it and it is a massive pain.

### DIFF
--- a/types/types/src/blockchain.rs
+++ b/types/types/src/blockchain.rs
@@ -214,6 +214,8 @@ pub enum BlockchainReadRequest {
     AltChainCount,
 
     /// Get transaction blobs by their hashes.
+    ///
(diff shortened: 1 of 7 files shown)


types/types/src/blockchain.rs
```diff
@@ -214,7 +214,7 @@ pub enum BlockchainReadRequest {
     AltChainCount,
 
     /// Get transaction blobs by their hashes.
```
redsh4de, line 216:
```suggestion
    /// Get transaction blobs by their hashes.
    ///
    /// Returned transactions must preserve request order, omitting missing hashes.
```

documenting the invariant

redsh4de (approved):
lgtm w/ a nit

PR #658, opened 2026-07-29
cuprated: add custom allocators

### LINKED ISSUE #336: Custom allocators
## What

Add custom allocators to cuprated behind a feature flag.

## Why
Targets that are statically linking (like musl libc by default) will inherit the default (potentially very slow) allocator. Rust permits to easily swap out an allocator and I think this might be beneficial to have a feature flag for it. 
The allocators we could support (or force):
- [jemalloc](https://github.com/jemalloc/jemalloc)
- [mimalloc](https://github.com/microsoft/mimalloc)

redsh4de:
### What
Add support for specifying custom allocators via feature flags

Closes #336 

### Why
Pre-requisite for proper musl support.

### Where
`cuprated`

### How
Added mimalloc and jemallocator as deps

### DIFF
--- a/binaries/cuprated/src/main.rs
+++ b/binaries/cuprated/src/main.rs
@@ -44,7 +44,7 @@ mod commands;
 static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
 
 #[cfg(all(
-    feature = "jemalloc",
+    any(feature = "jemalloc", target_env = "musl"),
     not(feature = "mimalloc"),
     not(target_env = "msvc")
 ))]
--- a/binaries/cuprated/src/main.rs
+++ b/binaries/cuprated/src/main.rs
@@ -44,7 +44,7 @@ mod commands;
 static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
 
 #[cfg(all(
-    any(feature = "jemalloc", target_env = "musl"),
+    feature = "jemalloc",
     not(feature = "mimalloc"),
     not(target_env = "msvc")
 ))]
--- a/binaries/cuprated/src/main.rs
+++ b/binaries/cuprated/src/main.rs
@@ -39,6 +39,18 @@ use cuprated::{
 mod args;
 mod commands;
 
+#[cfg(feature = "mimalloc")]
+#[global_allocator]
+static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
+#[cfg(all(
+    feature = "jemalloc",
+    not(feature = "mimalloc"),
+    not(target_env = "msvc")
+))]
+#[global_allocator]
+static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
 use crate::{
     args::Args,
     commands::{command_listener, Command},
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1543,6 +1543,7 @@ dependencies = [
  "hex",
  "hex-literal 1.1.0",
  "indexmap 2.14.0",
+ "mimalloc",
  "monero-address",
  "monero-oxide",
  "nu-ansi-term",
@@ -1562,6 +1563,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "thread_local",
+ "tikv-jemallocator",
  "tokio",
  "tokio-stream",
  "tokio-util",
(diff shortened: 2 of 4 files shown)


binaries/cuprated/src/main.rs
```diff
@@ -39,6 +39,18 @@ use cuprated::{
 mod args;
 mod commands;
 
+#[cfg(feature = "mimalloc")]
+#[global_allocator]
+static ALLOC: mimalloc::MiMalloc = mimalloc::MiMalloc;
```
SyntheticBird45, line 44: A "secure" variant would be appreciated:
https://docs.rs/crate/mimalloc/0.1.52/features#secure

binaries/cuprated/src/main.rs
```diff
@@ -39,6 +39,18 @@ use cuprated::{
...
+
+#[cfg(all(
+    any(feature = "jemalloc", target_env = "musl"),
```
SyntheticBird45: I would like to avoid that musl target_env. Alpine supports dynamically overrided allocator and this would force Rust on not relying on the system allocator and force usage of jemalloc. I would prefer this to be opt-in and let musl user rely on system allocator so they can use LD_PRELOAD

This is another topic worth discussing in meeting but I think mimalloc is worth being set as a default feature and included in every release artifacts.

SyntheticBird45 (changes requested):

binaries/cuprated/src/main.rs
redsh4de: added `mimalloc-secure` flag

binaries/cuprated/src/main.rs
redsh4de: removed the default on musl, both are opt-in in all cases now

SyntheticBird45 (approved):

Boog900 (approved):

kayabaNerve:
I understand I'm late and this was already merged, but why was this a "pre-requisite" for musl?

I understand it's feature-flagged but I STRONGLY believe applications SHOULD NOT set a distinct global allocator. If we assume a default distribution will enable `jemalloc`, as it's 'faster and better', then the default distribution won't work on a variety of systems I run because `jemalloc` trips those systems' hardened allocator.

Separately, the allocator (for applications as a whole) SHOULD be handled by the system. I understand why applications handle it, because the systems frequently do not, but that's why in distros I've contributed to, discussions I've participated in, and my own local systems, I do push for competent system allocators. Even if the application's allocator doesn't conflict, it still takes away the system's ability to choose the allocator if it's part of the binary's build flags (without building from scratch).

TL;DR I think Cuprate shouldn't specify any global allocator and this should be reverted, as a drive-by commenter.

redsh4de:
It being deemed a pre-requisite comes down to the musl allocator being [notoriously slow for multi-threaded applications, even to this day](https://nickb.dev/blog/default-musl-allocator-considered-harmful-to-performance/), and LD_PRELOAD only working for dynamically linked binaries. It would work "fine", so its not a hard pre-req, but for a performance oriented node i believe its a necessity to have the option be available

That said, this is not a default feature, so it's still left up to the user and doesn't take away choice. It being a option benefits environments like Alpine where Rust link statically by default, which the Dockerfile uses for the build stage, which FWIW is the only place where the feature is _currently_ set in the repo. (And even then if people really want to use `mallocng`, they just have to pass a empty FEATURES arg there.)

And since Rust defaults to static linking on musl, we cannot LD_PRELOAD `jemalloc` at runtime, it has to be baked in at build time. So it gives more flexibility for users who want to build a static cuprated as well

kayabaNerve:
So it isn't a "prerequisite", it's a "preference".

I do hear `LD_PRELOAD` is only available given dynamic linking though, and in a statically-linked world, the conflicts I was concerned about wouldn't occur.

I'd still support reverting this, or at least making it explicitly conditional to when static linking is occurring (`cfg(target_feature = "crt-static")`).

SyntheticBird45:
> So it isn't a "prerequisite", it's a "preference".
> 
> I do hear `LD_PRELOAD` is only available given dynamic linking though, and in a statically-linked world, the conflicts I was concerned about wouldn't occur.
> 
> I'd still support reverting this, or at least making it explicitly conditional to when static linking is occurring (`cfg(target_feature = "crt-static")`).

Please see my review comment here https://github.com/Cuprate/cuprate/pull/658#discussion_r3675088362
I asked it to not be default for musl and by default rely on the allocator because Alpine supports LD_PRELOAD to dynamically change. There are packages and scripts in the their repository for doing so: https://pkgs.alpinelinux.org/contents?name=jemalloc&repo=main&branch=edge&arch=x86_64.

As for NOT allowing hardcoding global allocators for glibc, I purely disagree. This is opt-in, people are supposed to know what they are doing when including new features and LD_PRELOAD is in my opinion a very bad mechanism.

PR #663, opened 2026-08-01
cuprated: add seed node config option and --seed-node argument

Fountain5405:
## Problem

`Network::FakeChain` has no built-in seed nodes — `clear_net_seed_nodes()` returns `[]` for it — and there is currently no way to supply any.

A single `cuprated --regtest` node is fine, but as soon as you want two or more to find each other there is no mechanism to point them at anything, and a node with an empty address book panics in `connection_maintainer`:

```
No seed nodes available to get peers from
```

So multi-node regtest isn't reachable today without editing the hardcoded seed table and rebuilding.

## Change

Two commits, either of which stands alone:

1. **`p2p.clear_net.seed_nodes`** — an optional config field, a list of socket addresses appended to the network's built-in seeds.
2. **`--seed-node <IP:PORT>`** — a repeatable CLI argument that appends to the same list, following the existing `--outbound-connections` override pattern, so a regtest network needs no config file at all:

   ```console
   $ cuprated --regtest --seed-node 127.0.0.1:18080
   ```

- Defaults to empty and is `#[comment_out = true]`, so Mainnet/Testnet/Stagenet behaviour is unchanged.
- Appended to, not replacing, the built-in seeds.
- ~36 lines, no new dependencies.
- One thing worth a reviewer's eye: `Args::apply_args` loses its `const` qualifier, since extending a `Vec` is not a const operation. Happy to drop the second commit if you'd rather keep it `const` and stay config-only.

## Why it's useful

- multi-node regtest, locally and in CI
- private / isolated networks
- network simulation — this is the one patch I needed to run `cuprated` under the [Shadow](https://shadow.github.io/) discrete-event network simulator alongside `monerod`

## Testing

`cargo fmt --all --check`, `cargo clippy -p cuprated --all-targets -- -D warnings`, `RUSTDOCFLAGS='-D warnings' cargo doc -p cuprated`, and `cargo test -p cuprated` all pass locally — including the `documented_config` round-trip test, which the new field satisfies.

Runtime check: `cuprated --regtest` with `seed_nodes` set loads the value, dials the configured address and completes a handshake.

Exercised at scale: a 306-node FakeChain network (180 `cuprated`, 126 `monerod`) bootstrapped entirely through this option under Shadow, reaching full block and transaction propagation across both implementations. That run was on the `0.0.9` base carrying this same patch; it is being re-run on this base now.

### DIFF
--- a/binaries/cuprated/src/args.rs
+++ b/binaries/cuprated/src/args.rs
@@ -1,4 +1,4 @@
-use std::{io::Write, path::PathBuf, process::exit};
+use std::{io::Write, net::SocketAddr, path::PathBuf, process::exit};
 
 use clap::builder::TypedValueParser;
 use serde_json::Value;
@@ -47,6 +47,14 @@ pub struct Args {
     #[arg(long)]
     pub outbound_connections: Option<usize>,
 
+    /// An extra seed node to connect to on startup, in addition to the
+    /// network's built-in seeds.
+    ///
+    /// Can be passed multiple times. `FakeChain` has no built-in seeds, so
+    /// this is how a regtest or otherwise isolated network is bootstrapped.
+    #[arg(long, value_name = "IP:PORT")]
+    pub seed_node: Vec<SocketAddr>,
+
     /// The PATH of the `cuprated` config file.
     #[arg(long)]
     pub config_file: Option<PathBuf>,
@@ -94,7 +102,7 @@ impl Args {
     /// Apply the [`Args`] to the given [`Config`].
     ///
     /// This may exit the program if a config value was set that requires an early exit.
-    pub const fn apply_args(&self, mut config: Config) -> Config {
+    pub fn apply_args(&self, mut config: Config) -> Config {
         if let Some(network) = self.network {
             config.network = network;
         }
@@ -112,6 +120,12 @@ impl Args {
         if let Some(outbound_connections) = self.outbound_connections {
             config.p2p.clear_net.outbound_connections = outbound_connections;
         }
+
+        config
+            .p2p
+            .clear_net
+            .seed_nodes
+            .extend_from_slice(&self.seed_node);
 
         config
     }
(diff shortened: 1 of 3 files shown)


binaries/cuprated/src/config.rs
```diff
@@ -280,7 +280,14 @@ impl Config {
     pub fn clearnet_p2p_config(&self) -> cuprate_p2p::P2PConfig<ClearNet> {
         cuprate_p2p::P2PConfig {
             network: self.network,
-            seeds: p2p::clear_net_seed_nodes(self.network),
+            seeds: {
+                // Built-in network seeds plus any user-configured seed nodes
+                // (`p2p.clear_net.seed_nodes`). FakeChain/regtest ships no
+                // built-in seeds, so a private/isolated network relies on these.
```
SyntheticBird45: nit: I don't think this comment is necessary.

binaries/cuprated/src/args.rs
```diff
@@ -113,6 +121,12 @@ impl Args {
             config.p2p.clear_net.outbound_connections = outbound_connections;
         }
 
+        config
+            .p2p
+            .clear_net
+            .seed_nodes
+            .extend(self.seed_node.iter().copied());
```
SyntheticBird45: use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice

binaries/cuprated/src/config.rs
```diff
@@ -280,7 +280,14 @@ impl Config {
...
+                let mut seeds = p2p::clear_net_seed_nodes(self.network);
+                seeds.extend(self.p2p.clear_net.seed_nodes.iter().copied());
```
SyntheticBird45: here too use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice

SyntheticBird45 (changes requested):

SyntheticBird45 (approved):

Boog900 (approved):

Issue #472, opened 2025-05-13
Expanded tier 1/2 build targets


hinto-janai:
## What
This is for discussing an expanded tier 1/2 build target list for `cuprated`.

Notable targets:
- x86_64-unknown-linux-musl
- aarch64-unknown-linux-musl
- x86_64-unknown-freebsd
- aarch64-unknown-freebsd
- aarch64-pc-windows-msvc
- x86_64-unknown-openbsd
- riscv64gc-unknown-linux-gnu
- riscv64gc-unknown-linux-musl

SyntheticBird45:
In order of priority imo:
1. x86_64/aarch64 musl linux
2. riscv64 linux gnu/musl
3. freebsd/openbsd
4. aarch64 windows

hinto-janai:
Noting that industry ([including Rust](https://blog.rust-lang.org/2025/10/30/Rust-1.91.0/#aarch64-pc-windows-msvc-is-now-a-tier-1-platform)) are continuing to push aarch64 Windows, I suspect it will become quite relevant in the future.

PR #665, opened 2026-08-02
cuprated: expand CI and release targets

redsh4de:
### What
Expand CI and release targets - add support for OpenBSD 7.9, FreeBSD, RISC-V Linux, x86 MUSL Linux. Part of #472 

### Why
More platforms = gud

### Where
`cuprated`

### How
- In both CI and release workflows, add jobs that build *BSD in virtual machines (`cross-platform-actions`), RISC-V via cross-compilation (`taiki-e/setup-cross-toolchain-action`), and x64 musl in an Alpine container.
- musl builds with `jemalloc`
- `monerod-download` action takes `os`/`arch` inputs now - `runner.*` describes the host, which didnt work for VM/cross jobs. OpenBSD and RISC-V have no monerod, so we skip those tests.
- Pinned `sysinfo` to 0.38.4 to lower the MSRV for OpenBSD.
- Update `platform.md`

### DIFF
--- a/.github/actions/monerod-download/action.yml
+++ b/.github/actions/monerod-download/action.yml
@@ -23,6 +23,12 @@ inputs:
     description: "Version to download"
     required: false
     default: v0.18.3.3
+  os:
+    description: "Operating system of the binary"
+    required: true
+  arch:
+    description: "Architecture of the binary"
+    required: true
 
 runs:
   using: "composite"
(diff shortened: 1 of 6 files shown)


redsh4de:
Tried but didnt work (and cba for now):

- **cross-compiling aarch64-musl.** musl toolchain is GCC 9.4.0, broken `arm_acle.h` ([PR100985](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100985)). Fixed in 9.5.0, upstream needs to update
- **aarch64-pc-windows-msvc** -  RandomX ARM64 assembly is written for GCCs assembler and MSVC can't build it.
- **riscv64gc-unknown-linux-musl** - RandomX (again)

SyntheticBird45 (approved):
I have zero comments. This looks perfect to me

Boog900 (approved):

PR #667, opened 2026-08-06
P2P: fix peer-set desync that can abort the node

thomasbuilds:
On clearnet a peer can have its `InternalPeerID` reused by a connection in the opposite direction: drop the outbound connection we hold, then reconnect inbound from the address we dialled. `ClosedConnectionFuture` carries only that id, so when it fires the entry in `peers` may belong to the new connection rather than the one that closed.

`remove_dead_peers` removed it regardless, dropping the `StoredClient` and disconnecting a live peer. `outbound_peers` was only pruned when that entry was outbound, so a slot taken over by an inbound connection left a stale id for `random_peer_for_stem` to `unwrap`. As the workspace sets `panic = "abort"`, that ends the process rather than the request.

Drop `outbound_peers` rather than fix its bookkeeping: it had a single consumer, and keeping it in sync with `peers` is what failed. Outbound peers are filtered from `peers` on demand, so an inbound peer cannot be selected for Dandelion++ stemming, and an entry is only removed once its own handle is closed.

### DIFF
--- a/p2p/p2p/src/peer_set.rs
+++ b/p2p/p2p/src/peer_set.rs
@@ -5,8 +5,8 @@ use std::{
 };
 
 use futures::{stream::FuturesUnordered, StreamExt};
-use indexmap::{IndexMap, IndexSet};
-use rand::{seq::index::sample, thread_rng};
+use indexmap::IndexMap;
+use rand::{seq::IteratorRandom, thread_rng};
 use tokio::sync::mpsc::Receiver;
 use tokio_util::sync::WaitForCancellationFutureOwned;
 use tower::Service;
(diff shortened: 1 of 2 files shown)


p2p/p2p/src/peer_set.rs
```diff
@@ -215,3 +207,115 @@ impl<N: NetworkZone> Service<PeerSetRequest> for PeerSet<N> {
...
+        peer_set.ready().await.unwrap();
+
+        let PeerSetResponse::StemPeer(stem) =
+            peer_set.call(PeerSetRequest::StemPeer).await.unwrap()
+        else {
+            panic!("peer set returned the wrong response");
+        };
+        assert!(stem.is_some());
+    }
+}
```
Boog900: Can you move these tests into `peer_set/tests.rs`

Boog900 (changes requested):

p2p/p2p/src/peer_set.rs
thomasbuilds: this has been done

PR #671, opened 2026-08-07
cuprated: allow onion addresses to be specified as seed nodes

jpk68:
### What

Allows `.onion` addresses to be specified as seed nodes, both in the `--seed_node` command-line argument, and in the config file.

### Why

This would be useful to have as an option.

### Where

`cuprated`, `net/wire/`

### How

Implements some `serde` traits for onion addresses, then adds them to the config code.

### DIFF
--- a/binaries/cuprated/src/args.rs
+++ b/binaries/cuprated/src/args.rs
@@ -1,11 +1,35 @@
-use std::{io::Write, net::SocketAddr, path::PathBuf, process::exit};
+use std::{io::Write, net::SocketAddr, path::PathBuf, process::exit, str::FromStr};
 
 use clap::builder::TypedValueParser;
 use serde_json::Value;
 
 use cuprate_helper::network::Network;
+use cuprate_wire::OnionAddr;
 
 use cuprated::{config::Config, version::CupratedVersionInfo};
+
+#[derive(Debug, Clone)]
+pub(crate) enum SeedNodeArg {
+    Clear(SocketAddr),
+    Tor(OnionAddr),
+}
+
+impl FromStr for SeedNodeArg {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(addr) = s.parse::<SocketAddr>() {
+            Ok(Self::Clear(addr))
+        } else if let Ok(addr) = s.parse::<OnionAddr>() {
+            Ok(Self::Tor(addr))
+        } else {
+            Err(format!(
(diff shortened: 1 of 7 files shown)


binaries/cuprated/src/args.rs
```diff
@@ -1,12 +1,36 @@
...
+            return Ok(Self::Clear(addr));
+        }
+
+        s.parse::<OnionAddr>().map(Self::Tor).map_err(|_| {
+            format!(
+                "'{s}' is not a valid seed node address; expected either \
+                 an \"ip:port\" or \"<56-character onion domain>.onion:port\" address"
+            )
+        })
+    }
```
SyntheticBird45, line 31: nit: put the rest in an `else` statement

net/wire/Cargo.toml
```diff
@@ -23,6 +24,7 @@ bytes                   = { workspace = true, features = ["std"] }
 thiserror               = { workspace = true }
 
 arbitrary = { workspace = true, features = ["derive"], optional = true }
+serde     = { workspace = true, features = ["std"], optional = true }
```
SyntheticBird45: is `std` feature here really necessary ?

SyntheticBird45 (changes requested):

SyntheticBird45 (approved):

binaries/cuprated/src/args.rs
```diff
@@ -1,12 +1,36 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(addr) = s.parse::<SocketAddr>() {
+            Ok(Self::Clear(addr))
+        } else {
+            s.parse::<OnionAddr>().map(Self::Tor).map_err(|_| {
+                format!(
+                    "'{s}' is not a valid seed node address; expected either \
+                     an \"ip:port\" or \"<56-character onion domain>.onion:port\" address"
+                )
+            })
```
Boog900: Seems like a good place for our first if let chain:


```suggestion
        } else if let Ok(addr) = s.parse::<OnionAddr>() {
            Ok(Self::Tor(addr))
        } else {
            Err(format!(
                "'{s}' is not a valid seed node address; expected either \
                     an \"ip:port\" or \"<56-character onion domain>.onion:port\" address"
            ))
```

SyntheticBird45: Sir, this is Rust 2021.

Boog900: It works for me, tbc I got confused this is not an if let chain

Boog900 (approved):

PR #675, opened 2026-08-07
RPC: Stream timeout (hardening pt.1)

SyntheticBird45:
> [!NOTE]
> This PR has been split. Some of the reviews reference code that has been removed from the branch.

# What

This PR replace the `axum::serve` usage in `cuprated` by a custom hyper serving loop under an `RpcServer` struct. By doing so we can wrap connection's `TcpStream` into a timeount wrapper struct called `WriteTimeout`. This allows us to kick connections to which we cannot send bytes after a configurable amount of time.

This PR also introduce a header read and body read timeout. The first will abort the connection. The latter will first respond with a `400 Bad Request` response and then abort the connection.

This is an important mitigation against inactive/keep-alive/unidirectional spamming connections.

### DIFF
--- a/binaries/cuprated/Cargo.toml
+++ b/binaries/cuprated/Cargo.toml
@@ -50,7 +50,7 @@ cuprate-types             = { workspace = true, features = ["json"] }
 cuprate-wire              = { workspace = true, features = ["serde"] }
 
 # TODO: after v1.0.0, remove unneeded dependencies.
-axum                  = { workspace = true, features = ["tokio", "http1", "http2"] }
+axum                  = { workspace = true, features = ["tokio", "http1"] }
 anyhow                = { workspace = true }
 arti-client           = { workspace = true, features = ["tokio", "rustls", "onion-service-client", "onion-service-service", "static"], optional = true }
 async-trait           = { workspace = true }
@@ -71,6 +71,8 @@ fjall                 = { workspace = true }
 futures               = { workspace = true }
 hex                   = { workspace = true }
 hex-literal           = { workspace = true }
+hyper                 = { workspace = true }
+hyper-util            = { workspace = true }
 indexmap              = { workspace = true }
 mimalloc              = { workspace = true, optional = true }
 monero-address        = { workspace = true }
@@ -78,6 +80,7 @@ monero-oxide          = { workspace = true, features = ["std", "compile-time-gen
 nu-ansi-term          = { workspace = true }
 paste                 = { workspace = true }
 pin-project           = { workspace = true }
+pin-project-lite      = { workspace = true }
 randomx-rs            = { workspace = true }
 rand                  = { workspace = true }
 rand_distr            = { workspace = true }
@@ -99,7 +102,7 @@ tor-hsservice         = { workspace = true, optional = true }
 tor-persist           = { workspace = true, optional = true }
 tor-rtcompat          = { workspace = true, optional = true }
 tower                 = { workspace = true, features = ["limit"] }
-tower-http            = { workspace = true, features = ["limit", "trace"] }
+tower-http            = { workspace = true, features = ["limit", "timeout", "trace"] }
 tracing-appender      = { workspace = true }
 tracing-subscriber    = { workspace = true, features = ["std", "fmt", "default"] }
 tracing               = { workspace = true, features = ["default"] }
@@ -118,4 +121,3 @@ serde_json  = { workspace = true, features = ["std"] }
 
 [lints]
 workspace = true
-
--- a/binaries/cuprated/src/config/rpc.rs
+++ b/binaries/cuprated/src/config/rpc.rs
@@ -58,6 +58,28 @@ config_struct! {
         /// Examples     | 0 (no limit), 5242880 (5MB), 10485760 (10MB)
         pub request_byte_limit: usize,
 
+        /// The time period during which the node can try sending data.
+        /// If a send operation do not complete within this Duration,
+        /// the connection is dropped.
+        ///
+        /// Type     | Duration
+        /// Examples | { secs = 10, nanos = 0 }, { secs = 29, nanos = 123 }
+        pub send_timeout: Duration,
+
+        /// The maximum time allowed to receive an HTTP/1 request header.
+        /// Receiving progress does not restart this deadline.
+        ///
+        /// Type     | Duration
+        /// Examples | { secs = 10, nanos = 0 }, { secs = 29, nanos = 123 }
+        pub header_read_timeout: Duration,
+
+        /// The maximum wall-clock time allowed to consume an HTTP request body.
+        /// Receiving progress does not restart this deadline.
+        ///
+        /// Type     | Duration
+        /// Examples | { secs = 10, nanos = 0 }, { secs = 29, nanos = 123 }
+        pub body_read_timeout: Duration,
+
         // TODO: <https://github.com/Cuprate/cuprate/issues/445>
     }
 
@@ -101,6 +123,9 @@ impl Default for UnrestrictedRpcConfig {
             port: DefaultOrCustom::Default,
             enable: true,
             request_byte_limit: 0,
+            send_timeout: Duration::from_secs(5),
+            header_read_timeout: Duration::from_secs(30),
+            body_read_timeout: Duration::from_secs(5),
         }
     }
 }
@@ -115,6 +140,9 @@ impl Default for RestrictedRpcConfig {
             // 1 megabyte.
             // <https://github.com/monero-project/monero/blob/3b01c490953fe92f3c6628fa31d280a4f0490d28/src/cryptonote_config.h#L134>
             request_byte_limit: 1024 * 1024,
+            send_timeout: Duration::from_secs(5),
+            header_read_timeout: Duration::from_secs(30),
+            body_read_timeout: Duration::from_secs(5),
         }
     }
 }
--- a/binaries/cuprated/src/rpc/server.rs
+++ b/binaries/cuprated/src/rpc/server.rs
@@ -1,25 +1,47 @@
 //! RPC server initialization and main loop.
 
 use std::{
+    cell::{Cell, UnsafeCell},
+    collections::HashMap,
     net::{IpAddr, SocketAddr},
+    num::NonZeroUsize,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
     time::Duration,
 };
 
 use anyhow::Error;
-use tokio::net::TcpListener;
-use tokio_util::sync::CancellationToken;
-use tower::limit::rate::RateLimitLayer;
-use tower_http::limit::RequestBodyLimitLayer;
-use tracing::{info, warn};
+use axum::Router;
+use hyper::{body::Incoming, server::conn::http1, Request};
+use hyper_util::{
+    rt::{TokioExecutor, TokioIo, TokioTimer},
+    service::TowerToHyperService,
+};
+use tokio::{
+    net::{TcpListener, TcpStream},
+    task::JoinSet,
+    time::timeout,
+};
+use tokio_util::{sync::CancellationToken, time::FutureExt};
+use tower::{limit::rate::RateLimitLayer, Service};
+use tower_http::{limit::RequestBodyLimitLayer, timeout::RequestBodyDeadlineLayer};
+use tracing::{debug, error, info, warn};
 
+use cuprate_helper::net::ip_is_local;
 use cuprate_rpc_interface::RouterBuilder;
 
 use crate::{
-    config::{restricted_rpc_port, unrestricted_rpc_port},
-    rpc::CupratedRpcHandler,
+    config::{restricted_rpc_port, unrestricted_rpc_port, RpcConfig},
+    rpc::{timeout::WriteTimeout, CupratedRpcHandler},
     txpool::IncomingTxHandler,
     LaunchContext,
 };
+
+/// The maximum amount of time we wait for connections to gracefully close
+/// after shutdown signal.
+const RPC_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Initialize the RPC server(s).
 ///
@@ -55,7 +77,7 @@ pub fn init_rpc_servers(launch_ctx: &LaunchContext, tx_handler: IncomingTxHandle
             continue;
         }
 
-        if !restricted && !cuprate_helper::net::ip_is_local(addr) {
+        if !restricted && !ip_is_local(addr) {
             if config
                 .unrestricted
                 .i_know_what_im_doing_allow_public_unrestricted_rpc
@@ -69,41 +91,59 @@ pub fn init_rpc_servers(launch_ctx: &LaunchContext, tx_handler: IncomingTxHandle
             }
         }
 
+        let (rpc_send_timeout, rpc_header_read_timeout, rpc_body_read_timeout) = if restricted {
+            (
+                config.restricted.send_timeout,
+                config.restricted.header_read_timeout,
+                config.restricted.body_read_timeout,
+            )
+        } else {
+            (
+                config.unrestricted.send_timeout,
+                config.unrestricted.header_read_timeout,
+                config.unrestricted.body_read_timeout,
+            )
+        };
+
+        // Initialize RPC handler service.
         let rpc_handler = CupratedRpcHandler::new(restricted, tx_handler.clone(), launch_ctx);
 
+        // Initialize Axum RPC router.
+        let rpc_router = init_rpc_router(rpc_handler, request_byte_limit, rpc_body_read_timeout);
+
+        // Build the RPC server.
+        let rpc_server = RpcServer::new(
+            SocketAddr::new(addr, port),
+            rpc_send_timeout,
+            rpc_header_read_timeout,
+            rpc_router,
+        );
+
+        info!(
+            restricted,
+            address = %addr,
+            "Starting RPC server"
+        );
+
+        // Launch RPC server.
         let shutdown_token = launch_ctx.task_executor.cancellation_token();
         launch_ctx.task_executor.spawn(async move {
-            run_rpc_server(
-                rpc_handler,
-                restricted,
-                SocketAddr::new(addr, port),
-                request_byte_limit,
-                shutdown_token,
-            )
-            .await
-            .unwrap();
+            if let Err(e) = rpc_server.run(shutdown_token).await {
+                error!(restricted, "RPC server error: {e:#}");
+            }
+
             info!(restricted, "RPC server shut down.");
         });
     }
 }
 
-/// This initializes and runs an RPC server.
-///
-/// The function will only return when the server itself returns or an error occurs.
-async fn run_rpc_server(
+/// Initialize the Axum router of the RPC server.
+fn init_rpc_router(
     rpc_handler: CupratedRpcHandler,
-    restricted: bool,
-    address: SocketAddr,
     request_byte_limit: usize,
-    shutdown_token: CancellationToken,
-) -> Result<(), Error> {
-    info!(
-        restricted,
-        address = %address,
-        "Starting RPC server"
-    );
-
-    let router = RouterBuilder::new()
+    body_read_timeout: Duration,
+) -> Router {
+    let mut router = RouterBuilder::new()
         .json_rpc()
         //
         .other_get_info()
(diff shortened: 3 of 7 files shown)


binaries/cuprated/src/rpc/server.rs
```diff
@@ -142,15 +173,215 @@ async fn run_rpc_server(
...
+    /// Maximum amount of connections per private IP address.
+    max_conn_count_private_ip: usize,
+    /// Maximum amount of connections per loopback IP address.
+    max_conn_count_loopback: usize,
+    /// Total maximum amount of connections by limited IP addresses.
+    max_conn_count: usize,
+    /// IP addresses that are excluded from any limitations.
+    excluded_ips: Vec<IpAddr>,
+    /// Amount of current connections per limited IP address.
+    per_ip_conn_count: HashMap<SocketAddr, usize>,
```
Boog900: This needs to be `IpAddr` not `SocketAddr`

binaries/cuprated/src/rpc/server.rs
```diff
@@ -142,15 +173,215 @@ async fn run_rpc_server(
...
-    Ok(())
+            remote_addr
+        });
+    }
+}
+
+#[derive(Clone)]
+/// A concurrently mutated cache of the number of connections of all the
+/// IP address currently connected to the RPC server.
+struct RpcLimitCache {
```
Boog900: The docs need updating

binaries/cuprated/src/rpc/server.rs
```diff
@@ -142,15 +173,215 @@ async fn run_rpc_server(
...
+    excluded_ips: Vec<IpAddr>,
+    /// Amount of current connections per limited IP address.
+    per_ip_conn_count: HashMap<SocketAddr, usize>,
+    /// Total amount of current connections by limited IP addresses.
+    total_conn_count: usize,
+}
+
+impl RpcLimitCache {
+    fn new(
+        max_conn_count_public_ip: usize,
```
Boog900: Can we remove this

binaries/cuprated/src/rpc/server.rs
```diff
@@ -142,15 +173,215 @@ async fn run_rpc_server(
...
+            total_conn_count: 0,
+            per_ip_conn_count: HashMap::new(),
+        }
+    }
+
+    /// Initialize an RPC cache from an RPC configuration reference.
+    /// (The wrapping sub then + is to convert 0 into MAX)
+    fn from_config(config: &RpcConfig, restricted: bool) -> Self {
+        if restricted {
+            Self::new(
```
Boog900: Then can this just build manually

binaries/cuprated/src/rpc/timeout.rs
```diff
@@ -0,0 +1,296 @@
...
+use pin_project_lite::pin_project;
+use tokio::{
+    io::{AsyncRead, AsyncWrite},
+    time::{sleep_until, Instant, Sleep},
+};
+
+/// A timeout state with a specified duration.
+///
+/// `D` implements [`ExtractDuration`] trait which
+/// permit custom types to compute a Timeout duration.
```
Boog900: this needs to be removed

Boog900 (changes requested):
Partial review

binaries/cuprated/src/config/rpc.rs
```diff
@@ -101,6 +168,13 @@ impl Default for UnrestrictedRpcConfig {
             port: DefaultOrCustom::Default,
             enable: true,
             request_byte_limit: 0,
+            public_ip_connection_limit: 0,
+            private_ip_connection_limit: 0,
+            loopback_connection_limit: 0,
+            total_connection_limit: 0,
+            excluded_ips_connection_limit: Vec::new(),
+            send_timeout: Duration::from_secs(5),
+            read_timeout: Duration::from_secs(10),
```
Boog900: 10 is low IMO, I think 30 would be better

binaries/cuprated/src/rpc/server.rs
```diff
@@ -142,15 +174,202 @@ async fn run_rpc_server(
...
+                if *conn_count >= limit.get() {
+                    return false;
+                }
+            }
+
+            if let Some(max_conn_count) = self.max_conn_count {
+                if self.total_conn_count >= max_conn_count.get() {
+                    return false;
+                }
+            }
```
Boog900: This needs to be moved up and out of this `if let`

binaries/cuprated/src/rpc/timeout.rs
```diff
@@ -0,0 +1,290 @@
...
+    }};
+}
+
+pin_project! {
+    /// A timeout wrapper around an [`AsyncRead`] + [`AsyncWrite`] implemented type.
+    ///
+    /// Returns a `TimedOut` error if `poll_read` have been returning
+    /// `Poll::Pending` for the timeout duration.
+    pub struct StreamTimeout<S> {
+        #[pin]
```
Boog900: IMO the timer should not be reset after every _byte_ has been read, it should reset once clear

Boog900 (changes requested):

binaries/cuprated/src/config/rpc.rs
SyntheticBird45: ok

binaries/cuprated/src/rpc/server.rs
```diff
@@ -69,40 +87,54 @@ pub fn init_rpc_servers(launch_ctx: &LaunchContext, tx_handler: IncomingTxHandle
...
-            )
-            .await
-            .unwrap();
-            info!(restricted, "RPC server shut down.");
-        });
+        launch_ctx
+            .task_executor
+            .spawn(rpc_server.run(shutdown_token));
+
+        info!(restricted, "RPC server shut down.");
```
redsh4de: this line needs to be moved back in the spawn after awaiting the run, otherwise it will log on startup every time after spawn

```suggestion
        launch_ctx.task_executor.spawn(async move {
            if let Err(e) = rpc_server.run(shutdown_token).await {
                error!(restricted, "RPC server error: {e:#}");
            }

            info!(restricted, "RPC server shut down.");
        });
```

redsh4de (changes requested):

binaries/cuprated/src/rpc/concurrency.rs
```diff
@@ -0,0 +1,267 @@
...
+                let semaphore = Arc::clone(&self.semaphore);
+                let timeout = self.timeout;
+                let retry_after = Arc::clone(&self.retry_after);
+
+                ResponseFuture::Queued {
+                    future: Box::pin(async move {
+                        // Buffer the body, look at the method in the JSON payload.
+                        let (parts, body) = request.into_parts();
+
+                        let bytes = axum::body::to_bytes(body, usize::MAX).await.unwrap();
```
redsh4de: can't panic here, will kill the node

```suggestion
                        let Ok(bytes) = axum::body::to_bytes(body, usize::MAX).await else {
                              return Ok(Response::builder()
                                  .status(StatusCode::BAD_REQUEST)
                                  .body(Body::empty())
                                  .unwrap());
                         };
```

binaries/cuprated/src/rpc/ratelimit.rs
```diff
@@ -0,0 +1,347 @@
...
+                cache.debit_time(ip, millis);
+
+                // Count streamed body bytes, debiting the response size.
+                Poll::Ready(Ok(response.map(|body| {
+                    Body::new(DebitBody::new(body, Arc::clone(cache), *ip))
+                })))
+            }
+        }
+    }
+}
```
redsh4de: We can apply something like this in order to cover incomplete requests

```suggestion
/// The accounting of an in-flight request, debited from the IP address'
/// budgets once the request is served or dropped.
struct Debit {
    /// The instant the request entered the service.
    start: Instant,
    ip: IpAddr,
    cache: Arc<RpcRateLimitCache>,
}

impl Debit {
    /// Debit the response, consuming this `Debit`: its
    /// [`Drop`] pays for the processing time.
    fn settle(self, response: Response<Body>) -> Response<Body> {
        // A response of known size is debited in full, the node
        // has already paid to produce it.
        if let Some(size) = response.body().size_hint().exact() {
            self.cache
                .debit_size(&self.ip, i64::try_from(size).unwrap_or(i64::MAX));

            return response;
        }

        // Otherwise the bytes are counted as they stream out.
        let (ip, cache) = (self.ip, Arc::clone(&self.cache));
        response.map(|body| Body::new(DebitBody::new(body, cache, ip)))
    }
}

impl Drop for Debit {
    fn drop(&mut self) {
        let millis = i64::try_from(self.start.elapsed().as_millis()).unwrap_or(i64::MAX);
        self.cache.debit_time(&self.ip, millis);
    }
}

pin_project! {
    #[project = ResponseFutureProj]
    pub(super) enum ResponseFuture<F> {
        /// The request was rate limited: respond `429 Too Many Requests`.
        RateLimited,
        /// The request is being served, its processing time and response size
        /// measured and debited from the IP address' budgets.
        Future {
            #[pin]
            future: F,
            debit: Option<Debit>,
        },
    }
}

impl<F, E> Future for ResponseFuture<F>
where
    F: Future<Output = Result<Response<Body>, E>>,
{
    type Output = Result<Response<Body>, E>;

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        match self.project() {
            ResponseFutureProj::RateLimited => Poll::Ready(Ok(Response::builder()
                .status(StatusCode::TOO_MANY_REQUESTS)
                .body(Body::empty())
                .unwrap())),
            ResponseFutureProj::Future { future, debit } => {
                let response = ready!(future.poll(cx))?;

                // The request has finished processing. Time to debit its budget.
                Poll::Ready(Ok(match debit.take() {
                    Some(debit) => debit.settle(response),
                    None => response,
                }))
            }
        }
    }
}
```

binaries/cuprated/src/config/rpc.rs
```diff
@@ -58,6 +58,147 @@ config_struct! {
...
+        /// in milliseconds.
+        ///
+        /// Every IP address has a processing time budget: the processing
+        /// time of each of its requests is subtracted from it, and it is
+        /// refilled by `income_time` every second. Requests received
+        /// while this budget is exhausted are rejected with
+        /// `429 Too Many Requests`.
+        ///
+        /// Type         | Number (milliseconds)
+        /// Valid values | >= 0 (0 rejects all requests)
```
redsh4de: why not have 0 mean no limits?

itd be conflicting if rpc is enabled but nobody can make requests imo

binaries/cuprated/src/rpc/ratelimit.rs
```diff
@@ -0,0 +1,347 @@
...
+            time: budget.max_budget_time,
+            last_request: now,
+        });
+
+        // Refill the budgets by their respective incomes.
+        let elapsed =
+            i64::try_from(now.duration_since(entry.last_request).as_millis()).unwrap_or(i64::MAX);
+
+        let refill_size = budget.income_size * (elapsed / 1000);
+        let refill_time = budget.income_time * (elapsed / 1000);
```
redsh4de: multiplication before division

```suggestion
        let refill_size = budget.income_size.saturating_mul(elapsed) / 1000;
        let refill_time = budget.income_time.saturating_mul(elapsed) / 1000;
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -136,21 +224,251 @@ async fn run_rpc_server(
...
+    /// Consume this server and start serving to incoming connections.
+    /// This method only returns errors is unable to listen on its address:port.
+    async fn run(mut self, shutdown_token: CancellationToken) -> Result<(), Error> {
+        // Start the listener.
+        let listener = TcpListener::bind(self.listening_address).await?;
+
+        loop {
+            tokio::select! {
+                res = listener.accept() => {
+                    let (socket, remote_addr) = res?;
```
redsh4de: server could die on a failed accept:

```suggestion
                    let (socket, remote_addr) = match res {
                          Ok(res) => res,
                          Err(e) => {
                              // A failed accept must not take the server down.
                              warn!("Failed to accept RPC connection: {e}");
                              tokio::time::sleep(ACCEPT_BACKOFF).await;
                              continue;
                          }
                      };
```

redsh4de (changes requested):
Partial review

binaries/cuprated/src/config/rpc.rs
SyntheticBird45: Because I am retarded and I forgot to remove that doc back when I was frustrated with 0 to max conversions. Will fix that

binaries/cuprated/src/rpc/ratelimit.rs
SyntheticBird45: I don't really like this. This adds another struct and that becomes confusing with DebitBody imo. This is also confusing to me. You are saying this would account for incomplete request but this is only debiting the response. I don't see a difference in logic here.

redsh4de: Yeah the logic is the same on request completion

the difference is when the request is not completed: if the client drops mid-request we never reach `cache.debit_time`, and `DebitBody` is never built either, so nothing gets charged. The `Drop` impl takes care of that

naming could be better though

SyntheticBird45: oh got it. make sense

binaries/cuprated/src/rpc/server.rs
```diff
@@ -136,21 +171,99 @@ async fn run_rpc_server(
...
+                .timer(TokioTimer::new())
+                .header_read_timeout(header_read_timeout);
+
+            // Serve RPC router to this connection
+            if let Err(err) = builder
+                .serve_connection(TokioIo::new(socket), TowerToHyperService::new(rpc))
+                .await
+            {
+                error!("Failed to serve RPC connection: {err:#}");
+            }
```
redsh4de: preserves graceful shutdown

```suggestion
                    self.serve(socket, remote_addr, shutdown_token.clone());
                },
                Some(res) = self.rpc_tasks.join_next() => {
                    if let Err(err) = res {
                        error!("RPC connection task failed: {err:#}");
                    }
                }
                () = shutdown_token.cancelled() => {
                    break;
                }
            }
        }

        while let Some(res) = self.rpc_tasks.join_next().await {
            if let Err(err) = res {
                error!("RPC connection task failed: {err:#}");
            }
        }

        Ok(())
    }

    fn serve(
        &mut self,
        socket: TcpStream,
        remote_addr: SocketAddr,
        shutdown_token: CancellationToken,
    ) {
        let rpc = self.rpc.clone();
        let send_timeout = self.send_timeout;
        let header_read_timeout = self.header_read_timeout;

        self.rpc_tasks.spawn(async move {
            let socket = WriteTimeout::new(socket, send_timeout);

            let mut builder = http1::Builder::new();
            builder
                .timer(TokioTimer::new())
                .header_read_timeout(header_read_timeout);

            let connection =
                builder.serve_connection(TokioIo::new(socket), TowerToHyperService::new(rpc));
            tokio::pin!(connection);

            tokio::select! {
                result = &mut connection => {
                    if let Err(err) = result {
                        error!("Failed to serve RPC connection: {err:#}");
                    }
                }
                () = shutdown_token.cancelled() => {
                    connection.as_mut().graceful_shutdown();

                    if let Err(err) = connection.await {
                        error!("Failed to shut down RPC connection: {err:#}");
                    }
                }
            }
```

redsh4de (changes requested):

binaries/cuprated/src/rpc/timeout.rs
```diff
@@ -0,0 +1,310 @@
...
+        let len = $len;
+
+        if len == 0 {
+            proj.$io.$poll($cx, $arg)
+        } else if let Poll::Ready(error) = proj.$timeout.as_mut().poll($cx) {
+            Poll::Ready(Err(error))
+        } else {
+            match proj.$io.$poll($cx, $arg) {
+                Poll::Ready(Ok(written)) => {
+                    if written == len {
```
redsh4de, line 74: Using `written == len` means the deadline never re-arms once the socket send buffer fills, so the whole remaining body has to drain within `send_timeout`. Slow clients get cut off mid-response


```suggestion
                    if written > 0 {
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -136,21 +176,129 @@ async fn run_rpc_server(
...
+                .header_read_timeout(header_read_timeout);
+
+            let connection =
+                builder.serve_connection(TokioIo::new(socket), TowerToHyperService::new(rpc));
+            tokio::pin!(connection);
+
+            tokio::select! {
+                result = &mut connection => {
+                    if let Err(err) = result {
+                        error!("Failed to serve RPC connection: {err:#}");
```
redsh4de: mb, these should prob be debug as they are remotely triggerable, can cause a log flood


```suggestion
                        debug!("Failed to serve RPC connection: {err:#}");
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -136,21 +176,129 @@ async fn run_rpc_server(
...
+                    }
+                }
+                () = shutdown_token.cancelled() => {
+                    connection.as_mut().graceful_shutdown();
+
+                    if let Err(err) = connection.await {
+                        error!("Failed to shut down RPC connection: {err:#}");
```
redsh4de:
```suggestion
                        debug!("Failed to shut down RPC connection: {err:#}");
```

binaries/cuprated/src/config/rpc.rs
```diff
@@ -58,6 +58,28 @@ config_struct! {
...
+        ///
+        /// Type     | Duration
+        /// Examples | { secs = 10, nanos = 0 }, { secs = 29, nano = 123 }
+        pub header_read_timeout: Duration,
+
+        /// The maximum wall-clock time allowed to consume an HTTP request body.
+        /// Receiving progress does not restart this deadline.
+        ///
+        /// Type     | Duration
+        /// Examples | { secs = 10, nanos = 0 }, { secs = 29, nano = 123 }
```
redsh4de: `nano` -> `nanos` in the last examples across all 3

binaries/cuprated/src/rpc/server.rs
```diff
@@ -136,21 +176,129 @@ async fn run_rpc_server(
...
+            while let Some(res) = self.rpc_tasks.join_next().await {
+                if let Err(err) = res {
+                    error!("RPC serving task failed: {err:#}");
+                }
+            }
+        })
+        .await
+        .is_err()
+        {
+            warn!("RPC tasks survived shutdown signal for more than 5 seconds... Dropping connections anyway.");
```
redsh4de:
```suggestion
            warn!("RPC tasks survived shutdown signal for more than {RPC_SHUTDOWN_TIMEOUT:?} seconds... Dropping connections anyway.");
```

redsh4de (changes requested):

binaries/cuprated/src/rpc/timeout.rs
Boog900: That is the whole point, we want slow clients to disconnect. Otherwise a client can juts read 1 byte and the timer will reset.

redsh4de (approved):

binaries/cuprated/src/rpc/server.rs
```diff
@@ -69,41 +91,59 @@ pub fn init_rpc_servers(launch_ctx: &LaunchContext, tx_handler: IncomingTxHandle
...
-                restricted,
-                SocketAddr::new(addr, port),
-                request_byte_limit,
-                shutdown_token,
-            )
-            .await
-            .unwrap();
+            if let Err(e) = rpc_server.run(shutdown_token).await {
+                error!(restricted, "RPC server error: {e:#}");
+            }
```
Boog900, line 133: Do we really want to leave the node running here?

binaries/cuprated/Cargo.toml
Boog900: We should remove the http2 feature on axum

Boog900 (changes requested):

binaries/cuprated/src/rpc/server.rs
SyntheticBird45: Not quite sure either but if we want to shutdown on listener bind failure then we would have to move TcpListener outside of this task, which would transform init_rpc_servers into an async fn. Since the task is stored in task_executor I believe it is more suited for the graceful shutdown pr to join all tasks and shutdown on failure.

Boog900: we can leave it as a TODO

Boog900 (approved):

PR #677, opened 2026-08-09
epee-encoding: fix memory DoS

Boog900:
Reported by: @xmrack

This did take a little bit of time to fix, I didn't like any of the obvious ways to fix this. I think what I have gone with is pretty nice.

### DIFF
--- a/net/wire/src/p2p/protocol.rs
+++ b/net/wire/src/p2p/protocol.rs
@@ -25,6 +25,9 @@ use cuprate_fixed_bytes::{ByteArray, ByteArrayVec};
 
 use crate::p2p::common::BlockCompleteEntry;
 
+/// This is a protocol level limit on the [`GetObjectsResponse`] message.
+const MAX_BLOCKS_IN_RESPONSE: usize = 100;
+
 /// A block that SHOULD have transactions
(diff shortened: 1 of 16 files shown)


net/wire/src/p2p/protocol.rs
```diff
@@ -53,7 +56,9 @@ pub struct NewTransactions {
 
 epee_object!(
     NewTransactions,
-    txs: Vec<Bytes>,
+    // 32 is the size of `Bytes` on the stack, no tx can be this small. We could set a limit
+    // higher than 32 but 32 gives us enough protection.
+    txs: Vec<Bytes> [min_element: 32],
```
Boog900, line 61: This .....

net/wire/src/p2p/protocol.rs
```diff
@@ -86,7 +91,7 @@ pub struct GetObjectsResponse {
 
 epee_object!(
     GetObjectsResponse,
-    blocks: Vec<BlockCompleteEntry>,
+    blocks: Vec<BlockCompleteEntry> [max_len: MAX_BLOCKS_IN_RESPONSE],
```
Boog900, line 94: and this is where the main fix is

SyntheticBird45 (approved):

PR #679, opened 2026-08-10
P2P: simplify code

Boog900:
This PR simplifies our connection code by removing the different types of client, keeping just one. It also removes the shared error.

Connections are dropped once the last client is dropped, as they close once the mpsc request channel is closed.


SyntheticBird45 (approved):
That `ServiceExt::ready` sure is handy. LGTM

PR #681, opened 2026-08-11
pruning: add pruning option

### LINKED ISSUE #657: Pruning for Cuprate
This issue documents a way to add pruning to Cuprate.

# Setting a pruned node.

monerod supports pruning a node while it is running. I don't think we need to support that. The 2 ways to make a node pruned should be a command line arg (i.e `--prune`) and a new config value in the main config file (`prune = true`).

Once a node is pruned it cannot be un-pruned so setting the prune config value to `prune = false` should do nothing. 

# Keeping track of pruning state

A metadata file should be added to the DB. This metadata file will hold 2 things: the database version and the pruning seed. 

This file should be located in the DB directory, its format should just be 12 bytes, the first 8 being the version, the next 4 the pruning seed. In `BlockchainDatabase::open_with_fjall_database` this file should be loaded before doing anything else, if not present the values of the version and pruning seed should be set to their default and a file should be created.

If the metadata file says the DB is not pruned and the config says it should be then prune the DB.

# Pruning the DB

Firstly we should change the prunable tapes type here: https://github.com/Cuprate/cuprate/blob/d57d59dde47125b80bc87c1b5cdcf200fccc78cd/storage/blockchain/src/database.rs#L135

to:

```rust
    pub(crate) prunable_blobs: Vec<Option<tapes::BlobTape>>,
```

When pruning we need to randomly generate a pruning stripe, we need to then update the metadata file with this stripe before doing anything. We then delete the tapes file for all other stripes from the one we selected. 

From the stripe we generate a pruning seed, which we then expose to the rest of Cuprate so it can pass it to P2P etc.

Where we currently use `prunable_blobs` we need to handle the case where data is requested for a tape we don't have, most likely we should just return an error.

# Tip blocks

monerod will store the top 5500 blocks unpruned, even if they are in a stripe we should prune, we need to handle this. 
A way to handle this is to make a new fjall keyspace for the prunable blobs of the txs in the top 5500 blocks and then when a new block is added we look at the block 5501 from the top and remove its txs from the new keyspace. 

All V2 prunable blobs should be added to this keyspace, if we are pruning, for simplicity. We need to fill this table in before pruning a non-pruned node too.

All places that look for prunable blobs will need to look in this new keyspace if they are close to the top of the chain.

# Make consistent

`BlockchainDatabase::make_consistent` is the way we currently make sure the tapes and fjall are in sync on startup. With pruning this is harder as prunable data could be missing from a new block in the tapes so cannot be added to fjall.

To fix this we need to pop some blocks from the tapes, up to where we have enough data. This probably looks like just always popping the top 5500 blocks from the tapes if we are pruned and the databases are out of sync. Then just having the tip prunable blob keyspace empty to repopulate with blocks once they are downloaded again.

p4xxus:
### What

Closes #657 

- a new metadata file stores the pruning seed, which is accessible via the `db` struct
- a new db entry called `prunable_tip` contains the transactions of the top 5500 blocks by TxId to blob
- if pruned:
  - new blocks are added in the prunable tip
  - retrieval of txs and blocks searches in this tip

### Why

to use less space for a monero node

### Where

- DB: 2 new fields (`prunable_tip` (`Keyspace`) and `metadata` (`Metadata`)), updated `prunable_blob` to use optional tape for prunable part
- config options
- cli argument
- adding and retrieving blocks as well as transactions
- `ContextCache`
- persited file structure: new metadata file containing pruning seed and db version (= 0)

### How

- a new metadata file stores the pruning seed
  - the seed is accessible via the `db` struct
  - it's populated via a config option in the config file or via a cli arg
- if the metadata file says not prune but the config / cli arg says prune, create a random seed, add the top 5500 blocks to the prunable tip and delete the other tapes

incoming block:
- if we prune:
  - only add the blob to the desired stripe
  - remove old txs in the `prunable_tip` so after adding the new blocks we're back to 5500 blocks
  - add the new blocks to the `prunable_tip`

retrieval of tx (`read_prunable_tape`):
- if we prune:
  - if the tx is in the top 5500 blocks, look in the prunable_tip
  - otherwise, return NotFound error

I used Gemini Pro 3.1 to find bugs and general help. It was thoroughly reviewed though.


binaries/cuprated/src/config/storage.rs
```diff
@@ -81,6 +81,11 @@ config_struct! {
     #[serde(deny_unknown_fields, default)]
     pub struct BlockchainConfig {
         #[comment_out = true]
+        /// Whether to prune the blockchain. This reduces the size of the blockchain by not keeping all the transactions.
```
Boog900:
```suggestion
        /// Whether to prune the blockchain. This reduces the size of the blockchain by removing most transaction proof data.
```

consensus/context/src/lib.rs
```diff
@@ -137,6 +137,8 @@ pub struct BlockchainContext {
     pub cumulative_difficulty: u128,
     /// Context to verify a block, as needed by [`cuprate_consensus_rules`]
     pub context_to_verify_block: ContextToVerifyBlock,
+    /// The pruning seed for the blockchain.
+    pub pruning_seed: u32,
```
Boog900: Can we make this the type from the cuprate_pruning crate

consensus/context/src/task.rs
```diff
@@ -62,6 +62,8 @@ pub(crate) struct ContextTask<D: Database> {
     top_block_hash: [u8; 32],
     /// The total amount of coins generated.
     already_generated_coins: u64,
+    /// The pruning seed.
+    pruning_seed: u32,
```
Boog900: and here

p2p/p2p-core/src/client/request_handler.rs
```diff
@@ -152,6 +152,12 @@ where
             panic!("Address book sent incorrect response!");
         };
 
+        let pruning_seed = if core_sync_data.pruning_seed == 0 {
+            PruningSeed::NotPruned
+        } else {
+            PruningSeed::decompress(core_sync_data.pruning_seed).expect("Our pruning seed is valid")
+        };
```
Boog900: `PruningSeed::decompress` handles the 0 case

storage/blockchain/src/metadata.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Gets the stripe index from the pruning seed.
+    ///
+    /// Note that this doesn't re-read the file.
+    ///
+    /// Returns `None` if the seed is not pruned (e.g. the seed is `0`).
+    #[inline]
+    pub const fn get_stripe_idx(&self) -> Option<u32> {
+        self.pruning_seed.get_stripe()
+    }
```
Boog900: This is not needed as we can just use the `pruning_seed` from the other function

storage/blockchain/src/metadata.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    #[inline]
+    pub const fn get_db_version(&self) -> u64 {
+        self.db_version
+    }
+
+    #[inline]
+    pub fn is_pruned(&self) -> bool {
+        self.pruning_seed != PruningSeed::NotPruned
+    }
```
Boog900: This isn't needed too

types/types/src/blockchain.rs
```diff
@@ -227,6 +228,9 @@ pub enum BlockchainReadRequest {
 
     /// Get the output indexes of a transaction.
     TxOutputIndexes { tx_hash: [u8; 32] },
+
+    /// Get the pruning seed of the blockchain.
+    PruningSeed,
```
Boog900: The service interface is a lot of overhead for this, IMO it should just be stored in the database struct with a getter function.

storage/blockchain/src/service/write.rs
```diff
@@ -175,7 +176,14 @@ fn write_blocks(db: &BlockchainDatabase, blocks: &[VerifiedBlockInformation]) ->
...
 
-    tx_rw.commit()?;
+    tx_rw.commit()?; // TODO: This can be done with one go: by calculating the tx_ids required for the `add_blocks_to_prunable_tip` with the blocks we already have. (both for removal and addition of tip blocks)
+
+    if db.pruning_seed != PruningSeed::NotPruned {
+        let mut tx_rw = db.fjall.batch().durability(Some(PersistMode::Buffer));
+        let tapes = db.linear_tapes.reader();
+        add_blocks_to_prunable_tip(db, blocks, &tapes, &mut tx_rw)?;
+        tx_rw.commit()?;
+    }
```
Boog900: This breaks atomic writes, you should move the logic from `add_blocks_to_prunable_tip` into `add_block_to_dynamic_tables`

storage/blockchain/src/ops/block.rs
```diff
@@ -338,6 +346,66 @@ pub fn add_block_to_dynamic_tables<'a, I: Iterator<Item = Cow<'a, Transaction<Pr
     Ok(())
 }
 
+pub fn add_blocks_to_prunable_tip(
+    db: &BlockchainDatabase,
+    blocks: &[VerifiedBlockInformation],
+    tapes: &impl TapesRead,
+    w: &mut fjall::OwnedWriteBatch,
+) -> DbResult<()> {
```
Boog900: IMO this should be broken up into 2 ...

storage/blockchain/src/ops/block.rs
```diff
@@ -338,6 +346,66 @@ pub fn add_block_to_dynamic_tables<'a, I: Iterator<Item = Cow<'a, Transaction<Pr
...
+            let tx_to_remove_to = get_block(&keep_from_height, None, tapes, db)?
+                .miner_transaction()
+                .hash();
+            let tx_id_to_remove_to = get_tx_id_from_hash(db, &tx_to_remove_to)?;
+
+            for tx_id in tx_id_to_remove_from..tx_id_to_remove_to {
+                w.remove(&db.prunable_tip, tx_id.to_be_bytes().as_slice());
+            }
+        }
+    }
```
Boog900: This should be moved to: `add_block_to_dynamic_tables`. We should calculate the `tx_id_to_remove_from..tx_id_to_remove_to` outside the function call and pass it in like we do for `numb_transactions`. Here is how you can do it with just the tapes:

```rust
tapes.read_entry(
                    &db.block_infos,
                    usize_to_u64(/*height we want to remove from*/),
                )?.map_or(0, |prev| prev.mining_tx_index)
..
tapes.read_entry(
                    &db.block_infos,
                    usize_to_u64(/*lowest height we want to keep*/),
                )?.map_or(0, |prev| prev.mining_tx_index)
```

storage/blockchain/src/ops/block.rs
```diff
@@ -338,6 +346,66 @@ pub fn add_block_to_dynamic_tables<'a, I: Iterator<Item = Cow<'a, Transaction<Pr
...
+                .ok_or(BlockchainError::NotFound)?
+                .as_ref()
+                .get_u64_le();
+            w.insert(
+                &db.prunable_tip,
+                tx_id.to_be_bytes(),
+                tx.tx_prunable_blob.as_slice(),
+            );
+        }
+    }
```
Boog900: This should be put into `add_tx_info_to_dynamic_tables`. That already has the `tx_id`. We should just add every tx if pruning. When writing blocks we wont every write more than `CRYPTONOTE_PRUNING_TIP_BLOCKS`

Boog900 (changes requested):
Thanks for working on this. The code looks great. This is only an initial review but I do have some suggestions.

storage/blockchain/src/ops/block.rs
p4xxus: isn't this much more inefficient?
Currently I'm removing from the lowest TxId in the `prunable_tip` up until the calculated one, having in total basically 2 db lookup (block and TxId retrieval, although it can be brought down to 1 by reading in block_infos as you said here). 

If I'd do as you suggested, I would need to calculate the lowest height we want to keep for each block, (I can still delete from the lowest txid upwards), thus having 1 DB lookup per Block.
Or am I missing sth?

types/types/src/blockchain.rs
p4xxus: I don't quite understand. How can I get access here to the pruning seed via another way?

storage/blockchain/src/ops/block.rs
Boog900: hmm yeah `add_block_to_dynamic_tables` takes each block individually. I was thinking we can just calculate the whole range to remove across all blocks. 

I think we should add a new function `add_blocks_to_dynamic_tables` that just calls `add_block_to_dynamic_tables` but does this range removal first and then call that new function in the `write_blocks` function.

When rebuilding fjall you will still need to use the current `add_block_to_dynamic_tables` but that should be ok as we don't necessarily have the pruning data anyway (see here: https://github.com/Cuprate/cuprate/issues/657 `Make consistent`).

types/types/src/blockchain.rs
Boog900: Like this:
```rust
impl BlockchainDatabase {
    pub fn pruning_seed(&self) -> PruningSeed {
    }
}
```

Boog900 (approved):

redsh4de (approved):

PR #684, opened 2026-08-13
RPC: Connection limit and fail-ban protection (hardening pt.2)

SyntheticBird45:
### What

This PR follows and is based upon https://github.com/Cuprate/cuprate/pull/675

It adds per-IP type simultaneous connection limits similar to https://github.com/monero-project/monero/pull/9765. This PR extends this mitigation by adding a distinction between loopback addresses and private addresses, as well as enabling user to exclude specific IP addresses from any restriction if needed (main production case would be reverse proxies).

This PR also adds a short-term ban mechanism for IP addresses that repeatedly fails to be served by the hyper stack (axum erros are unaccounted). Said banned IP addresses are immediately dropped until they are unbanned.


binaries/cuprated/src/rpc/server.rs
```diff
@@ -236,19 +260,178 @@ impl RpcServer {
...
+    }
+
+    fn is_ip_banned(&mut self, remote: IpAddr) -> bool {
+        if let Some(counter) = self.banned_ips.get(&remote) {
+            *counter >= 15
+        } else {
+            false
+        }
+    }
+
```
redsh4de, line 502: Add a `is_excluded` accessor so that the exclusion check can move to call sites:

```suggestion
    /// Whether this IP address is excluded from all limits.
    fn is_excluded(&self, remote_addr: IpAddr) -> bool {
        self.excluded_ips.contains(&remote_addr)
    }
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -221,10 +225,30 @@ impl RpcServer {
...
 
+                    // Check if IP is banned
+                    if self.ip_limit_cache.is_ip_banned(remote_addr.ip()) {
+                        continue;
+                    }
+
+                    // Check for total and per-IP connection limits.
+                    if !self.ip_limit_cache.check_and_track_connection(remote_addr.ip()) {
+                        continue;
+                    }
```
redsh4de: The fail-ban path never checks `excluded_ips`, so a reverse proxy - the one IP every client's failures funnel through - is the first thing banned. Gate both `select!` arms instead:

```suggestion
                    let ip = remote_addr.ip();

                    // Excluded IPs bypass the ban and all connection limits.
                    if !self.ip_limit_cache.is_excluded(ip) {
                        // Check the ban, then the total and per-IP connection limits.
                        if self.ip_limit_cache.is_ip_banned(ip)
                            || !self.ip_limit_cache.check_and_track_connection(ip)
                        {
                            continue;
                        }
                    }
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -221,10 +225,30 @@ impl RpcServer {
...
+
                     self.serve(socket, remote_addr);
                 },
                 Some(res) = self.rpc_tasks.join_next() => {
-                    let res = res?;
+                    let (addr, serve_failed) = res?;
+                    let ip = addr.ip();
+
+                    // Untrack connection from IP
+                    self.ip_limit_cache.remove_connection(&ip);
```
redsh4de, line 283:
```suggestion
                    if self.ip_limit_cache.is_excluded(ip) {
                        continue;
                    }

                    // Untrack connection from IP
                    self.ip_limit_cache.remove_connection(&ip);
```

With the gate in both arms, the internal checks in `check_and_track_connection` and `remove_connection` become redundant and can be removed

binaries/cuprated/src/rpc/server.rs
```diff
@@ -236,19 +260,178 @@ impl RpcServer {
...
+    }
+
+    /// Check the connection limit of the remote IP address, increase it and return `true` if allowed,
+    /// return `false` if disallowed. Excluded IP address is validated early and do not increase the total
+    /// connection count.
+    fn check_and_track_connection(&mut self, remote_addr: IpAddr) -> bool {
+        if self.excluded_ips.contains(&remote_addr) {
+            return true;
+        }
+
```
redsh4de:
```suggestion
    /// Check the connection limit of the remote IP address, increase it and
    /// return `true` if allowed, `false` if disallowed.
    ///
    /// Excluded IP addresses must be filtered out by the caller.
    fn check_and_track_connection(&mut self, remote_addr: IpAddr) -> bool {
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -236,19 +260,178 @@ impl RpcServer {
...
+        self.total_conn_count += 1;
+
+        true
+    }
+
+    /// Remove a connection of the remote IP address.
+    fn remove_connection(&mut self, remote_addr: &IpAddr) {
+        if self.excluded_ips.contains(remote_addr) {
+            return;
+        }
```
redsh4de:
```suggestion
    /// Remove a connection of the remote IP address.
    ///
    /// Excluded IP addresses must be filtered out by the caller.
    fn remove_connection(&mut self, remote_addr: &IpAddr) {
```

redsh4de (changes requested):

redsh4de:
The counter never decays, so this is a 3s ban every 15th failure over the nodes whole uptime. Maybe worth bounding the strikes to a window of a minute or so?

SyntheticBird45:
@redsh4de , I actually changed the semantic a little bit. There is now two `DelayQueue`, one for evicting the cache if the IP has stopped causing failure, and another one for if a banned IP's sentence is over.

There is no explicit list of what IP addresses are currently banned because looking up into DelayQueue is O(n) while in practice keeping the current HashMap O(1) (in most cases). So an IP address is *effectively* banned if its failure reached `RPC_FAILURE_BAN_THRESHOLD`. That is what `is_ip_banned` checks.

`banned_ips` has also been turned into a `per_ip_failures` `HashMap` since we don't need concurrency anymore.

Constants can be tweaked but at the moment, an IP address will keep its failure count for 5 second, this timer is reset everytime its increased, and erased when the IP is added to the unban queue.

redsh4de (approved):

redsh4de:
Needs rebase

binaries/cuprated/src/rpc/server.rs
```diff
@@ -233,10 +257,50 @@ impl RpcServer {
                         }
                     };
 
+                    let ip = remote_addr.ip();
```
Boog900: I think this is needed
```suggestion
                    let ip = remote_addr.ip().to_canonical();
```

binaries/cuprated/src/rpc/server.rs
```diff
@@ -233,10 +257,50 @@ impl RpcServer {
...
+
+                    // Excluded IPs bypass ban and connection limits
+                    if !self.ip_limit_cache.is_excluded(ip) {
+                        // Check if IP is banned and connection limits.
+                        if self.ip_limit_cache.is_ip_banned(remote_addr.ip())
+                        || !self.ip_limit_cache.check_and_track_connection(remote_addr.ip())
+                        {
+                            continue;
```
Boog900, line 268: We should probably log here when this happens

SyntheticBird45:
Rebased

binaries/cuprated/src/rpc/server.rs
```diff
@@ -180,25 +202,40 @@ fn init_rpc_router(
...
         }
     }
 
-    /// Consume this server and start serving connections incoming on `listener`.
-    async fn run(mut self, listener: TcpListener, shutdown_token: CancellationToken) {
+    /// Consume this server and start serving to incoming connections.
+    /// This method only returns errors if it is unable to listen on its address:port.
+    async fn run(mut self, shutdown_token: CancellationToken) -> Result<(), Error> {
+        // Start the listener.
+        let listener = TcpListener::bind(self.listening_address).await?;
```
redsh4de: Bind failure here won't be fatal anymore, regresses #688

I'd revert the restructure to keep init_rpc_servers returning Result like it currently is on main and binding there, & passing listener into run like prior

binaries/cuprated/src/rpc/server.rs
```diff
@@ -79,7 +96,7 @@ pub(crate) async fn init_rpc_servers(
                     "Starting unrestricted RPC on non-local address, this is dangerous!"
                 );
             } else {
-                anyhow::bail!("Refusing to start unrestricted RPC on a non-local address ({addr})");
+                panic!("Refusing to start unrestricted RPC on a non-local address ({addr})");
```
redsh4de: with that would keep this to be `anyhow::bail!` as it'll shut down the process gracefully

binaries/cuprated/src/rpc/server.rs
```diff
@@ -99,33 +116,38 @@ pub(crate) async fn init_rpc_servers(
...
+            rpc_send_timeout,
+            rpc_header_read_timeout,
+            rpc_limit_cache,
+            rpc_router,
+        );
 
         info!(
             restricted,
-            address = %address,
+            address = %addr,
```
redsh4de, line 140: switching to `addr` will log without the port, not sure if intentional

binaries/cuprated/src/rpc/server.rs
```diff
@@ -213,10 +250,51 @@ impl RpcServer {
                         }
                     };
 
+                    let ip = remote_addr.ip().to_canonical();
+
+                    // Excluded IPs bypass ban and connection limits
+                    if !self.ip_limit_cache.is_excluded(ip) {
+                        // Check if IP is banned and connection limits.
+                        if self.ip_limit_cache.is_ip_banned(remote_addr.ip())
+                        || !self.ip_limit_cache.check_and_track_connection(remote_addr.ip())
```
redsh4de:
```suggestion
                        if self.ip_limit_cache.is_ip_banned(ip)
                        || !self.ip_limit_cache.check_and_track_connection(ip)
```

redsh4de (changes requested):

binaries/cuprated/src/rpc/server.rs
SyntheticBird45: its not. I added `port`

binaries/cuprated/src/rpc/server.rs
```diff
@@ -213,10 +256,51 @@ impl RpcServer {
...
+                        }
+                    }
+
                     self.serve(socket, remote_addr, shutdown_token.clone());
                 },
-                Some(Err(err)) = self.rpc_tasks.join_next() => {
-                    debug!("RPC serving task failed: {err:#}");
+                Some(res) = self.rpc_tasks.join_next() => {
+                    let (addr, serve_failed) = res?;
+                    let ip = addr.ip();
```
redsh4de, line 276: drift from .to_canonical()

```suggestion
                    let ip = addr.ip().to_canonical();
```

We could either fix it like that or just key by `IpAddr` in rpc_tasks directly instead of `SocketAddr` to avoid the hazard altogether

redsh4de (changes requested):

redsh4de (approved):

binaries/cuprated/src/rpc/server.rs
```diff
@@ -99,27 +120,36 @@ pub(crate) async fn init_rpc_servers(
...
+            rpc_send_timeout,
+            rpc_header_read_timeout,
+            rpc_limit_cache,
+            rpc_router,
+        );
 
         info!(
             restricted,
-            address = %address,
+            address = %addr,
```
redsh4de, line 140:
```suggestion
            address = %addr,
            port = %port,
```

redsh4de (approved):

binaries/cuprated/src/rpc/server.rs
```diff
@@ -32,16 +41,27 @@ use crate::{
...
-/// # Errors
-///
-/// This function will return an [`Err`] if unrestricted RPC is started on a
-/// non-local address without the override option, or if an RPC listener cannot
-/// be bound.
+/// # Panics
+/// This function will panic if:
+/// - the server(s) could not be started
+/// - unrestricted RPC is started on non-local
+///   address without override option
```
Boog900, line 61: Aren't these errors now?

Boog900 (approved):

binaries/cuprated/src/rpc/server.rs
redsh4de: Yea, needs to b reverted

Boog900 (approved):

PR #685, opened 2026-08-13
RPC: Request rate limiting based on size/process time budget (hardening pt.3)

SyntheticBird45:
### What

This PR follows and is based upon https://github.com/Cuprate/cuprate/pull/684

It adds a new Axum middleware for rate limiting requests based on a processing time and response size budget. Each clients (IP address) is allocated a configurable process time budget and response size budget. These budgets are substracted when response is processed. Whenever the budget is below zero the clients is greeted with a `429 Too many requests` error. Budgets are refilled following an income variable.

The reasoning behind this system is limiting clients from using resource heavy requests, while permitting the quick use of cheap requests.

Original idea from @Boog900

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1544,6 +1544,8 @@ dependencies = [
  "futures",
  "hex",
  "hex-literal 1.1.0",
+ "hyper",
+ "hyper-util",
  "indexmap 2.14.0",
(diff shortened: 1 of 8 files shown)


redsh4de:
Rebase needed

PR #678, opened 2026-08-09
DB: add config for persistence

Boog900:

PR #693, opened 2026-08-24
cuprated: allow batch writing full verification batches

Boog900:
### What
 This PR caused a regression in full verification block sync speed: https://github.com/Cuprate/cuprate/pull/678, we always wrote fully verified blocks 1 by 1 so during full verification we would always be using the safest DB settings.
 
 This PR changes full verification block sync to allow us to batch write the blocks to DB. This should make us even faster than before the regression.

### DIFF
--- a/binaries/cuprated/src/blockchain/manager/handler.rs
+++ b/binaries/cuprated/src/blockchain/manager/handler.rs
@@ -256,6 +256,7 @@ impl super::BlockchainManager {
         let mut verification_context =
             VerificationContext::<ConsensusBlockchainReadHandle>::BatchPrepareCache(output_cache);
         let mut valid_blocks = Vec::with_capacity(prepped_blocks.len());
+        let mut error = false;
 
         for (block, txs) in prepped_blocks {
             let hash = block.block_hash;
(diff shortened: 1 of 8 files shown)


consensus/src/lib.rs
```diff
@@ -37,6 +37,18 @@ pub use cuprate_types::{
...
+#[expect(clippy::large_enum_variant)]
+pub enum VerificationContext<D> {
+    /// A full database
+    Database(D),
+    /// A batch prepared cache.
+    ///
+    /// This cache is only valid for the set of blocks it was created with, it should not be used for other blocks.
+    /// You must pass blocks in sequentially.
+    BatchPrepareCache(BatchPrepareCache),
+}
```
Boog900, line 50: This type is used instead of giving both the DB and an optional cache. I think it makes the code more safe, as it prevents someone from reading an out of date DB now that we don't write each block.

binaries/cuprated/src/blockchain/manager/handler.rs
redsh4de: it looks like the cache and database could de-sync if a batch contains a invalid block

we add the blocks to the cache until success, but on error we return out of the function without cleaning up the cache. Meanwhile we commit the blocks to db only if all the blocks are valid

So we should either commit the valid blocks that we have verified so far to db until the failure, or roll back the cache on error

redsh4de (changes requested):

binaries/cuprated/src/blockchain/manager/handler.rs
redsh4de: the quick fix would be to replace the return with a break so it falls through to the rest of the function

redsh4de (approved):

PR #697, opened 2026-08-30
DB: add a tx interface and change appropriate call sites

### LINKED ISSUE #539: Make the database service have a Tx interface
## What
Currently the database services have a interface that allows a single request at a time, creating a db tx each time it is called. Instead you should first need to call a function to start a DB tx before doing any calls.

## Why
Makes keeping the DB state consistent easier, often multiple requests are needed and we need to make sure each request stays in sync. 

## Where
the database service and and users.

Boog900:
closes #539

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -202,7 +202,7 @@ serde_json            = { version = "1", default-features = false }
 serde                 = { version = "1", default-features = false }
 strum                 = { version = "0.28", default-features = false }
 sysinfo               = { version = "0.38.4", default-features = false }
-tapes                 = { git = "https://github.com/Cuprate/Tapes.git", rev = "f28b6fb" }
+tapes                 = { git = "https://github.com/Cuprate/Tapes.git", rev = "001da519" }
 thiserror             = { version = "2", default-features = false }
 thread_local          = { version = "1", default-features = false }
 tikv-jemallocator     = { version = "0.7", default-features = false }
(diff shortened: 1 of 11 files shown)


redsh4de:
Rebase needed

storage/blockchain/src/service/read.rs
```diff
@@ -1350,8 +1694,18 @@ fn transactions(db: &BlockchainDatabase, tx_hashes: Vec<[u8; 32]>) -> ResponseRe
...
+    txs: Option<&ReadTransactions>,
+) -> BlockchainResponse {
+    let local_tapes;
+    let tapes = if let Some(txs) = txs {
+        let (_, tapes) = txs;
+        tapes
+    } else {
+        local_tapes = db.linear_tapes.reader();
+        &local_tapes
+    };
```
SyntheticBird45, line 1708: Is there really no way to avoid this being repeated ?

Cargo.toml
```diff
@@ -202,7 +202,7 @@ serde_json            = { version = "1", default-features = false }
 serde                 = { version = "1", default-features = false }
 strum                 = { version = "0.28", default-features = false }
 sysinfo               = { version = "0.38.4", default-features = false }
-tapes                 = { git = "https://github.com/Cuprate/Tapes.git", rev = "f28b6fb" }
+tapes                 = { git = "https://github.com/Cuprate/Tapes.git", rev = "001da519" }
```
SyntheticBird45, line 205: `04562f4`

PR #703, opened 2026-09-08
cuprated: stop the block downloader when shutting down

redsh4de:
### What
Ensures the block downloader task is stopped as the node is shutting down to prevent a panic on graceful shutdown

### Where
`cuprated`, `p2p`

### How
Return the spawned downloader task and abort it as we break out of the syncer loop when shutting down


binaries/cuprated/src/blockchain/syncer.rs
```diff
@@ -120,16 +120,14 @@ impl BlockchainSyncer {
...
-            let mut block_batch_stream =
+
+            let (mut block_batch_stream, downloader_task) =
                 clearnet_interface.block_downloader(our_chain.clone(), block_downloader_config);
 
             loop {
                 tokio::select! {
                     biased;
-                    () = shutdown_token.cancelled() => {
-                        tracing::info!("Blockchain syncer shut down.");
```
SyntheticBird45, line 130: Why not keeping it ?

redsh4de: After we break we'll go back to the outer loop that has the same log + return, line 97

SyntheticBird45 (approved):
i don't like the tuple being handed around, but apart from that LGTM

redsh4de:
Thoughts on wrapping the stream and task  into a `BlockDownloaderHandle` so it's not a tuple?

SyntheticBird45:
> Thoughts on wrapping the stream and task into a `BlockDownloaderHandle` so it's not a tuple?

i'm all for it

SyntheticBird45 (approved):

Boog900 (approved):

PR #707, opened 2026-09-14
Storage: Fix fjall rebuild

Boog900:
This PR: #680 added a new fjall keyspace to keep track of the top hash in the fjall DB. We use this hash to check if we need to rebuild fjall. Newly created DBs wont have this field so we assume that empty tapes and fjall databases without the top hash are in sync. Old databases would rebuild fjall automatically.

This PR: #681 added pruning. With old databases the tapes database metadata would fail to read resetting the metadata to empty. This is fine, databases would resync automatically.

However this only works if you take each step individually old DB -> #680 -> #681 if you go old DB -> #681 then the fjall top hash wont be there _and_ the tapes will be empty, meaning the database would see them as being in sync when they are not.


storage/blockchain/src/database.rs
```diff
@@ -448,7 +448,10 @@ impl BlockchainDatabase {
         let fjall_tip = fjall.get(&self.chain_tip, CHAIN_TIP_KEY)?;
 
         Ok(match (tapes_tip, fjall_tip.as_deref()) {
-            (None, None) => true,
+            // Old fjall DBs do not have the `chain_tip` key space, with the pruning upgrade this would
+            // make cuprate think an empty tapes and full fjall DB is in sync. So for now check that
+            // fjall is really empty.
+            (None, None) => fjall.is_empty(&self.block_heights)?,
```
SyntheticBird45, line 454: We should add a TODO to remind us to delete that branch after a period of time (once everyone migrated).

Boog900: It is harmless so I think we can just keep it around IMO

SyntheticBird45: I just wanna make sure we clean legacy reference at some point, but alr ig

SyntheticBird45 (approved):

storage/blockchain/src/database.rs
redsh4de: agree, i'd at least clear the comment after a while, branch can be kept imo

redsh4de (approved):

